PR #355, opened 2014-06-05
Update histogram_client.py

kidpixo:
Hi,
this should do the trick, calculating the initial bin size based on [this](http://www.fmrib.ox.ac.uk/analysis/techrep/tr00mj2/tr00mj2/node24.html).

I follow your example, extracting the mean bin size for all the `data`, but I would better have an array of binsize, because in my approach it depends on the single data distribution.

Anyway,  the `auto_nbin` is not set initially to the value from the `_auto_nbin` function, any idea?

Cheers,
Mario.

### DIFF
--- a/glue/clients/histogram_client.py
+++ b/glue/clients/histogram_client.py
@@ -220,8 +220,10 @@ class HistogramClient(Client):
         data = set(a.layer.data for a in self._artists)
         if len(data) == 0:
             return
-        dx = np.mean([d.size for d in data])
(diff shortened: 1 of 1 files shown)


astrofrog:
The histogram code has now been completely re-written

I'm going to close this since the scatter widget code has now been completely re-written, but hopefully this would be simpler to implement with the new viewer code.

@kidpixo - if you are interested in porting over your code to the new viewers, just email me or reach out to me on Slack (you can get an invite here: https://glueviz-slack-invite.herokuapp.com/)

PR #373, opened 2014-06-17
Tiled window layout

ChrisBeaumont:
Adds some intelligence to how windows are "snapped" to a regular layout

Makes window tiling undo-able


astrofrog:
I am going to close this, as it has been so long that the conflicts would be difficult to resolve - best start a new PR if we want to re-do this.

PR #535, opened 2014-12-27
Auto-add revision information to __version__?

astrofrog:
@astrofrog is there an easy way to automatically assign a unique and "valid" value for `__version__` for each commit? It would help when trying to reproduce bugs


ChrisBeaumont:
I assume that this works well enough with LooseVersion comparisons (at least for the stuff before `.dev`?)

astrofrog:
@ChrisBeaumont - ok, I'm pretty happy with it this way. If importing from a clean source directory it will simply say ``0.5.0.dev``. It only adds the hash when running ``setup``. If this passes, I'll merge if there are no objections.

PR #541, opened 2015-01-06
Add dependency for h5py

astrofrog:
Hi,
 I've noticed h5py is not in the dependencies list and it seems to me that is an important dependency since the tests fail if it's not installed and it's part of core/io.py

Cheers

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -3,17 +3,17 @@
 Installing Glue
 ===============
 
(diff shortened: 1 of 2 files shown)


astrofrog:
@noahfx - thanks for the report, I have attached a patch to this issue to fix this.

ChrisBeaumont:
This should also be added to `_deps.py`

astrofrog:
@ChrisBeaumont - thanks, done!

PR #551, opened 2015-01-09
Make glue.test() work

astrofrog:
... so as to be able to run testing on installed versions of glue.


astrofrog:
Patch attached, actually pretty easy, can add options later but this is already quite useful for now (this came up in the context of testing a conda-installed package).

PR #552, opened 2015-01-16
Reorganize widgets and clients into individual folders?

astrofrog:
At the moment, code for e.g. the histogram stuff is split into the client (in ``glue/clients``) and the widget (in ``glue/qt/widgets``). Of course, this separation makes sense in case we want to use different GUIs in future, but what I think might be easier to manage is something like this:

```
glue/
    viewers/
        histogram/
            client.py
            qt.py
            histogram.ui
        image/
            ...
```

i.e. to group the different viewers/widgets into individual folders and then inside that have a separation of the general glue client code, the qt code, and we can even add other front-ends in future, e.g. ``glue/viewers/histogram/web.py``.

In my opinion, this would be a step towards more modularity, but there may be reasons why we might not want to do that. @ChrisBeaumont, what do you think?

If we agree something like this would be good, I can go ahead and open a PR with the changes.


astrofrog:
This is very much work in progress, but I think that tests should continue to pass. The MacOS X app will definitely be broken for now because of having moved some of the ``ui`` files. At the moment, this still doesn't feel right because of some circular imports, and I think that the ``glue.qt`` folder should be further split into code that gets *used* by the widgets for example, and the actual current glue app (which is just one instance of a possible app once can build with the framework). I'll continue to work on this to see how one might organize this.

astrofrog:
The travis failure is due precisely to a circular import issue as mentioned above.

astrofrog:
Just to try, I've now also tried moving code specific to the actual glue application to a ``glue.app`` sub-package. The idea here is to separate code that can be used by other glue front-ends (e.g. PV slicer, other custom glue GUIs) from the current instance of the glue GUI. This reduces circular imports (because before the viewers needed ``glue.qt`` and also were needed by ``glue.qt``).

astrofrog:
This is of course still very much work in progress, but @ChrisBeaumont, feel free to leave comments anytime if you think this is going in the wrong direction or if you have suggestions for improvement.

astrofrog:
Here is an overview of the changes so far:

* Created the following sub-packages:

  * ``glue.viewers.dendro``
  * ``glue.viewers.ginga``
  * ``glue.viewers.histogram``
  * ``glue.viewers.image``
  * ``glue.viewers.scatter``
  
  Each of these contains the code for the generic client and the associated Qt widget (and ui file). The aim of this change is that to build other GUIs, anything related to the image widget will be in ``glue.viewers.image``, etc. It also helps isolated dependencies such as ginga into a single sub-package.
  
* Moved code relating specifically to the glue application to ``glue.app``.
  This therefore separates the specific glue application from the components
  which can be re-used to build other applications.

* The image, histogram, ginga, and scatter widgets are now added to the widget registry directly in ``config.py`` to avoid having to import the widgets into ``glue.qt.widgets`` and adding them to ``default_widgets`` (which would cause circular imports now some of the widgets are in ``glue.viewers``, since those in turn depend on ``glue.qt``). This is cleaner anyway since the defaults are then all contained in ``config.py``.

* Related to the above, for now I left ``enable_dendrograms`` in ``glue.qt.widgets`` and I make it add the dendrogram widget directly to the registry.

* For now, ``load_ui`` has an extra option to specify whether the file should be fetched in the global ui folder (``glue.qt.ui``) and if not, the path is kept as-is. This is because the ui files have now moved in some cases to be alongside the code. This currently will break the MacOS X app build, but I will address this once the rest is done.

* Removed ``tree_client.py`` since it doesn't appear to be used anywhere?

* Updated API docs accordingly.

Remaining to-dos:

* [ ] make sure that the ui files can be found by the mac app
* [ ] decide if we need to provide backward-compatibility for imports

@ChrisBeaumont - could you review this once you get a chance? There are some further changes/simplifications I would like to propose for the ``glue.clients`` and ``glue.qt`` sub-packages, but I don't want to make this PR unmanageable, so I was thinking of doing it in a separate PR.

astrofrog:
**Important Note:** if we merge this, we have to either modify or remove the ``ui_path`` patch in https://github.com/glue-viz/Travis-MacGlue/blob/master/glue_recipe.py#L9. I tested without that patch and it works fine.

astrofrog:
Just for info, the app still works after these changes: http://glue-astrofrog.s3-website-us-east-1.amazonaws.com/

ChrisBeaumont:
I will take a look!

On Tuesday, January 20, 2015, Thomas Robitaille <notifications@github.com>
wrote:

> Just for info, the app still works after these changes:
> http://glue-astrofrog.s3-website-us-east-1.amazonaws.com/
>
> —
> <https://github.com/glue-viz/glue/pull/552#issuecomment-70659813>.
>


-- 
Chris Beaumont
Senior Software Engineer
60 Garden Street, MS 42
Cambridge, MA 02138
chrisbeaumont.org

glue/app/glue_application.py
```diff
@@ -15,21 +16,23 @@
...
+                         GlueTabBar, load_ui, get_icon, nonpartial)
+from ..qt.widgets.glue_mdi_area import GlueMdiArea, GlueMdiSubWindow
+from ..qt.widgets.edit_subset_mode_toolbar import EditSubsetModeToolBar
+from ..qt.widgets.layer_tree_widget import PlotAction, LayerTreeWidget
+from ..qt.widgets.data_viewer import DataViewer
+from ..qt.widgets.settings_editor import SettingsEditor
+from ..qt.widgets.mpl_widget import defer_draw
+from ..qt.feedback import submit_bug_report
+
+ROOT = os.path.dirname(os.path.abspath(__file__))
```
ChrisBeaumont: I never get this logic exactly right -- have you checked that if it works when used:

- As a mac app
- From python setup.py develop
- From pip install glue-viz

ChrisBeaumont:
I think I'm on board with this. One thought I had when looking through the diffs: if we ever support something beyond qt, it would be idea to have an easy way to specify the backend like MPL. Right now, config.py explicitly imports from a lot of qt modules, which is cumbersome (especially if anyone want's to customize their config files). 

We don't have to implement such a switching feature in this PR, but do you have any idea if the reorganization would make such switching harder?

glue/app/glue_application.py
astrofrog: I'll make sure I try all of those. The Mac app definitely works.

astrofrog:
@ChrisBeaumont - I'll have a think about how we might be able to switch either GUI toolkit or plotting library.

astrofrog:
@ChrisBeaumont - having thought about it, I don't think switching would be harder. In particular, if we want to keep ``config.py`` as a global config regardless of backend/frontend, we can import generic classes for the viewers such as:

```
from glue.viewers.image import ImageViewer
```

ImageViewer could then be a factory class that returns e.g. a Qt Widget, or whatever we decide to use instead, based on a global configuration open. I definitely agree that ``config.py`` should be made less backend/frontend dependent.

astrofrog:
I'm going to work more on this as I'm still not entirely happy with it

astrofrog:
I was looking at the code and trying to figure out how one would change backends from matplotlib to something else, and there are a few things that I think we can do but wasn't sure whether to do as part of this PR:

* In glue.clients.layer_artists, the base classes are plotting-library independent, but the specific classes like ``LayerArtist`` are matplotlib-specific. We could move that whole file to ``glue.core`` (since it's pretty generic) and then also split up the file into the generic base classes and the matplotlib-specific classes (and add a Mpl suffix or prefix to them).

* Same thing in ``glue.clients.viz_client`` - some of it is non-matplotlib dependent, some is.

* In the image client, we currently call ``modest_image.imshow`` directly, but this could be made into a LayerArtist and abstracted out.

In general, I think it would be nice to isolate all the matplotlib code so as to be able to easily change it like you suggest. I need to think about this more, and whether it means that this PR is a bad idea in that it is scattering the Qt code over several places and we might in future then wish all the Qt code was in a single place (which is the way you had it).

More to think about, clearly!

astrofrog:
I guess maybe what I was getting at in this PR is that it would be good to have a layer of API that is GUI and plotting library independent, so that someone developing a custom tool for example could say "show me a scatter plot with this data" and it will take care of using the scatter widget and canvas for the correct plotting and GUI framework. So in a sense that was the idea behind having a ``glue.viewers`` namespace. ``glue.viewers.ImageViewer`` would then return a generic viewer which will appear differently depending on what GUI toolkit one is using. Then the ``config.py`` file can also use these more generic classes instead of importing from ``glue.qt``.

astrofrog:
One final note - I think it's still worthwhile trying to split out the specific glue app from ``glue.qt`` into e.g. ``glue.qtapp`` or ``glue.app`` since it is one specific app that can be built with the glue components, but not the only one.

I'll try and think some more about all these issues, but @ChrisBeaumont - if you have any thoughts, let me know!

ChrisBeaumont:
I don't think we need to add the "switch backend" functionality to this PR (especially since there isn't an alternative backend to switch to). rather, I just want to make sure the architecture doesn't evolve in a direction that makes such switching harder.

I think I agree with your basic approach. A config.py file would look something like

```
set_gui('qt')
set_plotting('matplotlib')
```

And from then not have to import or directly reference Qt's scatter widget compared to any alternative.

astrofrog:
I think I'd like to get back to this (though I'll need to start from scratch) because when debugging the dendrogram stuff it was inefficient to have dendrogram code in many different places, so I'd like to see the viewers organized internally as if they were plugins (much more self-contained). To make this more manageable, I think that based on the above discussion, I will first do:

* a pull request to organize things into folders per viewer

* another pull request later to add a layer of API that is backend/frontend independent and the ability to switch

@ChrisBeaumont - if you have no objections, I'll try and focus on this once some of the existing PRs are merged.

astrofrog:
Having thought about this a fair bit, I don't think that this will make it harder to switch plotting/GUI frameworks. In fact, each viewer/component could have a formal way of declaring which GUI frameworks and plotting libraries it supports (since not all components will support all plotting/GUI frameworks).

The final glue *application* should still be in one single module of course.

astrofrog:
I'm closing this since I'm now doing the refactoring in other pull requests, so this one is obsolete.

PR #557, opened 2015-01-21
Fix glue.test() to work in MacOS X app

astrofrog:
@ChrisBeaumont - I think this will allow us to run the tests in the MacOS X app. I'm building the app to test.

### DIFF
--- a/glue/__init__.py
+++ b/glue/__init__.py
@@ -128,6 +128,8 @@ def test(no_optional_skip=False):
     import os
     from pytest import main
     root = os.path.abspath(os.path.dirname(__file__))
+    if 'site-packages.zip' in root:
+        root = root.replace('site-packages.zip', 'glue')
     args = [root]
     if no_optional_skip:
         args.append('--no-optional-skip')


astrofrog:
This is not as trivial unfortunately. Looking into options, but I think that we'll either have to:

* Store site-packages.zip uncompressed and with the glue source files (not needed for other packages). This could be done as a patch in the app building process.

* Re-write the tests to use absolute imports and store them in a separate directory in the app to make sure it doesn't test the separate source install that exists alongside ``site-packages.zip``.

astrofrog:
For self-reference:

https://bitbucket.org/ronaldoussoren/py2app/issue/165/allow-site-packages-to-not-be-compressed

http://stackoverflow.com/questions/28073033/running-pytest-on-module-inside-site-packages-zip

ChrisBeaumont:
How much larger is the app with uncompressed site-packages? That sounds easier unless it gets really bulky

astrofrog:
I'm going to close this, because this solution doesn't work. I've opened an issue to keep track of it: https://github.com/glue-viz/glue/issues/751 (just trying to close stale pull requests that are no longer useful)

PR #558, opened 2015-01-22
Enable OS X testing and simplify build matrix

astrofrog:
We should wait until the multi-os is enabled on Travis (should be soon, I asked) then restart the job and merge if it passes.


ChrisBeaumont:
Looks good to me

PR #559, opened 2015-01-24
Reorganize utilities

astrofrog:
There are a number of utilities in various places in glue that are not specific to glue or data linking and only depend on the standard library or external dependencies. This is an attempt to start bringing together all these types of utility functions into a ``glue.utils`` sub-package and tries to split them into various categories since there are quite a few. These utilities are not allowed to import anything from glue itself (except ``external`` for ``six``) so it will help isolate that code.

The test coverage is quite low for this sub-package so this is still very much WIP since I'd like to increase the coverage to close to 100%.


astrofrog:
Note to self, ``small_view`` should stay in ``core`` since it is specific to ``Data`` slicing.

ChrisBeaumont:
I wonder if this would make importing logic hard to guess, because the namespace is very spread out? What do you think about importing all the sub-utils in the main utils namespace, and then using `utils.nonpartial` instead of `utils.misc.nonpartial` elsewhere in the codebase?

astrofrog:
@ChrisBeaumont - good idea, I'll do that! I agree at the moment it's a little too spread out.

astrofrog:
I've now added a number of tests so that direct coverage of the functions is now better (of course they were being indirectly exercised before by other tests, but it's nice to have some simple tests to test them directly)

@ChrisBeaumont - if this passes, are you happy with this?

ChrisBeaumont:
Looks good to me!

PR #566, opened 2015-02-22
Fix docs build

astrofrog:
Replaces #555. This uses ``setuptools`` to locally install astropy-helpers automatically when building the docs (had to hack around before finding how to do that). So this means that users will never see anything to do with astropy-helpers and won't have to install it, it will all be done transparently.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -11,9 +11,6 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys
-import os
-
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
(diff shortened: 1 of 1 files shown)


doc/conf.py
```diff
@@ -24,14 +21,21 @@
 # If your documentation needs a minimal Sphinx version, state it here.
 #needs_sphinx = '1.0'
 
+# The following ensures that astropy-helpers is downloaded from PyPI and
+# installed locally in this directory, since it is needed for the Sphinx
+# extensions.
+from setuptools import Distribution
+Distribution({'setup_requires': 'astropy-helpers'})
```
embray: I would just add astropy-helpers as a normal dependency for building the docs, and not use this.

astrofrog: In the end we do install astropy-helpers for travis, this is just for users who want to build locally. I thought we didn't want people to have astropy-helpers installed as a package? (and most users aren't familiar with virtualenv)

PR #567, opened 2015-02-22
Reorganize docs

astrofrog:
I reorganized the docs into sub-directories and am now working on getting rid of all warnings during the docs builds.


astrofrog:
Down to 30 warnings in Python 3

astrofrog:
@ChrisBeaumont - are you ok with this reorganization? (just making use of sub-directories in the ``doc`` directory). Also, note that a few images have been removed because they didn't appear to be used anywhere.

There are still 12 warnings but I don't have time to look into them now and they aren't important.

astrofrog:
I've now fixed the last of the warnings (and had to add a few exceptions to the ``nitpick`` list for now) and have added a test that will fail if any warnings are raised during the docs build. This ensures that we no longer introduce any wrong references or make formatting mistakes in docstrings in future.

astrofrog:
Apologies for the debugging, just trying to get to the bottom of warnings that are appearing only on Travis. Will be ready again for review soon.

ChrisBeaumont:
Does this reorganization break old links at all?

Otherwise, this seems like a clear improvement. :+1:

astrofrog:
@ChrisBeaumont - actually what I'm trying to do here is make sure there are absolutely no warnings (which would indicate broken links) so we should be ok. I will run a standard link checker on the docs to check.

The only thing it might break is links on other websites, so I'll check astrobetter for example to see if any of these links will break. Anywhere else I should check?

astrofrog:
Yay, the docs build is now warning free and Travis is passing again :)

astrofrog:
@ChrisBeaumont - the astrobetter post does have some links to some of the pages that have moved and those links will break (but I can fix them). Are there other places with links we should be concerned about or can I go ahead and merge this?

astrofrog:
Maybe I can use a plugin like https://github.com/stackforge/murano-specs/blob/master/doc/source/redirect.py to set up redirects from old page locations.

astrofrog:
Here's the original plugin: https://github.com/openstack/nova-specs/blob/master/doc/source/redirect.py will have a go at that now.

astrofrog:
@ChrisBeaumont - good news, the plugin was very easy to use and all links from other websites will now get redirected. Does this sound ok?

astrofrog:
(by the way, I do actually plan to make contributions to the docs, this is just some initial reorganization to make it easier to manage the docs as they grow!)

ChrisBeaumont:
Yeah if we can prevent old links from going stale, this is a definite improvement moving forward. Thanks!

astrofrog:
Just a note (for future reference) that this also changes the version number of glue to be PEP440 compliant (so e.g. ``0.4.0.dev123``). Can't use the hash in the version number but it's available as ``__githash__`` if needed.

astrofrog:
Ok, the docs should now be warning-free and all internal API references should be fixed :smile:

ChrisBeaumont:
image

Issue #568, opened 2015-02-23
Issue with S3 upload of docs


astrofrog:
Seems to raise an error:

```
if [ `git symbolic-ref -q --short HEAD` == 'master' ]; then aws s3 sync _build/html s3://docs.glueviz.org/en/latest; fi
/bin/sh: 1: [: ==: unexpected operator
```

Appears to have been broken for a little while.

@ChrisBeaumont - any idea what is going on in this case?

astrofrog:
(I can also look into it, but wondering if it's been a known issue)

ChrisBeaumont:
Hm this works for me *unless* I checkout an explicit revision and enter "detached head state"

```
#cbeaumont:glue cbeaumont$ git branch
* (detached from d2595d5)
  master
  tmp
#cbeaumont:glue cbeaumont$ git symbolic-ref -q --short HEAD # nothing
```

It works for me if I'm on a named branch.

The relevant fix would then be something like

```
`git symbolic-ref -q --short HEAD || echo '---'`
```

Unless you are getting an error in situations where you expect the test to pass

astrofrog:
Ok, thanks! This is happening on the master branch on Travis (just realized I forgot to say that above). Which should be a real branch. I will investigate.

astrofrog:
I'll do some debugging on Travis in my glue fork (will print out ``git symbolic-ref -q --short HEAD`` and ``git branch`` to see what might be going on).

ChrisBeaumont:
Yeah maybe travis switched its logic for how it checkouts repositories, such that it worked before but doesn't now. In any event we just need to change the test so that it compares the sha1 of the current working directory against the sha1 of master.

astrofrog:
It's a pain that we can't use the ``TRAVIS_BRANCH`` env variable:

> TRAVIS_BRANCH:For builds not triggered by a pull request this is the name of the branch currently being built; whereas for builds triggered by a pull request this is the name of the branch targeted by the pull request (in many cases this will be master)

(because for PRs it will be ``master``)

astrofrog:
Ah but we can use TRAVIS_PULL_REQUEST=false/true too!

astrofrog:
Although if we do that it won't work when not on Travis. One option is to remove the if statement from the Makefile and insert it in the ``.travis.yml`` file. But in any case, something like

```
[ $TRAVIS_BRANCH == master and $TRAVIS_PULL_REQUEST == false ]
```

should work.

[...]

PR #569, opened 2015-02-23
Fix publishing of docs to S3

astrofrog:
See #568

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -154,4 +154,4 @@ doctest:
 	      "results in $(BUILDDIR)/doctest/output.txt."
 s3:
 	aws s3 sync _build/html s3://www.glueviz.org/en/`git symbolic-ref -q --short HEAD`
-	if [ `git symbolic-ref -q --short HEAD` == 'master' ]; then aws s3 sync _build/html s3://docs.glueviz.org/en/latest; fi
+	aws s3 sync _build/html s3://docs.glueviz.org/en/latest; fi
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -154,4 +154,4 @@ doctest:
 	      "results in $(BUILDDIR)/doctest/output.txt."
 s3:
 	aws s3 sync _build/html s3://www.glueviz.org/en/`git symbolic-ref -q --short HEAD`
-	aws s3 sync _build/html s3://docs.glueviz.org/en/latest; fi
+	aws s3 sync _build/html s3://docs.glueviz.org/en/latest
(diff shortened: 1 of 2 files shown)


doc/Makefile
```diff
@@ -154,4 +154,4 @@ doctest:
 	      "results in $(BUILDDIR)/doctest/output.txt."
 s3:
 	aws s3 sync _build/html s3://www.glueviz.org/en/`git symbolic-ref -q --short HEAD`
-	if [ `git symbolic-ref -q --short HEAD` == 'master' ]; then aws s3 sync _build/html s3://docs.glueviz.org/en/latest; fi
+	aws s3 sync _build/html s3://docs.glueviz.org/en/latest; fi
```
ChrisBeaumont: I think the `;fi` is leftover

ChrisBeaumont:
Looks good to me

astrofrog:
Ok, let's try this in master

astrofrog:
Travis seems to be uploading the docs but I'm not seeing the latest version online. Even though the upload command shows:

```
...
upload: _build/html/python_guide/index.html to s3://docs.glueviz.org/en/latest/python_guide/index.html
...
```

http://www.glueviz.org/en/latest/python_guide/index.html does not work. @ChrisBeaumont - can you check what things look like on the S3 side since you have access to that?

doc/Makefile
```diff
@@ -154,4 +154,4 @@ doctest:
 	      "results in $(BUILDDIR)/doctest/output.txt."
 s3:
 	aws s3 sync _build/html s3://www.glueviz.org/en/`git symbolic-ref -q --short HEAD`
```
ChrisBeaumont: Ah @astrofrog I think this is the issue. We also want to change the backticks here to $TRAVIS_BRANCH, otherwise the docs are uploaded directly to glueviz.org/en instead of glueviz.org/en/branch. Sorry for not catching this

doc/Makefile
```diff
@@ -154,4 +154,4 @@ doctest:
...
-	if [ `git symbolic-ref -q --short HEAD` == 'master' ]; then aws s3 sync _build/html s3://docs.glueviz.org/en/latest; fi
+	aws s3 sync _build/html s3://docs.glueviz.org/en/latest
```
ChrisBeaumont: Also I think should be `www.glueviz.org` not `docs.glueviz.org` (this is a stale bug)

doc/Makefile
astrofrog: Thanks for spotting this, fixed this and the issue below in 197b862fa8ded05fa659d77f69b85f4ea305dc61

Issue #574, opened 2015-02-25
Use HTML formatting when creating a new component


astrofrog:
When defining new components, the use of ``{}`` to indicate attributes might be improved by instead using HTML highlighting syntax. For example by doing

```
text.setHtml('<font color="red"><b>data.ra</b></font> + <font color="blue"><b>data2.ra</b></font>')
```

we can get:

![screen shot 2015-02-25 at 2 50 44 pm](https://cloud.githubusercontent.com/assets/314716/6371627/f5f12c8c-bcfd-11e4-9979-4c98df1c3ddf.png)

It may even be possible to use CSS and make them look like tags which would improve readability a lot.

This would require writing a small translation layer to/from HTML but it might be worth it for readability.

cc @stargaser @ChrisBeaumont

astrofrog:
(I can hack around with this later to see if it would actually work)

astrofrog:
Turns out it's very easy :)

![screen shot 2015-02-25 at 16 25 19](https://cloud.githubusercontent.com/assets/314716/6373442/028c112a-bd0b-11e4-88cc-1bd5facc0fd6.png)

Now of course this is just a little visual formatting but it might be worth thinking about whether we can make this more readable. For example, maybe dataset names should be sanitized by default to not include extensions. Maybe we can also get rid of the parentheses and instead present it like

**2MASS_k.Pixel_x**

Anyway, @ChrisBeaumont - if you have any thoughts about this, let me know and I'll work on a more comprehensive PR if needed.

ChrisBeaumont:
My only motivation for the brackets was that, if you wanted to, you could
type it out. But maybe readability is more important than typability. I
wonder if it's also possible to add hover text over specific words? Then
you could also abbreviate the labels

On Wednesday, February 25, 2015, Thomas Robitaille <notifications@github.com>
wrote:

> Turns out it's very easy :)
>
> [image: screen shot 2015-02-25 at 16 25 19]
> <https://cloud.githubusercontent.com/assets/314716/6373442/028c112a-bd0b-11e4-88cc-1bd5facc0fd6.png>
>
> Now of course this is just a little visual formatting but it might be
> worth thinking about whether we can make this more readable. For example,
> maybe dataset names should be sanitized by default to not include
> extensions. Maybe we can also get rid of the parentheses and instead
> present it like
>
> *2MASS_k.Pixel_x*
>
> Anyway, @ChrisBeaumont <https://github.com/ChrisBeaumont> - if you have
[...]

PR #575, opened 2015-02-25
Improvements to new component widget

astrofrog:
### About this PR

This was inspired by the original discussion in #574. Two main things are implemented:

* Double clicking components adds them in bold and blue to the expression
* If you start typing the name of a component and press tab, a completion window comes up

### Caveats

This is however now functional at this point because the final expression needs to be extracted as HTML and parsed back to the original format with the ``{}``.

I think that the tab-completion addresses the comments in #574 that the ``{}`` may be required for typing directly since typing directly and using the tab completion turns the name into a bold/blue component, however, it does not deal with the case where the user types out the whole expression

### Demo

A demo is here: https://www.youtube.com/watch?v=lN_RVphq2hs&feature=youtu.be

### To-dos:

* [x] Deal with the use case where the user types out the whole name completely
* [x] Parse output HTML from expression
* [x] Try and find nicer names for the components
* [x] Add back and edit tool tip for expression

@ChrisBeaumont @stargaser - what do you think about this? I don't mind if you think we should just leave it as-is, since this was educational to work on anyway :)


ChrisBeaumont:
@astrofrog this is really nice! I never liked this dialog, but these changes would go a long way to making it more user-friendly.

While we are revamping this window, another useful feature would be to show a preview of what the expression evaluates to -- i.e. by computing the first few elements and printing to a window (probably a new window under "Add to", which doesn't need all the space). The biggest pain-point of this dialog is that the user might enter a typo, or forget to highlight a dataset in the "add to" window. If we could ensure the expression evaluates properly and give the user some visual feedback, before letting them close the window, that would be great.

astrofrog:
Good idea about the preview. I wonder if something like a histogram of values would be a nice way to show it. I guess it might not always be fast though.

ChrisBeaumont:
Yeah a histogram would be cool too (if speed becomes a problem you could do something different for big and small datasets)

astrofrog:
@ChrisBeaumont - I've spent the evening hacking on this, and I think this PR is ready for review/testing. Here is a video demonstrating the latest status (warning, it has sound):

[demo](http://www.youtube.com/watch?v=7H9wg9xYXvc)

What is not implemented yet is the ability to preview the results, but in the spirit ok keeping PRs manageable, I will do that in a separate PR (it will require more experimentation anyway)

astrofrog:
@stargaser - let me know what you think too since you've been playing around with this recently (see comment + video above)

glue/qt/widgets/custom_component_widget.py
```diff
@@ -114,12 +201,25 @@ def create_component(collection):
...
+            if widget.ui.exec_() == QDialog.Accepted:
+                if len(str(widget.ui.expression.toPlainText())) == 0:
+                    QMessageBox.critical(widget.ui, "Error", "No expression set",
+                                         buttons=QMessageBox.Ok)                    
+                elif widget._number_targets == 0:
+                    QMessageBox.critical(widget.ui, "Error", "Please specify the target dataset(s)",
+                                         buttons=QMessageBox.Ok)
+                elif len(widget.ui.new_label.text()) == 0:
+                    QMessageBox.critical(widget.ui, "Error", "Please specify the new component name",
+                                         buttons=QMessageBox.Ok)
```
ChrisBeaumont: How about trying to compute the first element of the new component, and displaying an errorbox if an exception is thrown?

glue/utils/qt/autocomplete_widget.py
```diff
@@ -0,0 +1,127 @@
...
+        if self.completer:
+            self.disconnect(self.completer, 0, self, 0)
+        if not completer:
+            return
+
+        completer.setWidget(self)
+        completer.setCompletionMode(QtGui.QCompleter.PopupCompletion)
+        completer.setCaseSensitivity(QtCore.Qt.CaseInsensitive)
+        self.completer = completer
+        self.connect(self.completer, QtCore.SIGNAL("activated(const QString&)"), self.insert_completion)
```
ChrisBeaumont: Any reason you didn't use self.completer.activated.connect(self.insert_completion)? That's the style used elsewhere in the code

glue/utils/qt/autocomplete_widget.py
```diff
@@ -0,0 +1,127 @@
...
+# http://qt-project.org/doc/qt-4.8/tools-customcompleter.html
+
+from ...external.qt import QtGui, QtCore
+
+
+__all__ = ["CompletionTextEdit"]
+
+try:  # Python 2
+
+    QString = QtCore.QString
```
ChrisBeaumont: So we're using version 2 of the Qt api, which is supposed to do away with QString and just use unicode instead. Is this really needed here? I suspect there's a cleaner way that avoids QString

glue/utils/qt/autocomplete_widget.py
astrofrog: Part of this code was copied from elsewhere, but I'll fix this

glue/utils/qt/autocomplete_widget.py
astrofrog: Ah I was using this as standalone for testing so maybe the API wasn't getting set. I'll see if I can simplify all this now it's being used in glue.

astrofrog: Fixed!

glue/utils/qt/autocomplete_widget.py
astrofrog: Fixed!

glue/qt/widgets/custom_component_widget.py
astrofrog: Good idea, I'll try and add that tomorrow.

stargaser:
@astrofrog Looks great! I like the color changing from red to blue when the component is properly defined. Thanks also for putting in checking for the cases when the user forgets to select the Add to or to make a label -- this tripped me up before.

Blue with an underline always says to me "hyperlink", so I'm wondering if it would be better to not have the underlining, if it is possible?

ChrisBeaumont:
> Blue with an underline always says to me "hyperlink"

Agreed, what about red + green? More specifically something like #5cb85c and #d9534f (what bootstrap uses, a bit less garish than pure colors)

astrofrog:
Ah yes I should have said the colors are completely arbitrary and open to suggestions. Will try @ChrisBeaumont's suggestions.

astrofrog:
Yes, this looks much better. Only issue with green/red is it's the worst combination for color blindness. Would leaving the underline under the green be ok and not look like a link?

ChrisBeaumont:
Oh yeah. I guess Blue/Red would also work (#337ab7 blue, of course :))

stargaser:
Here is a color-blind friendly palette in R code, from http://www.cookbook-r.com/Graphs/Colors_%28ggplot2%29/

# The palette with grey:
cbPalette <- c("#999999", "#E69F00", "#56B4E9", "#009E73", "#F0E442", "#0072B2", "#D55E00", "#CC79A7")

astrofrog:
@ChrisBeaumont @stargaser - fixed!

astrofrog:
@ChrisBeaumont - do you foresee any backward-compatibility issues with loading existing glue 'saved state' files?

glue/qt/widgets/custom_component_widget.py
astrofrog: I think this can be combined with https://github.com/glue-viz/glue/issues/577 which will require calculating values before the dialog is closed anyway.

ChrisBeaumont:
The only relevant state is for the new component links, and those don't
appear to have changed. We should be fine

On Saturday, February 28, 2015, Thomas Robitaille <notifications@github.com>
wrote:

> @ChrisBeaumont <https://github.com/ChrisBeaumont> - do you foresee any
> backward-compatibility issues with loading existing glue 'saved state'
> files?
>
> —
> <https://github.com/glue-viz/glue/pull/575#issuecomment-76525007>.
>


-- 
Chris Beaumont
Senior Software Engineer
60 Garden Street, MS 42
Cambridge, MA 02138
chrisbeaumont.org

astrofrog:
Ok, thanks! In that case I'll merge this unless there are any objections. Thanks for the reviews!

PR #576, opened 2015-02-27
Fix StyleDialog in Python 3

astrofrog:
@ChrisBeaumont - quick question, is what I am doing here the best way to make sure the style dialog becomes non-blocking when testing, or is there an easier way?


astrofrog:
I'm going to go ahead and merge this since this is currently broken in Python 3.

ChrisBeaumont:
Sorry for missing this earlier. This looks fine. I often end up using mock to mock out the function that actually displays widgets or blocks, whenever possible. Usually the logic in the test doesn't depend on the widget actually materializing itself.

PR #578, opened 2015-02-27
Replace old coordinate transformation functions from APLpy by Astropy

astrofrog:
This requires Astropy 0.4 in order to work.

Note that Astropy now allows us to have many more kinds of coordinate conversion so we may want to try and think of a more general way to handle these. I think this should actually be developed as a plugin really (a default plugin, but a plugin nevertheless).


astrofrog:
Remaining todo: add conversion accuracy tests (just a couple) to make sure it's all good.

astrofrog:
I'm going to try and refactor the coordinate conversions as a plugin since they are domain-specific (as a general rule, I guess any domain-specific stuff should be plugin-ized?)

astrofrog:
This now includes some refactoring into a plugin.

@ChrisBeaumont - a quick question: do we want to deprecate the single coordinate conversion, e.g ``radec2glon`` and so on. I'm not suggesting removing it, but I can add a flag to helpers that indicates whether they are deprecated, and if so we keep them internally but don't display them in the list? If we do that I will then be able to add a few more common coordinate conversions (but the list would be unmanageable if we had to create one class and four functions for each possibility)

astrofrog:
(just for info the last commit adds an example of how we can deprecate the functions, they then no longer appear in the list)

ChrisBeaumont:
Deprecating sounds fine.  They are almost never useful

On Saturday, February 28, 2015, Thomas Robitaille <notifications@github.com>
wrote:

> (just for info the last commit adds an example of how we can deprecate the
> functions, they then no longer appear in the list)
>
> —
> <https://github.com/glue-viz/glue/pull/578#issuecomment-76523306>.
>


-- 
Chris Beaumont
Senior Software Engineer
60 Garden Street, MS 42
Cambridge, MA 02138
chrisbeaumont.org

astrofrog:
I have now deprecated the single functions for coordinate conversions, and added a few more for the class-based helper. I also added the ability to give a helper class a display name for the drop-down menu so it doesn't have to use the class name:

screen shot 2015-02-28 at 8 32 44 pm

Now one downside of the approach I am using to create the link classes is that the result (once glued) looks like:

screen shot 2015-02-28 at 8 29 50 pm

which is ugly because of the 'forward' and 'backward'. In addition to that, I think only two links should be shown because celestial coordinates are always converted in pairs.

I'll need to think about it more but @ChrisBeaumont, if you have any thoughts, let me know!

astrofrog:
I've now made it so that the name of a PartialFunction includes the class name which I think makes it more readable:

screen shot 2015-03-07 at 12 46 00 pm

I also added accuracy tests for the coordinate transforms.

@ChrisBeaumont - this is now ready for review!

ChrisBeaumont:
Looks good!

PR #584, opened 2015-03-03
Workaround for non-working import in glue.qt.widgets.__init__.py

ejeschke:
- glue.config imports default_widgets from glue.qt.widgets, but
  this fails to import ginga correctly, even though later imports
  (from other modules) of default_widgets correctly imports ginga.
  Therefore the Ginga viewer doesn't show up in the list of possible
  viewers.  This workaround in config fixes the issue.

### DIFF
--- a/glue/config.py
+++ b/glue/config.py
@@ -222,9 +222,17 @@ class QtClientRegistry(Registry):
     def default_members(self):
         try:
             from .qt.widgets import default_widgets
+            # NOTE: necessary because somehow default_widgets does not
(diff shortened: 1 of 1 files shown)


astrofrog:
This works for me. In the long term I think that the definition of which widgets are available to the app need to live alongside the app in a different directory to where the widgets live (I had started to do something like this as part of https://github.com/glue-viz/glue/pull/552 which ended up being a PR that was likely too large). I will work on this soon.

PR #585, opened 2015-03-05
Move all ginga widget code to a self-contained plugin

astrofrog:
Since the ginga widget is an optional add-on (depending on whether the user has or does not have ginga installed), I think it would make sense to consider moving all ginga-related code to a self-contained plugin in ``glue.plugins.ginga``. In addition to helping better compartmentalize various parts of the glue code, it would help ensure that we provide an API that is sufficient to develop this kind of widget without having to make any changes in the main glue code base.

I'm happy to work on this unless there are any objections.

cc @ChrisBeaumont @ejeschke


astrofrog:
Currently this PR introduces a new ``load_plugins`` function that forces plugins to be loaded, in turn allowing plugins to register themselves in the right place. This means that any other code in glue doesn't need to know about the ginga widget and also changes what was done in #584. #578 takes the same approach so whichever of these PRs is merged first means the other will need to be rebased, but should be easy.

astrofrog:
A quick 'grep' reveals that there are no instances of ginga or Ginga anywhere else in the glue code apart from in ``plugins/ginga_viewer`` (except for ``tests/helper.py`` but that's fine)

astrofrog:
@ejeschke - could you test this branch out and make sure it all works as expected?

ejeschke:
As far as I can tell, this is working.

astrofrog:
@ChrisBeaumont - I'll leave this open for now so that you can comment on the plugin loading mechanism, in case you see any downsides. If we agree with it I can actually extend this to the PV slicer and Spectrum tool, i.e. ensure they are responsible for registering themselves.

glue/__init__.py
```diff
@@ -118,6 +118,11 @@ def make_selector_func(roi):
     from .qt.custom_viewer import CustomViewer
     return CustomViewer.create_new_subclass(name, **kwargs)
 
+# Load built-in plugins
+from .config import load_plugins
+load_plugins()
```
ChrisBeaumont: Hmm this will end up importing things like Qt immediately. We've always tried to keep glue.core importable without Qt (although maybe this isn't a useful restriction)

glue/plugins/ginga_viewer/qt_widget.py
```diff
@@ -5,9 +5,9 @@
 import numpy as np
 
 from ...external.qt.QtGui import (QAction,
-                                  QToolButton, QToolBar, QIcon,
-                                  QActionGroup, QWidget,
-                                  QVBoxLayout, QColor, QImage, QPixmap)
+                                   QToolButton, QToolBar, QIcon,
```
ChrisBeaumont: Whitespace -- should these lines be left one character for pep8?

glue/plugins/ginga_viewer/__init__.py
```diff
@@ -0,0 +1,10 @@
+try:
+    from .client import *
+    from .qt_widget import *
+except ImportError:
+    import warnings
+    warnings.warn("Could not import ginga plugin, since ginga is required")
```
ChrisBeaumont: I wonder if this is a logger.info-level message.

ChrisBeaumont:
I left a few minor notes/questions, but otherwise this looks good to me!

glue/__init__.py
astrofrog: @ChrisBeaumont - ah right, indeed it would be nice to avoid the Qt import. I will think about this a bit more.

astrofrog: Another way to load the plugins would be to actually *not* have the plugins auto-register but have that instead be an app-specific thing. That is, maybe people building other glue apps don't want those plugins loaded by default. So as part of the glue *app* we could have an explicit loading of the plugins.

glue/plugins/ginga_viewer/qt_widget.py
astrofrog: Indeed, not sure what happened there.

glue/__init__.py
astrofrog: Another intermediate solution is to have ``load_plugins`` be called when the app starts up. I should make that load all plugins, but I also need to make it possible to import a plugin without importing all of them. I'll work on this a bit and see what works best.

ChrisBeaumont: Last week I was playing around with the idea of using a lazily-loaded module. Instead of having to worry about when to import the plugins/config modules, we could set them up to transparently import the first time their modules are accessed. IOW, something like

```
import glue.config  # nothing happens here
...
glue.config.plugins  # plugins is a property which calls load_plugins and caches the result
```

That would simplify some of the logic of when something can/cant be imported, and delay slow imports until something is actually needed. 

You need to do a little hacking to get that to work -- http://stackoverflow.com/a/880550 . I wasn't convinced it was worth the magic, but I wonder what you think?

astrofrog: That could work too - I can look into all this while traveling later this week.

ChrisBeaumont: Cool. In my mind the priorities are (in decreasing order):

1) Anything which relies on import for auto-discovery (like the collection of available plugins, data viewers, whatever) gets imported by the time it's first used.
2) Such modules are imported as late as possible, so that tasks like  `glue --version`, `from glue import core`, run quickly
3) The two tasks above run if Qt isn't installed

glue/plugins/ginga_viewer/__init__.py
astrofrog: Fixed

astrofrog: I created a centralized logger (in ``glue.logger``) and also added a ``--verbose`` option to the glue command-line script so that we can see INFO messages (default is WARN+)

astrofrog:
For the plugin loading, I've now made it that the plugin has to be explicitly loaded, and I also made it so that ``load_all_plugins`` is called when the glue application is started. This means that people building other apps will need to explicitly load plugins, whereas for the main glue app, all plugins will be loaded by default, which I think is the way we want it. I'll merge this this week if there are no objections with this.

glue/plugins/ginga_viewer/qt_widget.py
astrofrog: Fixed

astrofrog:
The sphinx warning is unrelated and is in master already. I will look into that later this week.

astrofrog:
@ChrisBeaumont - just to check, are you ok with the solution that's here for now? If so I'll go ahead and merge (the travis failure is unrelated and already in master)

ChrisBeaumont:
Fine by me

astrofrog:
Ok, thanks!

PR #590, opened 2015-03-17
Improve loading of plugins

astrofrog:
In #578 and #585, I tried to improve the loading of plugins so that the loading itself is done from the plugin directories, and the only thing outside is that ``GlueApplication`` calls ``load_all_plugins``. I'm still not quite happy with it, so this issue is meant to be a reminder that we need to improve this a little.

In particular, for example the link helpers in #578 can get registered just by importing the plugins directory, whereas the ginga plugin *has* to be explicitly loaded, which doesn't seem very consistent.

### DIFF
--- a/glue/config.py
+++ b/glue/config.py
@@ -56,12 +56,14 @@ class Registry(object):
         return []
 
     def add(self, value):
-        """ Add a new item to the registry """
+        """
+        Add a new item to the registry.
+        """
         self._members.append(value)
 
-    def add_plugin(self, value):
+    def lazy_add(self, value):
         """
-        Add a plugin that will populate this registry
+        Add a reference to a plugin which will be loaded when needed.
         """
         self._plugins.append(value)
 
--- a/glue/config.py
+++ b/glue/config.py
@@ -249,10 +249,12 @@ class QtToolRegistry(Registry):
 
     def __init__(self):
         self._members = {}
+        self._lazy_members = []
         self._loaded = False
 
     @property
     def members(self):
+        self._load_lazy_members()
         if not self._loaded:
             defaults = self.default_members()
             for key in defaults:
--- a/glue/config.py
+++ b/glue/config.py
@@ -34,7 +34,7 @@ class Registry(object):
 
     def __init__(self):
         self._members = []
-        self._plugins = []
+        self._lazy_members = []
         self._loaded = False
 
     @property
@@ -42,7 +42,7 @@ class Registry(object):
         """ A list of the members in the registry.
         The return value is a list. The contents of the list
         are specified in each subclass"""
-        self._load_plugins()
+        self._load_lazy_members()
         if not self._loaded:
             self._members = self.default_members() + self._members
             self._loaded = True
@@ -64,12 +64,12 @@ class Registry(object):
         """
         Add a reference to a plugin which will be loaded when needed.
         """
-        self._plugins.append(value)
+        self._lazy_members.append(value)
 
-    def _load_plugins(self):
+    def _load_lazy_members(self):
         from .plugins import load_plugin
-        while self._plugins:
-            plugin = self._plugins.pop()
+        while self._lazy_members:
+            plugin = self._lazy_members.pop()
             load_plugin(plugin)
 
     def __iter__(self):
--- a/glue/config.py
+++ b/glue/config.py
@@ -11,11 +11,10 @@ Objects used to configure Glue at runtime.
 
 __all__ = ['Registry', 'SettingRegistry', 'ExporterRegistry',
            'ColormapRegistry', 'DataFactoryRegistry', 'QtClientRegistry',
-           'LinkFunctionRegistry', 'LinkHelperRegistry',
-           'ProfileFitterRegistry',
+           'LinkFunctionRegistry', 'LinkHelperRegistry', 'QtToolRegistry',
+           'SingleSubsetLayerActionRegistry', 'ProfileFitterRegistry',
            'qt_client', 'data_factory', 'link_function', 'link_helper',
-           'colormaps',
-           'exporters', 'settings', 'fit_plugin', 'auto_refresh']
+           'colormaps', 'exporters', 'settings', 'fit_plugin', 'auto_refresh']
 
 
 class Registry(object):
--- a/glue/config.py
+++ b/glue/config.py
@@ -35,6 +35,7 @@ class Registry(object):
 
     def __init__(self):
         self._members = []
+        self._plugins = []
         self._loaded = False
 
     @property
@@ -42,6 +43,7 @@ class Registry(object):
         """ A list of the members in the registry.
         The return value is a list. The contents of the list
         are specified in each subclass"""
+        self._load_plugins()
         if not self._loaded:
             self._members = self.default_members() + self._members
             self._loaded = True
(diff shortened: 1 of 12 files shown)


astrofrog:
I've attached pseudo code here to demonstrate what I think is my preferred option for now.

ChrisBeaumont:
Agreed that this should be standardized somehow.

If at all possible we should prevent the user or plugin developer from having to worry about import logic. The registry definitely feels like the right place for all hooks to live, and ideally the registry API is a single hook (ie the decorator/`add` method). 

I wonder if each registry group should define when it should be loaded?

astrofrog:
@ChrisBeaumont - see attached code, where the plugins are added on a per-registry basis, and each registry loads its own plugins only when ``members`` is accessed. Is that what you meant by *each registry group should define when it should be loaded*?

astrofrog:
This works nicely by the way - the ginga widget is only loaded at the point of dragging a dataset into the main window! The exporters are loaded sooner because that registry must be accessed sooner.

astrofrog:
However, the plugins are *registered* before the user's ``config.py`` is parsed, which allows the user to remove any plugins if needed.

astrofrog:
One final thing - from the plugin developer's point of view, they just need to define a ``load_plugin`` function that adds things to the registry as normal, and then also add an entry in ``register_plugins`` if the plugin is going in ``glue.plugins``, otherwise when using the plugin they just need to register it (we can provide docs for this).

glue/plugins/__init__.py
```diff
@@ -3,10 +3,17 @@
 from . import ginga_viewer
 
 
-def load_all_plugins():
+def register_plugins():
+    from ..config import qt_client, exporters
+    qt_client.add_plugin('glue.plugins.ginga_viewer')
+    exporters.add_plugin('glue.plugins.export_d3po')
+    exporters.add_plugin('glue.plugins.export_plotly')
```
ChrisBeaumont: So is the idea that someone who has their own plugin would manually call the appropriate `registry.add_plugin` from their `config.py`?

glue/plugins/export_plotly.py
```diff
@@ -305,6 +304,10 @@ def save_plotly(application, label):
     plotly.sign_in(user, apikey)
     plotly.plot(*args, **kwargs)
 
-exporters.add('Plotly', save_plotly, can_save_plotly, outmode='label')
-settings.add('PLOTLY_USER', 'Glue')
-settings.add('PLOTLY_APIKEY', 't24aweai14')
+def load_plugin():
+    from ..logger import logger
+    from ..config import exporters, settings
+    exporters.add('Plotly', save_plotly, can_save_plotly, outmode='label')
```
ChrisBeaumont: Can you help me understand the distinction between adding an item to a registry, calling `add_plugin` on a registry, and calling `load_plugin`? 

Or a similar question, what is the difference between a plugin and an item in a registry?

glue/plugins/__init__.py
```diff
@@ -3,10 +3,17 @@
...
     """
-    Load built-in plugins
+    Load plugin referred to by name 'plugin'
     """
-
-    from .ginga_viewer import load_ginga_viewer_plugin
-    load_ginga_viewer_plugin()
+    import importlib
+    module = importlib.import_module(plugin)
+    module.load_plugin()
```
ChrisBeaumont: Probably best to give an informative error message here if the plugin-author didn't create a `load_plugin` function.

glue/plugins/__init__.py
astrofrog: In this approach, yes, or they could directly load it (see below)

astrofrog:
**Adding an item to a registry** using ``add`` is the same as it was previously, and it just adding normal entries.

**Registering a plugin ``add_plugin``** tells the registry that there is a plugin that needs to be loaded in order for the ``members`` attribute to be complete. The first time ``members`` is accessed, any plugins are loaded (see next item). In that sense, a plugin entry in the registry is a pointer to deferred registry entries.

**Loading a plugin** means actually adding the proper entries to a registry (note that when the registries load plugins, they remove the *plugin* entry, so they are loaded only once)

Registries can contain two kinds of things - normal members entries (in ``_members``) and plugins (in ``_plugins``). Plugins are always just strings pointing to a sub-package. The first time ``members`` is accessed, all plugins in ``_plugins`` are loaded, ``_plugins`` is emptied, and loading the plugins adds entries to ``_members``.

To summarize what happens inside glue: as soon as glue is started up, we *register* all built-in plugins internally using ``add_plugin``. Glue then loads the user's ``config.py`` file which can for example remove plugins from the registries (still need a proper mechanism for that that doesn't involve touching ``_plugins`` since it is private - maybe I can just make it public). They can also register their own plugins. Then, in glue, if you access ``qt_tools.members``, it will then first load all plugins registered with ``qt_tools``.

In fact, what I've ended up doing is very close to what sphinx does:

* Add plugins by name in:

https://github.com/glue-viz/glue/blob/master/doc/conf.py#L32

(equivalent to what I call registering the plugin)

* Each plugin defines a ``setup`` function that does the actual set-up and is called internally in Sphinx:

https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/sphinx/ext/changelog_links.py#L74

The difference is that in Sphinx, all plugins are loaded at the start. If we don't want that to happen, then we need to make them be loaded as needed, which is why plugins need to be registered on a per-registry basis.

However, the terminology is confusing (registering plugins with registries...) so an alternative API is to simply have registries have a ``plugins`` attribute which is a list, and users can append, insert, or remove form that list. So then one could do:

```
from glue.config import qt_tools
qt_tools.plugins.append('my.nice.qt_tool')
```

What do you think? The alternative would be to have a single list of plugins like sphinx, so e.g.:

```
from glue.config import plugins
plugins.append('my.nice.qt_tool')
```

but then it would be hard to determine when plugins should be loaded and we might end up importing qt straight up.

However, we could allow both - that is, have a general purpose plugins registry for when users don't care if it's loaded straight away, and then allow plugins to also be added on a per-registry basis. Then we get the best of both world.

What do you think?

ChrisBeaumont:
> In that sense, a plugin entry in the registry is a pointer to deferred registry entries.

This sounds like the essential distinction between a normal registry item and a plugin -- the plugin mechanism exists solely to add things to a registry in the future.

I think the names `plugin` and `registry` are confusing, if this is the main difference between them (to my casual ears plugin and registry sound like synonyms, so it's hard to keep the distinction straight). Maybe `add_plugin` should be called something else (`lazy_add`, maybe?). Then a "plugin" is more clearly just an alternative, more advanced way of adding things to a registry. What do you think?

I like using `setup` instead of `load_plugin`. One other alternative, instead of relying on a naming convention, the string could just be a path to the function to run: `my.nice.qt_tool.setup`.

Once we decide on an API, this feature should be added to the sphinx docs as well.

astrofrog:
@ChrisBeaumont - I agree with all your points. It's possible to write plug-in functionality and load it immediately not using lazy loading, so I agree the terminology was wrong. I have renamed ``add_plugin`` to ``lazy_add`` and ``load_plugin`` to ``setup``. I decided to stay with requiring the function to set it up should be ``setup`` (might as well settle on a standard, right?). I've now also added sections to the ``customization.rst`` file. Let me know what you think!

Note that currently I still need to make the spectrum tool and PV slicing tool use this. As a side note, should the registry of models to fit live in the spectrum tool plugin?

glue/config.py
```diff
@@ -35,6 +34,7 @@ class Registry(object):
 
     def __init__(self):
         self._members = []
+        self._plugins = []
```
ChrisBeaumont: might as well call this something like `lazy_members`

astrofrog: Actually, should this then just be a public attribute and we get rid of ``lazy_add``, so that users can append but also remove from ``lazy_members``?

ChrisBeaumont: Oh sorry -- I meant to type `_lazy_members`. I don't have any opinions on whether it should be public or private though.

ChrisBeaumont:
Much clearer to me know! looks good

astrofrog:
Thanks for the review! I've now moved the PV slicer and Spectrum tool to use this too, so if it passes, I'll go ahead and merge :)

PR #598, opened 2015-03-27
Parse HDUList objects from qglue

### LINKED ISSUE #532: Allow data_factories to return hdulist objects

ChrisBeaumont:
Closes #532. This allows users to pass HDUList objects into qglue, which will be packed into data objects (with WCS-parsed coordinate information). This also allows data factories to return HDUList objects, since they are parsed using the same logic.

### DIFF
--- a/glue/core/io.py
+++ b/glue/core/io.py
@@ -1,18 +1,21 @@
 from __future__ import absolute_import, division, print_function
 
 
-def extract_data_fits(filename, use_hdu='all'):
-    '''
-    Extract non-tabular HDUs from a FITS file. If `use_hdu` is 'all', then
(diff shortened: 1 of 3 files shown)


astrofrog:
We might want to make this more general in future, for example the user may want to pass an HDUList with several tables and want each of these to be passed as a new dataset? But for now I think this is ok. I guess an intermediate step is to check if anything doesn't have the same shape and add them as independent datasets?

ChrisBeaumont:
Yeah a future graphical importer would make this more flexible. As it stands now it does the same thing that loading a fits file does, so it just simplifies the process of passing opened fits files into `qglue`

PR #599, opened 2015-03-29
Allow load_ui to load local files

astrofrog:
With this approach, ``load_ui`` first checks if the file exists at the path specified, and if not, falls back to the one in the global glue directory. This allows ``load_ui`` to be used by external plugins.

Another option is to add a keyword argument that says ``in_global_ui``, defaults to ``False``, and gets set to ``True`` everywhere ``load_ui`` is used inside glue. That was the approach in https://github.com/glue-viz/glue/pull/552.

@ChrisBeaumont - what do you think?

### DIFF
--- a/glue/qt/qtutil.py
+++ b/glue/qt/qtutil.py
@@ -725,9 +725,12 @@ def _load_ui_pyqt4(path, parent):
     return loadUi(path, parent)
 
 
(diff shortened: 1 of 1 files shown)


ChrisBeaumont:
+1

astrofrog:
Ok, merging!

PR #600, opened 2015-03-31
Scatter Plot Grouping Plugin

aak65:
I gave the scatter grouping plugin a shot, but haven't been able to get it to work completely yet. I was hoping you guys may be able to point me in the right direction.

I put everything in a plugin folder, but some parts of it (such as the util.py and the ui file) may belong better somewhere else. The issue I couldn't figure out is how to get the custom ui file to load properly. The extra grouping combo box isn't being seen by QWidget (line 324 in scattergroupwidget.ui).

Also, I put this in my config file:

```
from glue.config import qt_client
from glue.plugins.scatter_group.qt_widget import ScatterGroupWidget
qt_client.add(ScatterGroupWidget)
```

ps. I'm putting this up as a separate PR because I worked off a fresh branch on this, and I didn't want to mess up the comments and history on the 'grouping' branch.


astrofrog:
@aak65 - I'll take a look at this now. I *think* the ui issue might be solved by https://github.com/glue-viz/glue/pull/599

astrofrog:
Ok, so the main issue here seems to be that ``ScatterWidget.__init__`` calls ``_connect``, but your ``_connect`` requires the ui file to be read in already. By that point, the ``ScatterWidget`` has loaded its own ui file, so it doesn't work.

At the moment you are overloading all of the stuff in ``ScatterWidget.__init__`` anyway, so one quick workaround is to replace the super call with:

```
        super(ScatterWidget, self).__init__(session, parent)
```

so that it will skip the code in ``ScatterWidget.__init__`` itself.

Another thing is that the load_ui call should be given the full path to the file, so:

```
        self.ui = load_ui(os.path.join(os.path.dirname(__file__), 'scattergroupwidget.ui'))
```

Finally, you should only inherit from ``ScatterWidget``, not ``DataViewer``, since ``ScatterWidget`` already inherits from ``DataViewer``.

Now obviously, changing the super call is not the proper solution, and a better solution would be that the the ui file is loaded in a method that you can overload, and you then won't need to re-define ``__init__``. So you could create a ``def load_ui(self)`` method on ``ScatterWidget`` then overload just that from your subclass. Similarly, I think the only other thing that changes is the code to set up the client, so you could also make that its own method ``_setup_client`` (basically we should not have to overload ``__init__`` directly.

Now for some reason even with the hacks above the widget to edit the properties is not showing in the left panel, so I can look into that later. But at least it will no longer crash.

@ChrisBeaumont - do you agree that the solution is to have ``ScatterWidget.__init__`` call methods rather than do everything itself?

astrofrog:
Ah, it's an easy fix for the options not appearing - the call to ``load_ui`` should include ``self.option_widget`` to tell it where the options should appear:

```
        self.ui = load_ui(os.path.join(os.path.dirname(__file__), 'scattergroupwidget.ui'), self.option_widget)
```

Ok, so to get things to work in the short term:

* Change the ``super`` call to:

```
        super(ScatterWidget, self).__init__(session, parent)
```

* Fix the ``self.ui = `` line as shown above

* Don't inherit from ``DataViewer``

Once you have that working, you can then try and make it so you don't overload all of ``__init__``, so in ``ScatterWidget``, define two new methods, ``setup_ui`` and ``setup_client`` that you can then overload in your sub-class, and then no longer overload ``__init__`` directly.

aak65:
@astrofrog Thank you for that explanation! I removed the `__init__` overloading in `client` & `qt_widget`, defined those methods you suggested, and the plugin seems to be working fine.

This functionality is awesome and I think we can definitely make use of it for a few 'bio-glue' visualization plugins.

glue/plugins/scatter_group/__init__.py
```diff
@@ -0,0 +1,10 @@
+def setup():
+    from ...logger import logger
```
ChrisBeaumont: @astrofrog this logger module is new to me -- is that something you added? Before I was calling logging.getLogger(__name__) to organize loggers into per-module hierarchies. Is that not needed anymore?

glue/plugins/scatter_group/client.py
```diff
@@ -0,0 +1,239 @@
...
+    A client class that uses matplotlib to visualize tables as scatter plots.
+    """
+    xmin = CallbackProperty(0)
+    xmax = CallbackProperty(1)
+    ymin = CallbackProperty(0)
+    ymax = CallbackProperty(1)
+    ylog = CallbackProperty(False)
+    xlog = CallbackProperty(False)
+    yflip = CallbackProperty(False)
+    xflip = CallbackProperty(False)
```
ChrisBeaumont: You probably don't need to redefine most of these, since ScatterClient has them already

ChrisBeaumont:
> @ChrisBeaumont - do you agree that the solution is to have ScatterWidget.__init__ call methods rather than do everything itself?

Sounds right to me @astrofrog

glue/plugins/scatter_group/client.py
```diff
@@ -0,0 +1,239 @@
...
+        add_callback(self, 'xflip', self._set_limits)
+        add_callback(self, 'yflip', self._set_limits)
+        add_callback(self, 'xmin', self._set_limits)
+        add_callback(self, 'xmax', self._set_limits)
+        add_callback(self, 'ymin', self._set_limits)
+        add_callback(self, 'ymax', self._set_limits)
+        add_callback(self, 'xatt', partial(self._set_xydata, 'x'))
+        add_callback(self, 'yatt', partial(self._set_xydata, 'y'))
+        add_callback(self, 'gatt', partial(self._set_xydata, 'g'))
+        add_callback(self, 'jitter', self._jitter)
```
ChrisBeaumont: Likewise, you can call super() here and just add the callback to gatt

glue/plugins/scatter_group/client.py
```diff
@@ -0,0 +1,239 @@
...
+
+        :param layer: the layer to add
+        :type layer: :class:`~glue.core.data.Data` or :class:`~glue.core.subset.Subset`
+        """
+        if layer.data not in self.data:
+            raise TypeError("Layer not in data collection")
+        if layer in self.artists:
+            return self.artists[layer][0]
+
+        result = ScatterGroupLayerArtist(layer, self.axes)
```
ChrisBeaumont: Hmm this is the only changed line in this function, which suggests theres a better way to organize the original class. What about changing the main ScatterClient.add_layer to something like `result = self.layer_artist_class(layer, self.axes)`, setting `layer_artist_class=ScatterLayerArtist` at the class level in the main class, and `layer_artist_class=ScatterGroupLayerArtist` in your class?

glue/plugins/scatter_group/client.py
```diff
@@ -0,0 +1,239 @@
...
+
+        self._update_axis_labels()
+        self._pull_properties()
+        self._redraw()
+
+    def restore_layers(self, layers, context):
+        """ Re-generate a list of plot layers from a glue-serialized list"""
+        for l in layers:
+            cls = lookup_class(l.pop('_type'))
+            if cls != ScatterGroupLayerArtist:
```
ChrisBeaumont: Same comment here, if we lookup which class to compare to using an attribute, then we don't need to modify this function in your new viewer

glue/plugins/scatter_group/client.py
```diff
@@ -0,0 +1,239 @@
...
+                raise ValueError("Scatter client cannot restore layer of type "
+                                 "%s" % cls)
+            props = dict((k, context.object(v)) for k, v in l.items())
+            layer = self.add_layer(props['layer'])
+            layer.properties = props
+
+    def _update_layer(self, layer, force=False):
+        """ Update both the style and data for the requested layer"""
+        if self.xatt is None or self.yatt is None:
```
ChrisBeaumont: Maybe this method should be 

```python
for art in self.artists.get(layer, []):
    art.gatt = self.gatt
super(...)
```
to minimize code duplication?

glue/plugins/scatter_group/client.py
```diff
@@ -0,0 +1,239 @@
...
+        self._layer_updated = True
+        for art in self.artists[layer]:
+            art.xatt = self.xatt
+            art.yatt = self.yatt
+            art.gatt = self.gatt
+            art.force_update() if force else art.update()
+        self._redraw()
+
+    def _on_component_replace(self, msg):
+        old = msg.old
```
ChrisBeaumont: Similar comment here, call super() to avoid code duplication

glue/plugins/scatter_group/client.py
```diff
@@ -0,0 +1,239 @@
...
+            g = self.layer[self.gatt].ravel()
+        except IncompatibleAttribute as exc:
+            self.disable_invalid_attributes(*exc.args)
+            return False
+
+        self.artists = self._axes.plot(x, y, '.')
+
+        df = DataFrame({'g': g, 'x': x, 'y': y})
+        groups = df.groupby('g')
+        if int(len(groups)) < int(len(x)):
```
ChrisBeaumont: Why wrap in int here?

glue/plugins/scatter_group/qt_widget.py
```diff
@@ -0,0 +1,154 @@
...
+    xflip = ButtonProperty('ui.xFlipCheckBox', 'invert the x axis?')
+    yflip = ButtonProperty('ui.yFlipCheckBox', 'invert the y axis?')
+    xmin = FloatLineProperty('ui.xmin', 'Lower x limit of plot')
+    xmax = FloatLineProperty('ui.xmax', 'Upper x limit of plot')
+    ymin = FloatLineProperty('ui.ymin', 'Lower y limit of plot')
+    ymax = FloatLineProperty('ui.ymax', 'Upper y limit of plot')
+    hidden = ButtonProperty('ui.hidden_attributes', 'Show hidden attributes')
+    xatt = CurrentComboProperty('ui.xAxisComboBox',
+                                'Attribute to plot on x axis')
+    yatt = CurrentComboProperty('ui.yAxisComboBox',
```
ChrisBeaumont: Don't need to redefine the repeated properties

glue/plugins/scatter_group/__init__.py
astrofrog: Yes, I changed it when setting up the lazy loading of modules - the issue I ran into is that I think we really only should have one logger for the whole application otherwise I think it's hard to set the logging level globally - or am I wrong?

In other packages, such as Astropy, we have a single logger but it's still able to display in what sub-pacakge the message occurred. Are there other reasons we might want individual loggers?

glue/plugins/scatter_group/qt_widget.py
```diff
@@ -0,0 +1,154 @@
...
+    def _load_ui(self):
+        self.ui = load_ui(os.path.join(os.path.dirname(__file__), 'scattergroupwidget.ui'), self.option_widget)
+
+    def _setup_client(self):
+        self.client = ScatterGroupClient(self._data,
+                                         self.central_widget.canvas.fig,
+                                         artist_container=self._container)
+
+    def _connect(self):
+         ui = self.ui
```
ChrisBeaumont: call super to avoid code duplication (sorry if you are unfamiliar with the `super()` idea! @JudoWill or I can explain it if need be)

glue/plugins/scatter_group/qt_widget.py
```diff
@@ -0,0 +1,154 @@
...
+    hidden = ButtonProperty('ui.hidden_attributes', 'Show hidden attributes')
+    xatt = CurrentComboProperty('ui.xAxisComboBox',
+                                'Attribute to plot on x axis')
+    yatt = CurrentComboProperty('ui.yAxisComboBox',
+                                'Attribute to plot on y axis')
+    gatt = CurrentComboProperty('ui.groupComboBox',
+                                'Attribute to group time series')
+
+    def _load_ui(self):
+        self.ui = load_ui(os.path.join(os.path.dirname(__file__), 'scattergroupwidget.ui'), self.option_widget)
```
ChrisBeaumont: Note to self/@astrofrog -- will `load_ui` work like this if we bundle this in the Mac Binary app? I've had issues in the past

ChrisBeaumont:
Very Cool! 

image

I left a bunch of small comments about using super() to cut down on code duplication, but this is looking great. Since this is bundled as a plugin instead of the main scatter viewer, I'm fine with merging this once its cleaned up. Unless @astrofrog has ideas about whether to put this in this repo or the experimental plugin repo? (I vote that it goes here, until we have an easy GUI way to install plugins from GH within the app). 

Should we add a piece of documentation somewhere, describing how to enable this?

glue/plugins/scatter_group/qt_widget.py
astrofrog: @ChrisBeaumont - right, I need to check that. I had solved that issue as part of https://github.com/glue-viz/glue/pull/552 so I will take a look this evening to see what I had done there.

astrofrog:
@ChrisBeaumont - putting it in the core package plugins is perfectly fine, as it's pretty generic. It would indeed be good to have a section in the docs describing available plugins. I like the idea of having a GUI plugin selector in future!

glue/plugins/scatter_group/qt_widget.py
aak65: Sorry for all the duplication @ChrisBeaumont! I educated myself a bit more about inheritance in Python and trimmed down the code.

glue/plugins/scatter_group/client.py
aak65: Thanks for the suggestion!

astrofrog:
@aak65 - thanks for cleaning this up!

In general, for this kind of plugin, I wonder whether there might be a way to avoid duplicating the ``ui`` file, and whether we could instead adjust the widget once it has been loaded using ``load_ui``. I'd need to look into it more to see how we can support this cleanly, but maybe @ChrisBeaumont has some ideas in the mean time?

ChrisBeaumont:
> I wonder whether there might be a way to avoid duplicating the ui file, 

Yeah so if you wanted to cut down on duplication in the ui file, the strategy would be to override the function that builds the widget, call super() to construct the standard scatterplot widget, and then add your extra UI on top of that. In this case I think it's only a few extra widgets, so you could do that by instantiating widgets in python directly. Or you could have a (very tiny!) ui file that just adds the extra combo box, call load_ui a second time to build that widget, and then insert it into the right spot.

We probably don't want to make the assumption that all plugins can be created in this way

ChrisBeaumont:
Ooo I like that we're failing builds where coverage decreases now :) @aak65 if we're going to put this in the main glue repository, we should add some tests

astrofrog:
I'm going to close this since the scatter widget code has now been completely re-written, but hopefully this would be simpler to implement with the new viewer code.

@aak65 - if you are interested in porting over your code to the new viewers, just email me or reach out to me on Slack (you can get an invite here: https://glueviz-slack-invite.herokuapp.com/)

PR #619, opened 2015-05-12
Write docs on how to make a fully custom Qt widget for glue

astrofrog:
I will do this once #617 is finalized


astrofrog:
@ChrisBeaumont - this is currently a WIP, and the language needs improving, but feel free to already comment on this. Should we explicitly state (as I tried to near the start) that it's best/cleanest to separate the widget functionality from the ``DataViewer`` subclass, which deals with interfacing it with glue?

I wonder whether it would benefit the code inside glue to also follow this approach - and it would make all the existing data viewers usable outside glue (though that may not be very meaningful for some).

astrofrog:
@ChrisBeaumont - does this look ok? If so, I'd like to merge before 0.5. It could be expanded further, but I think it would be good to already have this.

ChrisBeaumont:
Looks good!

PR #636, opened 2015-05-22
Bug in tab name when creating new tab

astrofrog:
To reproduce issue:

* Open glue
* Create extra tab
* Go to tab 2
* Remove tab 1
* Create new tab, which is now called Tab 2

Result: two tabs are called Tab 2


astrofrog:
Code attached

ChrisBeaumont:
+1

PR #637, opened 2015-05-22
Bug when closing tab and pressing cancel

astrofrog:
To reproduce:

* Open glue
* Load dataset, and drag onto main area to open a viewer (don't think it matters which)
* Create new tab, go to new tab
* Click to close Tab 1
* Press cancel in dialog

Result: Tab 1 closes anyway!

### DIFF
--- a/glue/qt/glue_application.py
+++ b/glue/qt/glue_application.py
@@ -2,6 +2,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import os
 import sys
 import webbrowser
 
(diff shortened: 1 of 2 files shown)


astrofrog:
Code attached!

astrofrog:
@ChrisBeaumont cd823c6 is because when showing data viewers such as image viewer with cube, there is a hidden spectrum widget. The other commit is to make sure that when closing a tab there is only a single warning, not 1 + n where n is the number of data viewers.

Issue #650, opened 2015-06-02
Adding Text and Number boxes into custom viewers


JudoWill:
When I'm making custom viewers I'm finding a lot of use-cases where I need to pass arbitrary strings or numbers into the custom viewer. 

I've found the relevant regions in `qt/custom_viewer` code. I know I need to subclass the FormElement but I' having trouble connecting it to the GUI. I know I need to subclass `QtGui.QWidget` to generate the object but I don't understand which QTGui object I need to attach to it.

Any suggestions @ChrisBeaumont ?

ChrisBeaumont:
I think the basic process is:

```
class MyFormElement(FormElement):
    def _build_ui(self):
       w = QLineEdit()
       w.textChanged.connect(nonpartial(self.changed))

    def value(self):
        return self.ui.text()

    @classmethod
    def recognizes(cls, params):
        return params == 'custom_string'

v = data_viewer(foo='custom_string'))
```

The _build_ui returns the widget that the user interacts with, and must connect state changes to the changed method. The value method returns the piece of data that eventually gets passed into functions like update_data. the recognizes method is the hook that says MyFormElement wants to take ownership of the foo='custom_string' keyword.

If you get something to work, it would be great to push some/all of the following into glue:
1) A form to type in numbers
2) A form to type in strings
3) Sphinx documentation about how to create new custom form elements.

astrofrog:
This was mostly addressed in https://github.com/glue-viz/glue/pull/653 but needs documentation

PR #653, opened 2015-06-03
Added new FormElements to represent floats and string input text boxes

JudoWill:
Some commits related to issues #650. @ChrisBeaumont , I'm getting a weird error related to saving the custom viewer, can you take a look and point me in the right direction.

Some doc changes are coming in the next push.

### DIFF
--- a/glue/qt/custom_viewer.py
+++ b/glue/qt/custom_viewer.py
@@ -275,6 +275,7 @@ class MissingSettingError(KeyError):
 class SettingsOracle(SettingsOracleInterface):
 
     def __init__(self, settings, **override):
+
         reserved_words = set(['axes', 'layer', 'self'])
         for key in settings.keys():
             if key in reserved_words:
--- a/glue/qt/custom_viewer.py
+++ b/glue/qt/custom_viewer.py
@@ -115,12 +115,13 @@ class AttributeInfo(np.ndarray):
     """
 
     @classmethod
-    def make(cls, id, values, categories=None):
+    def make(cls, id, values, comp, categories=None):
         values = np.asarray(values)
         result = values.view(AttributeInfo)
         result.id = id
         result.values = values
         result.categories = categories
+        result._component = comp
         return result
 
     @classmethod
@@ -142,14 +143,14 @@ class AttributeInfo(np.ndarray):
         categories = None
         if isinstance(comp, core.data.CategoricalComponent):
             categories = comp._categories
-        return cls.make(cid, values, categories)
+        return cls.make(cid, values, comp, categories)
 
     def __gluestate__(self, context):
         return dict(cid=context.id(self.id))
 
     @classmethod
     def __setgluestate__(cls, rec, context):
-        return cls.make(context.object(rec['cid']), [])
+        return cls.make(context.object(rec['cid']), [], None)
 
 
 class ViewerState(object):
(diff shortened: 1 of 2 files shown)


glue/qt/tests/test_custom_viewer.py
```diff
@@ -9,7 +9,10 @@
 from ...core import Data
 from ...core.subset import SubsetState
 from ...core.tests.util import simple_session
-from ..custom_viewer import FormElement, NumberElement, ChoiceElement, CustomViewer, CustomSubsetState, AttributeInfo
+from ..custom_viewer import FormElement, NumberElement, \
+    ChoiceElement, CustomViewer, \
+    CustomSubsetState, AttributeInfo, \
+    FloatElement, TextBoxElement
```
ChrisBeaumont: nit: prefer using parentheses instead of backslashes

ChrisBeaumont:
Nice -- can you add a few screenshots to this PR?

What is the specific error message you get? For state saving to work you need to implement the state property (which it looks like you attempt to do via the `ValueProperty` descriptor. That descriptor auto-creates a state property that, behind the scenes, calls ui.value()/ui.setValue() -- but my guess is that ui.value() isn't a valid method on a text widget, so instead you need to manually implement the state getter/setter.

glue/qt/custom_viewer.py
```diff
@@ -1091,6 +1090,61 @@ def value(self, layer=None, view=None):
...
+            return None
+
+    def set_value(self, val):
+        self._widget.setText(str(val))
+
+
+    @classmethod
+    def recognizes(cls, params):
+        try:
+            if params.startswith('float'):
```
ChrisBeaumont: We should think about how exactly float and text widgets should be recognized -- what you have might be fine, but it's an important decision. Should we be considering other alternatives to 'float()' and 'str()'?

JudoWill: Yeah, I agree these may not be the best choices. I'm open to suggestions. I was thinking it would also be useful to recognize just floats ... something akin to `a=2.7`.

ChrisBeaumont: Yeah that might be a better option to start with (assuming it doesn't collide with any other FormElement -- I need to check). Likewise the text element could recognize strings (again, in a way that doesn't break the behavior of special strings like 'att')

JudoWill: While I know the idea of the custom_viewers is to make things easier for non-developers; have you thought about allowing the settings definitions to recognize form elements directly? Something like:
`cv = custom_viewer('CV', a=NumberElement(1, 10))`


We could support both pretty easily.

ChrisBeaumont: That's a good idea (I would probably docuement that that in the docstring, but perhaps not sphinx documentation)

JudoWill:
@ChrisBeaumont , I'm putting together a medium-sized plugin using a subclass of the `TextBoxElement` to run a statsmodel formula. Since it requires a little subclassing of many core glue items I was going to write a Sphinx tutorial on all of those pieces. Do you want me to include it as part of this PR or as a different one?

ChrisBeaumont:
If it's equally convenient, 2 PRs is probably easier to review!

JudoWill:
Done, let me spend a little time cleaning up and improving some test coverage and then I'll call this one ready.

JudoWill:
screen shot 2015-06-03 at 5 24 27 pm

Boring-est custom_viewer using the text and number inputs.

astrofrog:
@JudoWill - there is a real failure here with Python 2.6 due to the fact that 2.6 doesn't support the ``{}`` set notation (use ``set([])`` instead). I suspect we'll drop Python 2.6 support soon, but not just yet.

JudoWill:
Fixed py26 issue. Ready to merge.

ChrisBeaumont:
Looks good to me! Thanks @JudoWill

astrofrog:
:+1:

PR #656, opened 2015-06-03
Fixes for recent changes to ginga

ejeschke:
- fixed mode setting API call
- commented out logging to stderr

### DIFF
--- a/glue/plugins/ginga_viewer/qt_widget.py
+++ b/glue/plugins/ginga_viewer/qt_widget.py
@@ -44,20 +44,22 @@ class GingaWidget(ImageWidgetBase):
 
     def __init__(self, session, parent=None):
 
-        self.logger = log.get_logger(name='ginga', log_stderr=True)
+        self.logger = log.get_logger(name='ginga', level=20, null=True,
+                                     # uncomment for debugging
+                                     #log_stderr=True
+                                     )
 
         self.canvas = ImageViewCanvas(self.logger, render='widget')
 
         # prevent widget from grabbing focus
         self.canvas.set_follow_focus(False)
-        self.canvas.enable_overlays(True)
 
         # enable interactive features
         bindings = self.canvas.get_bindings()
         bindings.enable_all(True)
-        self.canvas.set_callback('none-move', self.motion_readout)
-        self.canvas.set_callback('draw-event', self._apply_roi_cb)
-        self.canvas.set_callback('draw-down', self._clear_roi_cb)
+        self.canvas.add_callback('none-move', self.motion_readout)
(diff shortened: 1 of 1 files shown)


astrofrog:
@ejeschke - thanks! could you rebase on the latest master? (just to check if Travis passes now)

astrofrog:
@ejeschke - just to check, does this break compatibility with previous versions of ginga? Can we make it explicit which versions of ginga we support?

ChrisBeaumont:
> Can we make it explicit which versions of ginga we support?

@ejeschke can correct me if I'm wrong but I think the API of ginga evolves pretty rapidly (as does the version number) -- if so we might only be able to support the current latest version. 

The most important thing to ensure is that we use an appropriate version of ginga in the Mac app

astrofrog:
I wonder whether we should consider actually including the ginga plugin for glue in ginga itself, which will allow it to evolve alongside ginga? I suspect that the plugin interface will remain backward-compatible, so this might offer more stability and allow different versions of ginga to work with glue?

ejeschke:
Ok, I will rebase it.   Actually, I don't think the user-facing API changes that radically from version to version.  What I would suggest is tying to the major/minor version numbers and not the release number.

I would suggest for now that the plugin remain in glue--it looks like the plugin was edited quite a bit (at least I don't recognize a lot of the code), but perhaps that all happened at the stage that you refactored the method of handling glue plugins?  But this really all depends on whether you have a method to allow people to distribute plugins that allows them to fully integrate into glue--I'd be very interested in your thoughts because I'm going through some of the same issues, as you can read [here](https://github.com/ejeschke/ginga/issues/92).

astrofrog:
@ejeschke - developing a plugin as a separate package is easy (especially now that the ginga glue plugin code is all self-contained in one directory) and indeed it might be nice to have ``glue.plugins`` namespace packages since it enables ``glue`` to auto-discover installed plugins. I'll do some experiments with that to see if it would be a possibility.

astrofrog:
@ejeschke - just to be clear, we could then still bundle the plugin with glue and even have it auto-install as a dependency but it's more that for development you would have complete control over the plugin code.

astrofrog:
@ejeschke - in any case I agree that for now we should merge this and discuss later whether to move the plugin outside the core glue code. Could you rebase in the next couple of days so that I can go ahead and merge?

ejeschke:
@astrofrog, I have been traveling. Will rebase tomorrow when I get back and update the PR.

ejeschke:
Hi @astrofrog, this PR should be rebased.

astrofrog:
@ejeschke - thanks!

PR #669, opened 2015-06-19
What's new in 0.5?

astrofrog:
This adds a page describing new features and changes in 0.5.


astrofrog:
@ChrisBeaumont - I've started to try and re-organize the front page, and separating things relating to building new viewers or customizing glue from using glue without customization, to make the top-level TOC a little less imposing.

I noticed that http://www.glueviz.org/en/stable/custom_subsets.html could maybe be moved to http://www.glueviz.org/en/latest/python_guide/glue_from_python.html#defining-new-subsets - do you agree?

astrofrog:
I was thinking of also making a new heading about domain-specific documentation, in which we can include the docs for the dendrogram or the spectrum analysis, and where others can include docs about other fields in future. Does this sound reasonable?

ChrisBeaumont:
> I noticed that http://www.glueviz.org/en/stable/custom_subsets.html could maybe be moved to http://www.glueviz.org/en/latest/python_guide/glue_from_python.html#defining-new-subsets - do you agree?

Sounds good to me. I don't give that link out often, so it's fine by me if the old one dies

astrofrog:
@ChrisBeaumont - as usual, I think we can set up a redicted URL (easy to do now), so I can always do this.

astrofrog:
@ChrisBeaumont - does this look ok to you? The main change is that the top-level TOC is now split into sections:

screen shot 2015-07-02 at 9 48 39 pm

ChrisBeaumont:
Yeah this looks good to me. I wonder if for 0.5 we should de-emphasize the mac app option? Anaconda is almost as easy, and much more robust at this point.

chris

astrofrog:
@ChrisBeaumont - I agree. Another downside of the app is that is going to be tricky to install plugins. We could keep the app as being a useful resource for when we want to do a quick demo to someone, or for users who really don't need any customization, but I agree it makes sense to recommend Anaconda.

As a side note, several times I've seen glueviz being one of the recommended apps when using the Anaconda launcher, which is pretty cool!

astrofrog:
@ChrisBeaumont - I've re-organized the installation instructions to put Anaconda first. I need to try out that and the Canopy instructions to make sure they work as described.

doc/installation.rst
```diff
@@ -3,58 +3,66 @@
...
-Easy Option (Recommended for Windows, Unix users)
--------------------------------------------------
+Once Anaconda or Miniconda are `installed
+<http://continuum.io/downloads.html>`_ for your operating system, you
+can install glue using::
 
-We recommend using the `Anaconda
-<http://continuum.io/downloads.html>`_ Python distribution from
-Continuum Analytics. Anaconda includes all of Glue's main dependencies.
+    conda install glueviz
```
ChrisBeaumont: sidenote: `conda install glueviz` currently fails on python 3. I assume that's because conda still thinks glue isn't python-3 compatible? We should:

 * Add python 3 to the classifiers list on setup.py
 * Tell someone from Continuum that they can relax that restriction

astrofrog: Yes, good point. I can do both of those things.

astrofrog: I've added the classifiers in 3eaad03efea7c051c1d0fbd99c132cf968a7d129. I'll be in touch with the Continuum folk about Python 3 and other dependencies (https://github.com/glue-viz/glue/issues/683)

PR #671, opened 2015-06-19
Use latest version of qt-helpers

astrofrog:
The only potentially controversial change here is that before, when importing glue, an application instance was started and was given the glue icon. However, now the icon is set only when starting up the *actual* glue application. This means that other packages importing from Glue won't automatically be branded with the glue icon, which I think makes sense (but note that importing glue still starts up the QApplication instance).


astrofrog:
By the way, this is also a first step to getting glue working with PyQt5

astrofrog:
@ChrisBeaumont - does this seem ok to you? (mainly the change relating to ``get_qapp``). I removed the tests since they are now run in the qt-helpers repository and weren't being run by default here anyway.

astrofrog:
The coverage stats appear to be erratic because the file I added should be excluded anyway, so I think we can count this as passing.

ChrisBeaumont:
Looks good

PR #677, opened 2015-06-27
Look for plugins via entry points

astrofrog:
This is a WIP, but it shows how glue can load plugins via entry points. The logger doesn't currently work because the level is only set once ``main()`` is executed.

This brings up a more fundamental philosophical question which is whether we should be doing all this in ``__init__.py`` (short answer: I don't think so). This means that if anyone writes a package that imports something from glue, all plugins will be loaded, and some may take a while. We could instead do all this (loading internal plugins, loading external plugins, and loading user configuration) when the application is started instead.

However, if someone wants to use glue from a Python script but needs to use one of the plugins (which somehow enhances command-line usage) then this won't necessarily work for them.

As a compromise, we could put the loading of plugins and of the configuration in a function that gets executed when ``GlueApplication`` is instantiated, and which users can also call from Python scripts if needed:

```python
import glue
glue.load_config()
glue.load_plugins()
```

Yes, this is a bit more manual for the case where someone wants to use glue from Python without launching the application and needs access to plugins, but that seems like a corner case?

@ChrisBeaumont - what do you think about this idea?

### DIFF
--- a/glue/__init__.py
+++ b/glue/__init__.py
@@ -118,9 +118,37 @@ def custom_viewer(name, **kwargs):
     from .qt.custom_viewer import CustomViewer
     return CustomViewer.create_new_subclass(name, **kwargs)
 
+from .logger import logger
+
 # Register default plugins (but don't load them)
 from .plugins import register_plugins
+logger.info("Registering built-in plugins")
 register_plugins()
+
+# Search for plugins installed via entry_points. Basically, any package can
+# define plugins for glue, and needs to define an entry point using the
+# following format:
+#
+# entry_points = """
+# [glue.plugins]
+# webcam_importer=glue_exp.importers.webcam:setup
+# vizier_importer=glue_exp.importers.vizier:setup
+# dataverse_importer=glue_exp.importers.dataverse:setup
+# """
+#
+# where ``setup`` is a function that does whatever is needed to set up the
+# plugin, such as add items to various registries.
+
+logger.info("Loading external plugins")
+from pkg_resources import iter_entry_points
+for item in iter_entry_points(group='glue.plugins', name=None):
+    try:
+        function = item.resolve()
+        function()
+    except Exception as exc:
+        logger.info("Loading plugin: {0} failed (Exception: {1})".format(item.name, exc))
+    else:
+        logger.info("Loading plugin: {0} succeeded".format(item.name))
 
 # Load user's configuration file
 from .config import load_configuration
--- a/glue/__init__.py
+++ b/glue/__init__.py
@@ -120,36 +120,6 @@ def custom_viewer(name, **kwargs):
 
 from .logger import logger
 
-# Register default plugins (but don't load them)
-from .plugins import register_plugins
-logger.info("Registering built-in plugins")
-register_plugins()
-
-# Search for plugins installed via entry_points. Basically, any package can
-# define plugins for glue, and needs to define an entry point using the
-# following format:
-#
-# entry_points = """
-# [glue.plugins]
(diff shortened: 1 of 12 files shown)


astrofrog:
I've updated the glue-exp repository to define the entry points to demonstrate how it would work:

https://github.com/glue-viz/glue-exp/blob/master/setup.py#L11

ChrisBeaumont:
This change sounds fine as long as it plays nicely with the mac app. 

I like your compromise solution, of packaging loading logic into 1-2 functions, and calling from GlueApplication

astrofrog:
I decided to leave the configuration loading in ``__init__.py`` for now since a lot of things rely on that (and more specifically ``env``). However, I've now moved the plugin loading to a function in ``glue.main`` and it gets called when starting up ``GlueApplication``.

astrofrog:
I do want to address the config loading in future, but just in a separate PR.

astrofrog:
Note to self: I need to prevent plugins from being loaded twice.

astrofrog:
For now, the Mac app doesn't see the entry points. Looking into it.

astrofrog:
Bummer: https://bitbucket.org/ronaldoussoren/py2app/issue/1/py2app-doesnt-support-egg-metadata

I still think entry points are the way to go, so will try and see if there are ways to work around this.

astrofrog:
Note to self: look into how easy it would be to use http://www.sveinbjorn.org/platypus instead of py2app if the worst comes to the worst

astrofrog:
This now works and requires the following patch in the mac builder repository:

https://github.com/glue-viz/Travis-MacGlue/commit/a7ae36cc4b7204eae2284e356bb8795dffcea37d

The issue is that py2app doesn't know about entry points, so what we instead do is use a wrapper function called ``iter_plugin_entry_points`` that iterates over the entry points for plugins, and in the glue recipe for py2app we patch this to hard-code the entry points in the app. In the non-app version, this function loads plugins dynamically.

astrofrog:
Just for the record, this is the patched version of the function:

```
from pkg_resources import EntryPoint
def iter_plugin_entry_points():
    yield EntryPoint('ginga_viewer', 'glue.plugins.ginga_viewer', attrs=('setup',))
    yield EntryPoint('export_plotly', 'glue.plugins.export_plotly', attrs=('setup',))
    yield EntryPoint('export_d3po', 'glue.plugins.export_d3po', attrs=('setup',))
    yield EntryPoint('pv_slicer', 'glue.plugins.tools.pv_slicer', attrs=('setup',))
    yield EntryPoint('spectrum_tool', 'glue.plugins.tools.spectrum_tool', attrs=('setup',))
    yield EntryPoint('coordinate_helpers', 'glue.plugins.coordinate_helpers', attrs=('setup',))
from glue import plugin_helpers
plugin_helpers.iter_plugin_entry_points = iter_plugin_entry_points
```

at the moment. If we add plugins to the built-in glue, these will get automatically added to the app in future versions of the app. If we want to add external plugins to the app, we have to remember to add them to the ``glue_recipe.py`` file in the Mac builder.

astrofrog:
Ok, I'm pretty happy with this, and I'd like to include it in 0.5 so that it will already understand plugins. The entry_points mechanism applied to plugins is pretty cool, and I wish I'd known about it sooner!

@ChrisBeaumont - since you ok-ed this before modulo the Mac app issue, and that this is now resolved, I'll go ahead and merge, but feel free to leave comments on this and I can fix post-merge!

astrofrog:
Just FYI, the latest 'master' build here: http://mac.glueviz.org/ works fine (basically the export plugins are there)

PR #682, opened 2015-07-02
Implemented plugin manager

astrofrog:
A simple GUI will follow (in this PR). Plugins are enabled unless explicitly disabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,7 +95,7 @@ before_install:
   # now set up a plugin configuration file with some plugins disabled. This is
   # to make sure that when we run the tests, glue will ignore this
   # configuration and will instead run tests for all plugins.
-  - if [ NO_CFG_FILES = false ]; then mkdir ~/.glue; printf "[plugins]\nspectrum_tool = 0\n" >> ~/.glue/plugins.cfg; fi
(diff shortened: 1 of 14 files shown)


astrofrog:
This now implements a GUI manager for plugins:

screen shot 2015-07-05 at 11 08 23 pm

Users need to restart Glue for changes to take effect, but otherwise it works as expected.

astrofrog:
@ChrisBeaumont - this is ready for review if you are interested

ChrisBeaumont:
Looks good to me!

astrofrog:
I think I'll need to add some tests for this, will do this week.

astrofrog:
This now includes unit tests, and also implements a way to override the config dir during tests (in fact, it's overridden by default by a temp directory and can be further overridden by individual tests)

astrofrog:
Ok, this is ready for a proper review.

One open question is how to deal with plugins and session files. If I use a plugin and save a session which contains a reference to that plugin, then load the session up, should it load the required plugins even if the required plugins were disabled before? Should it give an error? What about plugins that aren't even installed?

astrofrog:
@ChrisBeaumont - would you like to review this before I merge? Do you have any thoughts about the question of saving sessions that use plugins?

ChrisBeaumont:
This looks good to me. Can we attempt to load saved plugins, with graceful
fallback if they aren't available?

astrofrog:
For the loading of plugins when restoring sessions, I've opened this issue: https://github.com/glue-viz/glue/issues/721 (since this is an issue that is independent on this PR)

Issue #688, opened 2015-07-03
Enable/disable selection dragging?


astrofrog:
At the moment, if I select a region then want to select a smaller region inside, I can't do this because it will just drag the ROI rather than drawing a new one. I wonder if selection dragging should be its own selection mode in the selection toolbar?

cc @stscieisenhamer @ChrisBeaumont

astrofrog:
Since the current change in behavior may be confusing for existing users, unless I hear any objections I will add a new selection mode for dragging selections but will leave 'replace' as the default mode. I would like to squeeze this into 0.5.1 in the next couple of days.

astrofrog:
@stscieisenhamer @ChrisBeaumont - could you review #689?

PR #689, opened 2015-07-03
Added new selection ‘Move’ mode

astrofrog:
Fixes https://github.com/glue-viz/glue/issues/688 by adding a dedicated 'Move' mode for selections which is not enabled by default.

### DIFF
--- a/doc/getting_started/index.rst
+++ b/doc/getting_started/index.rst
@@ -55,11 +55,13 @@ We can highlight the west arm of W5 using the rectangle selector:
 
 Notice that this highlights the relevant pixels in the image, adds a new subset (which we've named ``west arm``) to the data manager, and adds a new visualization layer (also labeled ``west arm (w5)``) in the visualization dashboard.
 
-We can redefine this subset by dragging a new rectangle in the image. Alternately, we could define a second subset by clicking the ``New Subset`` button (next to the folder button).
+We can redefine this subset by dragging a new rectangle in the image, or we can
+also move around the current subset by pressing the 'control' key and clicking
+on the subset then dragging it. Alternately, we could define a second subset by
+clicking the ``New Subset`` button (next to the folder button).
 
 .. _multi_selection_note:
 .. note:: When multiple subsets are defined, only the highlighted entries in the data manager are affected when drawing new subsets. If no subsets are highlighted, then a new subset is created.
-
 
 You can edit the properties of a visualization layer (color, name, etc.) By double-clicking on the entry in the visualization dashboard.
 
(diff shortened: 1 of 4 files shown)


astrofrog:
One issue I'm having is that if I click outside a ROI, then everything works fine but there is a small delay because it re-computes the subset anyway.

@ChrisBeaumont - do you have any suggestions for how to short-circuit things in the case where I click outside the region when in move mode? (if not, I'll dig into the code)

ChrisBeaumont:
The move mode doesn't feel like the other edit subset modes for me for a
few reasons:
 - the logic of updating the subset is the same as the replace mode, and
not a distinct operation
 - the other subset modes define what happens when a new subset state
should be applied. This mode defines details about how mouse behavior works
 - one of the design considerations for subset modes was that they are a
black box as far as the rest of the code base is concerned. This mode
breaks that contract, since this or adds several new branches to the code
base checking the mode state

If the intention is to toggle between scrubbing and redefining subsets,
perhaps this should be controlled by another mechanism. For example,
holding a modifier key like shift while dragging. Or maybe making it so
backspace just clears the current subset

astrofrog:
@ChrisBeaumont - thanks for reviewing this and for the insight. I agree that this is a bit different to a mode. I like the suggestion of having a modifier key for switching to scrubbing as opposed to backspace for clearing selections. I'll look into it.

astrofrog:
@ChrisBeaumont - thanks for your suggestion, I think the code is nicer now. I decided to use the control key for now, since shift, option, and command already have common meanings with selections that we may want to implement in glue some day (i.e. what we were doing in bermuda). Would you be happy with this implementation?

astrofrog:
@ChrisBeaumont - just a small question: why does the circular selection appear to work in pixel coordinates while others appear to work in data coordinates?

ChrisBeaumont:
They should probably all work in pixel coordinates, because a circle on the
screen is not a circle in data space for, eg, projections and log
transforms.

astrofrog:
@ChrisBeaumont - ok I'll open another issue for that. I'll try and find some examples where the effects would be noticeable to try and fix this.

astrofrog:
@ChrisBeaumont - ok, thanks! This PR now works nicely (there was a subtle bug I introdued that was causing what looked like slowness/lagging in the scrubbing) Once this passes, I'll merge it later this evening or tomorrow if there are no objections.

stscieisenhamer:
Only because somewhere I was mentioned, but I was on vacation and then dealing with the resulting work-catchup, but wanted to drop...

:+1:

Issue #676, opened 2015-06-24
Refactor logic branches based on Component type -- use polymorphism instead


ChrisBeaumont:
From #601:

we should refactor logic branches in the viewers that look like:

```
if comp.categorical:
    state = CategoricalRoiSubsetState.from_range(comp, self.component, lo, hi)
else:
  ...
```
And instead add build_range_subset_state methods to the Component classes.

There are some other examples where we have logic branches based on comp.categorical, where we should follow a similar refactor

JudoWill:
Read my mind @ChrisBeaumont

astrofrog:
@JudoWill - did that in https://github.com/glue-viz/glue/pull/692

PR #692, opened 2015-07-05
ROI refactoring

JudoWill:
I've done some refactoring of the ROI logic into the Component classes as suggested in #676 . I've removed all of the logic from the ScatterClient and HistogramClients and it transparently deals with the issues involved with categorical axes.

@ChrisBeaumont, we discussed this before, but dealing with PolygonalROIs with Categorical axes is not intuitive. Currently if a PolyROI is passed to a mixed axis I expand it to its bounding rectangle and then generate everything. This covers all of my use-cases but may be confusing to others. The other option is to only accept RectangularROIs and raise an error otherwise; but without some GUI hinting this may also be confusing. Thoughts?

### DIFF
--- a/glue/clients/histogram_client.py
+++ b/glue/clients/histogram_client.py
@@ -5,6 +5,7 @@ import numpy as np
 from ..core.client import Client
 from ..core import message as msg
 from ..core.data import Data, CategoricalComponent
+from ..core.roi import RangeROI
 from ..core.subset import RangeSubsetState, CategoricalRoiSubsetState
 from ..core.exceptions import IncompatibleDataException, IncompatibleAttribute
 from ..core.edit_subset_mode import EditSubsetMode
@@ -378,16 +379,9 @@ class HistogramClient(Client):
         if self.xlog:
             lo = 10 ** lo
             hi = 10 ** hi
-
-        comp = list(self._get_data_components('x'))
-        if comp:
-            comp = comp[0]
-            if comp.categorical:
-                state = CategoricalRoiSubsetState.from_range(comp, self.component,
-                                                             lo, hi)
-            else:
-                state = RangeSubsetState(lo, hi)
-                state.att = self.component
+        nroi = RangeROI(min=lo, max=hi, orientation='x')
+        for comp in self._get_data_components('x'):
+            state = comp.subset_from_roi(self.component, nroi, coord='x')
             mode = EditSubsetMode()
             visible = [d for d in self.data if self.is_layer_visible(d)]
             focus = visible[0] if len(visible) > 0 else None
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -16,7 +16,7 @@ from .component_link import (ComponentLink, CoordinateComponentLink,
 from .subset import (Subset, InequalitySubsetState, SubsetState,
                      RoiSubsetState, RangeSubsetState,
                      CategoricalRoiSubsetState, AndState)
-from .roi import (PolygonalROI, CategoricalRoi, RangeROI, RectangularROI,
+from .roi import (PolygonalROI, CategoricalROI, RangeROI, RectangularROI,
                     XRangeROI, YRangeROI)
 from .hub import Hub
 from .util import split_component_view, row_lookup
@@ -243,11 +243,18 @@ class Component(object):
         raise NotImplementedError
 
     def subset_from_roi(self, att, roi, other_comp=None, other_att=None, coord='x'):
-        """ Create an SubsetState object
+        """ Create a SubsetState object from an ROI.
+
+        This encapsulates the logic for creating subset states with
+        Components. See the documentation for CategoricalComponents for caveats
+        involved with mixed-type plots.
+
         :param att: attribute name of this Component
         :param roi: an ROI object
         :param other_comp: The other Component for 2D ROIs
         :param other_att: The attribute name of the other Component
+        :param coord: The orientation of this Component
+        :param is_nested: True if this was passed from another Component.
         :return: A SubsetState (or subclass) object
         """
 
@@ -255,6 +262,7 @@ class Component(object):
 
         if hasattr(roi, 'min') and hasattr(roi, 'ori'):
             # RangeROI and its subclasses
+            assert roi.ori in set('xy')
             lo, hi = roi.range()
             if roi.ori == coord:
                 subset_state = RangeSubsetState(lo, hi, att)
@@ -267,7 +275,7 @@ class Component(object):
             elif roi.ori != coord:
                 subset_state = RangeSubsetState(lo, hi, other_att)
             else:
-                raise AssertionError('RangeROI.ori must be "x" or "y"')
+                raise AssertionError
         else:
             if isinstance(other_comp, CategoricalComponent):
                 return other_comp.subset_from_roi(other_att, roi,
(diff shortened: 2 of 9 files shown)


ChrisBeaumont:
> Currently if a PolyROI is passed to a mixed axis I expand it to its bounding rectangle

I think this is the right behavior. IIUC this behavior would always do the same thing as mapping a categorical variable onto an numerical variable (so that each point snaps to an integer) and then applying  the original polygon. 

Can you remind me how we ensure that we apply the ROI consistently when different datasets have different category labels? For instance if I have a histogram of a dataset with categories {A, B, D, E} and I select the {B, D} range, would the label C be selected in a second dataset?

JudoWill:
The way things are currently implemented, if you did a range-selection (including using PolyROI) on a plot, label C would be part of the ROI in the second dataset from your example. It would NOT be a part of the ROI built from the first dataset. This is because I use an interator to generate CategoricalROIs for each dataset on the plot using the range. However, one can make a CategoricalROI that doesn't require categories to be consecutive through other methods (like through the ipython console).

astrofrog:
@JudoWill - side note: the merge conflict is because of changes in the changelog. Can you rebase and move the entry to the v0.6 section?

ChrisBeaumont:
> This is because I use an iterator to generate CategoricalROIs for each dataset on the plot using the range.

Does the behavior of the ROI change depending on which of the two datasets are added to the plot when the ROI is defined? Should we obey the contract that if I draw a shape on a particular coordinate system, then the corresponding ROI behaves the same regardless of which datasets have been plotted to that coordinate system?

JudoWill:
> Should we obey the contract that if I draw a shape on a particular coordinate system, then the corresponding ROI behaves the same regardless of which datasets have been plotted to that coordinate system?

Its not that I disagree, its more label C doesn't exist in the first dateset and as such it wouldn't be added to CategoricalROI ... since C won't even be in the base component it will take some spegettification of the code to make that work correctly.  I'm not sure which is more "intuitive": Adding a label to the ROI that doesn't exist in the dataset (but is in the plot) or not adding a label to the ROI from the plot because it doesn't exist in the dataset. Looking at my general use-cases it doesn't make a difference so I'm happy to implement it either way.

glue/core/data.py
```diff
@@ -238,6 +242,55 @@ def __str__(self):
...
+        :param att: attribute name of this Component
+        :param roi: an ROI object
+        :param other_comp: The other Component for 2D ROIs
+        :param other_att: The attribute name of the other Component
+        :param coord: The orientation of this Component
+        :param is_nested: True if this was passed from another Component.
+        :return: A SubsetState (or subclass) object
+        """
+
+        assert coord in set('xy')
```
ChrisBeaumont: minor nit: converting to set probably adds more overhead than it saves :). Also, this should be a ValueError and not an AssertionError, since it could be user-triggered

ChrisBeaumont:
I'm not sure that I have a strong opinion on whether label C should be included. I just think that it should either be included or not independent of whether or not a dataset is plotted (or even loaded) at the time of ROI definition. 

Thinking out loud about this a bit more:
 - My understanding (correct me if I'm wrong) is that at the moment rois turn into categorical ROIs when necessary, and that CategoricalROIs are backed by a set of labels to include in the ROI. That satisfies the above property, so I'm happy with that
- An alternative is to have something like a 'CategoricalRangeRoi' that's defined not by a set of labels but by a (start, stop) pair. This is only well-defined if the categorical component values are sortable (ie it makes sense if the category labels define a gradient of something, but not for labels like state names). We don't distinguish between these two types of components, so we probably can't implement a CategoricalRangeRoi. Maybe we'd want to in the future.

Anyways, this refactor + the current behavior sounds good to me once the merge conflict is fixed.

glue/clients/histogram_client.py
```diff
@@ -378,16 +379,9 @@ def apply_roi(self, roi):
...
-            comp = comp[0]
-            if comp.categorical:
-                state = CategoricalRoiSubsetState.from_range(comp, self.component,
-                                                             lo, hi)
-            else:
-                state = RangeSubsetState(lo, hi)
-                state.att = self.component
+        nroi = RangeROI(min=lo, max=hi, orientation='x')
+        for comp in self._get_data_components('x'):
+            state = comp.subset_from_roi(self.component, nroi, coord='x')
```
ChrisBeaumont: nice. Will this PR close #676?

glue/core/data.py
```diff
@@ -438,6 +491,52 @@ def _update_data(self):
...
```
ChrisBeaumont: ditto

astrofrog:
Just a quick note on the selection question - I think it basically depends whether you are dealing with nominal or ordinal categorical data. I think there are some datasets where order is not meaningful - for example countries. Furthermore we might want to allow users to change the orders in which the categorical data is displayed. So I personally don't think we should really assume any ordering of categorical data by default. Does this mean that we maybe need two different types of categorical with different ROI behavior?

As I've mentioned in other issues, I think that we need to allow users to customize the type of data beyond what auto-detection can give us. For instance, not all string data is categorical (a string of timestamps). So this could be a further dimension of the data description - is the categorical data ordinal or not? Then different ROI behavior would apply.

JudoWill:
> My understanding (correct me if I'm wrong) is that at the moment rois turn into categorical ROIs when necessary, and that CategoricalROIs are backed by a set of labels to include in the ROI.

Correct. The ROI is constructed from the plot using something like a RangeROI but the actual CategoricalROI (and CategoricalSubset) returned contains the raw labels for lookup. This helps it play nice when applied across other datasets.

> Furthermore we might want to allow users to change the orders in which the categorical data is displayed.

That would be cool. Are you thinking of a GUI type thing some sort of data definition file (or through the `config.py` file). If we could find a way to get a user-defined ordering and attach it to the `CategoricalComponent.categories` variable then all of the plots *should* automagically update.

> Does this mean that we maybe need two different types of categorical with different ROI behavior?

Probably ... my general backdoor solution has always been to make my ordinal categories labeled in lexigraphical order. But you can't really guess this from the data itself, however, it could be done through some clever subclassing. We could make an `OrdinalBaseclass(CategoricalComponent)` that users could subclass to implement themselves much like the custom_viewers.

astrofrog:
@JudoWill - could you rebase and wrap this up in the coming week? Here are some thoughts:

For polygonal and circular ROIs, I think that we should make these work with categorical data - basically for each category, the ROI should be interpolated at the location of those categories in the plot so that the selection is 'intuitive'. Note that for polygonal ROIs, since these don't have to be convex there could be multiple sub-ranges for a given category. Sounds like maybe something that should be implemented in a separate PR and leave this disabled for now?

Regarding missing categories, we can only apply ROIs to missing categories if the order of the categories that aren't plotted is meaningful and I'm not convinced that's always the case. In addition, it's not clear how one would then deal with the circular and polygonal ROIs for missing categories, since one would have to pick a position to interpolate to. The only 'clean' solution would be to always show all the categories in a plot, if a dataset is linked with another that has more categories (then sometimes some of the categories will be empty, but at least that avoids any ambiguity with selection).

For now, I would suggest that we stick with @JudoWill's implementation that the missing label C would not be part of the categorical ROI, and we can then review that later if/once we allow to distinguish between ordinal and nominal categories, or if we adjust the plotting of categorical data to always show all categories for linked data (such that there are never any 'missing' categories from plots). 

@ChrisBeaumont - does that sound like a good plan?

astrofrog:
@JudoWill - I have some time this week to work on glue, so would you like me to rebase this PR and implement my own suggestions?

astrofrog:
Closing in favor of #834 which includes the commits from here, rebased. I'll work on finishing up the remaining comments here in that PR.

astrofrog:
I've now merged this in in #834 - thanks @JudoWill!

PR #696, opened 2015-07-09
Cleaned up __all__ for helpers and add has_extension.

stscieisenhamer:
With regard to data loaders, not sure what changed, but a very short while ago the line

```from glue.core.datafactories import has_extension```

worked fine. Now, with lastest master, it does not. This may be intentional, but I cannot tell and am putting up the PR because the example data loader repo has this all over it.

Delete this if non-sequitur.


astrofrog:
I re-organized the files for ``data_factories`` and indeed missed ``has_extension`` from ``__all__`` - thanks for the fix!

Issue #697, opened 2015-07-09
Spectrum tool and PV Slice tool do not appear in 0.5.1 


keflavich:
In 0.5.1, when I load a cube, the spectrum extraction and PV extraction tools are missing.  They are present for the same data in 0.4.

astrofrog:
Did you install glue with conda or some other way? Also, do you see the 'Export' sub-menu in the 'File' menu, and does it contain 'Export D3PO' and 'Export Plotly'? Finally, can you let me know what the output is in the terminal when you start up glue with ``glue --verbose``?

keflavich:
Installed from git clone with python setup.py develop.  More details later

astrofrog:
Basically the spectrum tool and PV slice thing are now treated as plugins and registered via entry_points. It's possible that entry points don't work properly with 'develop'. Will try it out.

astrofrog:
Hmm no it seems to work. So later, when you get a chance, send the output of ``glue --verbose``

keflavich:
OK, so there's clearly something wrong with entry points that is specific to my setup

```
INFO:glue:Input arguments: ['/Users/adam/anaconda/envs/astropy27/bin/glue', '--verbose', 'ALMA_Outflow_b6_12M_12CO.fits']
INFO:glue.core.hub:Subscribing DataCollection (0 data sets)
	 to DataAddComponentMessage
INFO:glue.core.hub:Broadcasting DataCollectionAddMessage:
	 Sent from: DataCollection (1 data set)
	  0: ALMA_Outflow_b6_12M_12CO
INFO:glue:Loading external plugins
INFO:glue.core.hub:Subscribing <glue.qt.data_collection_model.DataCollectionModel object at 0x113f3b770> to DataCollectionDeleteMessage
INFO:glue.core.hub:Subscribing <glue.qt.data_collection_model.DataCollectionModel object at 0x113f3b770> to SubsetDeleteMessage
INFO:glue.core.hub:Subscribing <glue.qt.data_collection_model.DataCollectionModel object at 0x113f3b770> to DataCollectionAddMessage
INFO:glue.core.hub:Subscribing <glue.qt.data_collection_model.DataCollectionModel object at 0x113f3b770> to SubsetCreateMessage
2015-07-10 01:31:08.429 python[44817:507] modalSession has been exited prematurely - check for a reentrant call to endModalSession:
WARNING:glue.config:could not get default tools for GlueVTKViewer
[...]

PR #699, opened 2015-07-10
Fix plugin loading

astrofrog:
While investigating https://github.com/glue-viz/glue/issues/697, we found that plugins don't get loaded for setuptools < 11.3. This fixes compatibility, and plugin loading now works for setuptools >= 0.6.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@ v0.6 (unreleased)
 v0.5.2 (unreleased)
 -------------------
 
-* No changes yet
+* Fix loading of plugins with setuptools < 11.3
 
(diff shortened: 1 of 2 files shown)


ChrisBeaumont:
Should we warn users at setuptools-import-time if their version is too old? Should we add an old version of setuptools to the travis build matrix?

astrofrog:
@ChrisBeaumont - to be honest I'm having trouble finding a version that's not compatible. I went all the way down to 0.6 and it still worked. Having said that, we could simply make one of the existing builds use setuptools 0.6 for example (no need to add a new build)

PR #703, opened 2015-07-12
Add two functions to __all__ to preserve backward-compatibility

astrofrog:
@keflavich - I noticed in your script you had to hack around this but this was an oversight, not a deliberate API change.


keflavich:
Thanks, :+1:

PR #710, opened 2015-07-24
Allow to specify which extension is used to get the wcs info

saimn:
Hi,
I have a FITS cube which has the WCS info only in the first extension (with the data), whereas glue expect it to be in the primary header. This PR make it possible to load the cube in glue from a python script, specifying the extension in `load_data`. I don't know how it would be possible to have this option in the UI, I didn't look at the UI code yet, but with some guidance I could try to add an option for this.
I also fixed the use of the `use_hdu` keyword which wasn't passed to the function where it is used.

### DIFF
--- a/glue/core/data_factories/gridded.py
+++ b/glue/core/data_factories/gridded.py
@@ -25,12 +25,19 @@ def is_fits(filename):
         return False
 
 
-def gridded_data(filename, format='auto', **kwargs):
+def gridded_data(filename, format='auto', wcs_ext=0, **kwargs):
     """
     Construct an n - dimensional data object from ``filename``. If the
     format cannot be determined from the extension, it can be
     specified using the ``format`` option. Valid formats are 'fits' and
     'hdf5'.
+
+    :param wcs_ext:
+        Specify the number of the extension used to read the wcs info.
+    :param kwargs:
+        Extra arguments passed to the function used to extract data
+        (``extract_data_fits`` or ``extract_data_hdf5``).
+
     """
     result = Data()
 
@@ -42,7 +49,7 @@ def gridded_data(filename, format='auto', **kwargs):
     if is_fits(filename):
         from ...external.astro import fits
         arrays = extract_data_fits(filename, **kwargs)
-        header = fits.getheader(filename)
+        header = fits.getheader(filename, ext=wcs_ext)
         result.coords = coordinates_from_header(header)
     elif is_hdf5(filename):
         arrays = extract_data_hdf5(filename, **kwargs)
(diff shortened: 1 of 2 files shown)


glue/core/data_factories/gridded.py
```diff
@@ -25,12 +25,19 @@ def is_fits(filename):
         return False
 
 
-def gridded_data(filename, format='auto', **kwargs):
+def gridded_data(filename, format='auto', wcs_ext=0, **kwargs):
```
saimn: BTW, the format keyword is not really used in the function: it is just printed if the format is unknown, but the docstring tells it should override the format detection, which is not true.

stscieisenhamer:
@saimn A point of information, we're currently developing a generalized cube loading tool. See https://github.com/spacetelescope/cube-tools

The loader is called STcube and, through configuration, allows configuration of how a cube, along with error and mask information, is read from a FITS file.

Its in a very early stage, but feedback is more than welcome.

saimn:
@stscieisenhamer Nice, I have tested and it works well ! It seems you are also adding some analysis tools ? I didn't found in the UI something related to specview, but maybe it is too early to use this.
I'm quite interested in these kind of analysis tools,  and integration in glue, to look at a source, its spectrum, find a redshift ...

stscieisenhamer:
@saimn Yes, for JWST we're building a whole suite of tools, much of it centered around glue. Feel free to continue any discussion in the cube-tools repo. And yes, things are very, very early, though by the end-of-year, at we're hoping for at least an overall framework will be done.

astrofrog:
@saimn - @stscieisenhamer has now improved the FITS reader, so that it should hopefully now work with your file. Could you try out the latest developer version of glue and let me know if that works?

saimn:
@astrofrog yes, I saw there is a lot of activity on this, I will try the latest version.

saimn:
@astrofrog It works, but I'm not sure about the merge datasets window: I have 3 extensions, DATA, STAT and DQ, it first proposes me to merge STAT and DQ, if I accept it proposes to merge STAT (with what ?), and if I accept again I end up with only the STAT extension, my DATA extension is missing.
If I refuse to merge, then I get my 3 extensions.
Anyway, we can close this PR, I will play a bit more with glue ;-)

astrofrog:
@saimn - can you send me your data file to thomas.robitaille@gmail.com? I'd like to reproduce the steps you were trying to understand better how we can improve it.

saimn:
The file is a bit too big, but you can get a similar one here:
http://data.muse-vlt.eu/HDFS/v1.0/DATACUBE_HDFS_v1p0.fits.gz  (2.6 Gb)
(from http://muse-vlt.eu/science/hdfs-v1-0/ cf the Data Cube section for the meaning of the extensions)

PR #711, opened 2015-07-24
Fix dev version generation in setup.

saimn:
The current commit hash/number detection was broken.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 from __future__ import print_function
 
-from setuptools import setup, Command, find_packages
+from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
 
 import os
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -23,16 +23,17 @@
     command_number = 'git rev-list --count HEAD'
 
     try:
-        commit_hash = subprocess.check_output(command, shell=True).decode('ascii').strip()
-        commit_number = subprocess.check_output(command, shell=True).decode('ascii').strip()
+        commit_hash = subprocess.check_output(command_hash, shell=True)\
+            .decode('ascii').strip()
```
astrofrog: I this this is a case where breaking the line doesn't help readability - could you remove the line break? (also in the line below)

setup.py
```diff
@@ -23,16 +23,17 @@
...
     except Exception:
         pass
     else:
-
         # We write the git hash and value so that they gets frozen if installed
         with open(os.path.join('glue', '_githash.py'), 'w') as f:
-            f.write("__githash__ = \"{githash}\"\n".format(githash=commit_hash))
-            f.write("__dev_value__ = \"{dev_value}\"\n".format(dev_value=commit_number))
+            f.write("__githash__ = \"{}\"\n".format(commit_hash))
+            f.write("__dev_value__ = \"{}\"\n".format(commit_number))
```
astrofrog: These changes aren't really necessary - can you revert them? (also, ``{}`` won't work on Python 2.6)

astrofrog:
I've implemented the minor changes and opened a new pull request (that includes your commit): https://github.com/glue-viz/glue/pull/720

saimn:
Thanks for having made the changes to merge this Thomas, I was in vacation ;-)

setup.py
ChrisBeaumont: I tend to format long method chains like this:

```
commit_hash = (subprocess
               .check_output(command_hash, shell=True)
               .decode('ascii')
               .strip()
)
```

PR #719, opened 2015-08-15
Improve format identification

astrofrog:
At the moment, glue uses a default factory based on filename in order to open files. This can lead to issues in cases where readers define more specialized versions of common formats. For instance, the astrodendro package stores dendrograms in FITS and HDF5 files, but there is already a general FITS and HDF5 reader (which is also the default factory for any file whose name ends in ``hdf5`` or ``fits``).

This PR makes the following changes:

* The dendrogram identifier function now checks more than the filename, and checks that the file structure corresponds to what would be expected from a dendrogram. Note that for the FITS format, the HDUs were missing names, so instead for old files I rely on the layout of the file which is unlikely to mimic a dendrogram file just by chance. I've also opened a PR to add HDU names for the astrodendro package, see https://github.com/dendrograms/astrodendro/pull/144 - and have preemptively also added these names to the logic.

* The format identification to no longer have a 'default', but instead to check with all identifier functions which ones could read the files.

* The data factories now have the concept of 'priority'. For example, the dendrogram FITS/HDF5 loader has a higher priority than the general one, if the file truly is a dendrogram file, because then we should use the more specialized reader.

With these changes, one can do for example ``glue dendro.fits`` and it will correctly open up as a dendrogram dataset.

There are failing tests, so this is still a WIP, but @ChrisBeaumont let me know if you have any thoughts in the mean time about this approach.


glue/config.py
```diff
@@ -264,21 +264,33 @@ def new_factory(file_name):
...
 
-    def __call__(self, label, identifier=None, default=''):
-        from .core.data_factories import set_default_factory
+        def get_priority(fact):
+            try:
+                return fact.priority
+            except AttributeError:
+                return 0
+
+        return [self.item(f, f.label, f.identifier, get_priority(f)) for f in __factories__]
```
ChrisBeaumont: when is priority not defined in a namedtuple?

glue/core/data_factories/dendro_loader.py
```diff
@@ -34,3 +36,69 @@ def load_dendro(file):
...
+
+        from ...external.astro import fits
+
+        hdulist = fits.open(file)
+
+        # In recent versions of Astropy, we could do 'DATA' in hdulist etc. but
+        # this doesn't work with Astropy 0.3, so we use the following method
+        # instead:
+        try:
+            hdulist['DATA']
```
ChrisBeaumont: Minor nit: does this pattern load data from disk? If so, it's an overly expensive test

glue/core/data_factories/gridded.py
```diff
@@ -71,10 +71,8 @@ def is_gridded_data(filename, **kwargs):
 
 gridded_data.label = "FITS/HDF5 Image"
 gridded_data.identifier = is_gridded_data
+gridded_data.priority = 2
```
ChrisBeaumont: Ah I see, we are not always adding 'item' namedtuples to the factory registry, so `priority` may not be defined. Is that the issue?

glue/core/data_factories/helpers.py
```diff
@@ -272,48 +267,52 @@ def data_label(path):
...
-    for func, _, identifier in data_factory:
+    # We no longer try the 'default' factory first because we actually need to
+    # try all identifiers and select the one to use based on the priority. This
+    # allows us to define more specialized loaders take priority over more
+    # general ones. For example, a FITS file that is a dendrogram should be
+    # loaded as a dendrogram, not a plain FITS file.
+
+    valid_formats = []
+
+    for func, label, identifier, priority in data_factory:
```
ChrisBeaumont: Can we iterate over factories in priority order, and return the first one that can read the data?

glue/core/data_factories/helpers.py
```diff
@@ -272,48 +267,52 @@ def data_label(path):
...
+        if is_format:
+
+            valid_formats.append((priority, label, func))
+
+    priorities = list(zip(*valid_formats))[0]
+
+    highest_priority = max(priorities)
+
+    if priorities.count(highest_priority) > 1:
+        raise ValueError("Ambiguous file type")
```
ChrisBeaumont: This seems too extreme. Perhaps we can choose an arbitrary factory and issue a warning?

glue/core/data_factories/gridded.py
astrofrog: Yes, it was for backward-compatibility in case anyone was defining it like us here. But then it's probably true that most other people are using the decorator (in which case ``priority`` does have a default value). Maybe we should in fact also use the decorator internally, then ``priority`` would always be defined.

glue/core/data_factories/dendro_loader.py
astrofrog: I don't *think* so but I'll do some tests.

glue/core/data_factories/helpers.py
astrofrog: Yes, good idea. In fact if we want to know if things are ambiguous we need to continue until the priority then decreases (i.e. if we try one with priority 10, we have to try the others with priority 10 before we know if it's ambiguous or not, and whether to emit a warning about that).

glue/core/data_factories/helpers.py
astrofrog: Ok, sounds good.

glue/core/data_factories/dendro_loader.py
astrofrog: No, this does not read any data from the disk, since ``fits.open`` defaults to using memory mapping, so it only reads data at the last possible minute.

glue/core/data_factories/helpers.py
astrofrog: This is now done in 56eae08cbeb01f6918b45f79027822a80caf1d05

glue/core/data_factories/helpers.py
astrofrog: Done in 56eae08cbeb01f6918b45f79027822a80caf1d05

astrofrog:
@ChrisBeaumont - I have implemented your suggestions. In a separate pull request, I can change all the internal data loaders to use the ``@data_factory`` decorator too (I'll do it separately because in some cases it requires defining the identifier before the data factory, but I don't want to clutter the diffs here).

Does this seem ok now?

glue/core/data_factories/dendro_loader.py
ChrisBeaumont: Cool -- I also checked to see what happens when memmap=False, but hdulist[label] still doesn't seem to read any data (until `.data` is accessed).

glue/core/data_factories/gridded.py
ChrisBeaumont: That sounds like a good idea in a followup PR

ChrisBeaumont:
+1

glue/core/data_factories/dendro_loader.py
astrofrog: Ah interesting, I always assumed that with memmap=False the whole file was read *immediately*

astrofrog:
@ChrisBeaumont - thanks for reviewing!

PR #720, opened 2015-08-16
Fix dev version generation in setup [rebased]

astrofrog:


astrofrog:
Thanks for the fix @saimn!

PR #611, opened 2015-08-20
Expose true categorical data in ``CategoricalComponent``

astrofrog:
For a categorical component, ``data['name']`` and ``component.data`` both return numerical values that are mapped to the components, but there is no easy (public) way to get the original data as a Numpy array as far as I can tell. One can either use the private attribute ``component._categorical_data`` but this is not ideal because not part of the public API, and there is also ``to_series`` which can be used to get a dataframe. Can we add a way to get the original data as a Numpy array? (either by making ``_categorical_data`` public or adding ``to_array``?

### DIFF
--- a/glue/clients/histogram_client.py
+++ b/glue/clients/histogram_client.py
@@ -229,10 +229,10 @@ class HistogramClient(Client):
         c = list(self._get_data_components('x'))
         if c:
             c = c[0]
-            if isinstance(c, CategoricalComponent):
-                val = min(c._categories.size, 100)
+            if c.categorical:
+                val = min(c.categories.size, 100)
                 if not calculate_only:
-                    self.xlimits = (-0.5, c._categories.size - 0.5)
+                    self.xlimits = (-0.5, c.categories.size - 0.5)
 
         if not calculate_only:
             self.nbins = val
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import operator
 import logging
+import warnings
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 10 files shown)


astrofrog:
I've started working on this. Next step is to deprecate ``data``, but I need to explore how feasible this is because there are quite a few places where we just use ``.data`` for all components.

astrofrog:
Turned out to be easier than expected :) I tested that ``data`` is no longer used for categorical components by making it raise an exception and running all tests. I've now removed the exceptions and added a warning instead.

astrofrog:
@ChrisBeaumont - could you review this?

Note, there is one backward-incompatible change: ``CategoricalComponent.numeric`` now returns ``False``. I could change it ``True``, but I felt that it would make more sense to return ``False`` so that clients can deal correctly with categorical data. Is that ok?

glue/clients/histogram_client.py
```diff
@@ -229,10 +229,10 @@ def _auto_nbin(self, calculate_only=False):
         c = list(self._get_data_components('x'))
         if c:
             c = c[0]
-            if isinstance(c, CategoricalComponent):
-                val = min(c._categories.size, 100)
+            if c.categorical:
```
ChrisBeaumont: This is fine for now. As discussed in other threads, this kind of logic branch on component type information should probably instead be implemented as category methods, so that histogram clients don't need to know as much about how components behave. But I think @JudoWill's PR will address that

glue/core/data.py
```diff
@@ -405,6 +406,38 @@ def __init__(self, categorical_data, categories=None, jitter=None, units=None):
...
+    @property
+    def categories(self):
+        """
+        The categories.
+        """
+        return self._categories
+
+    @property
+    def data(self):
+        warnings.warn("The 'data' attribute is deprecated. Use 'codes' "
```
ChrisBeaumont: Does glue code currently trigger this warnign?

ChrisBeaumont: Ah just saw your note that it does not (at least in the tests)

ChrisBeaumont:
looks good to me

glue/clients/histogram_client.py
astrofrog: Ok, sounds good

PR #729, opened 2015-08-21
Add automatic file name extension when saving sessions

astrofrog:
When I save a session file, it saves it exactly with the name I enter in the dialogue (e.g. "mysession"), when when I try to load it again, that session file is not shown in the "Open" dialogue, because it does not have the correct file name extension. Either I should be able to select all files (even those with extensions that are not "glu") or glu should add the extension automatically when saving.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@ Full changelog
 v0.6 (unreleased)
 -----------------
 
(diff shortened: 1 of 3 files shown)


astrofrog:
@ChrisBeaumont - any objections?

ChrisBeaumont:
*1

PR #690, opened 2015-08-21
Improve cube slider

astrofrog:
At the moment, for a cube with 2000 slices, it's impossible to view some of the slices because the slider can't be moved to a high enough resolution. We could add step forward/back buttons on the side of the slider for more fine-tuned control

### DIFF
--- a/glue/qt/widgets/data_slice_widget.py
+++ b/glue/qt/widgets/data_slice_widget.py
@@ -2,12 +2,13 @@ from functools import partial
 from ...compat.collections import Counter
 
 from ...external.qt.QtGui import (QWidget, QSlider, QLabel, QComboBox,
-                                  QHBoxLayout, QVBoxLayout)
+                                  QHBoxLayout, QVBoxLayout, QPushButton, QLineEdit)
 from ...external.qt.QtCore import Qt, Signal
 
 from ..widget_properties import (TextProperty,
                                  ValueProperty,
                                  CurrentComboProperty)
+from ..qtutil import nonpartial
 
 
 class SliceWidget(QWidget):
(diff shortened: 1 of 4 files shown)


astrofrog:
I've implemented an initial version, which looks like this:

screen shot 2015-08-21 at 10 25 52 pm

@stscieisenhamer - does this seem ok?

This is a bit ugly at the moment code-wise, in particular for the hiding/showing. I think that I should make the whole slider + buttons thing a widget that we can then hide and show in a single line.

Also, needs tests, so not ready for merging!

astrofrog:
I updated this a little:

screen shot 2015-08-21 at 10 46 48 pm

and now the whole slider + buttons lives in a single widget in a ui file so it's easier to hide/show.

I also added some tests, so this is ready for final review.

stscieisenhamer:
:+1:

ChrisBeaumont:
What do you think about tightening up the spacing between these buttons?

glue/qt/widgets/data_slice_widget.py
```diff
@@ -45,33 +46,54 @@ def __init__(self, label='', pix2world=None, lo=0, hi=10,
...
-                                    self.slice_changed.emit(self.mode))
-        slider.valueChanged.connect(lambda x: slider_lbl.setText(str(x)))
-        layout.addWidget(slider_lbl)
+        slider = load_ui('cube_slider')
+        slider.slider
+
+        slider.slider.setMinimum(lo)
+        slider.slider.setMaximum(hi)
+        slider.slider.setValue((lo + hi) / 2)
+        slider.slider.valueChanged.connect(lambda x:
```
ChrisBeaumont: nit: use nonpartial instead of lambda

glue/qt/widgets/data_slice_widget.py
```diff
@@ -45,33 +46,54 @@ def __init__(self, label='', pix2world=None, lo=0, hi=10,
...
+        imax = self._ui_slider.slider.maximum()
+        value = self._ui_slider.slider.value()
+        if action == 'first':
+            value = imin
+        elif action == 'last':
+            value = imax
+        elif action == 'prev':
+            value = max(value - 1, imin)
+        elif action == 'next':
+            value = min(value + 1, imax)
```
ChrisBeaumont: nit: do you want to raise an error for other values of 'action'?

astrofrog:
@ChrisBeaumont - thanks for the review! I've now reduced the spacing:

screen shot 2015-08-24 at 10 10 33 pm

glue/qt/widgets/data_slice_widget.py
astrofrog: This was what you had originally, and it turns out that it was needed because ``self.mode`` isn't actually defined until ``__init__`` has run. Does this make sense?

glue/qt/widgets/data_slice_widget.py
astrofrog: Done

ChrisBeaumont:
tiiiiiiight

astrofrog:
:laughing:

PR #732, opened 2015-08-25
FITS containers [rebased]

astrofrog:
Fixes #704 

@stscieisenhamer - as you can see, this preserves your authorship (though I had to add a commit as part of the rebase) :)


stscieisenhamer:
@astrofrog Instead of me even attempting anything other than commenting, to fix the final test, the two str.format lines (lines 51 and 81) in containers.py need to read:
`
'{0}[{1}]'.format...
`
Then we're good. Note: this was fixed already, but completely lost, probably about the same time my repo went bad.

astrofrog:
@stscieisenhamer - fixed, thanks!

astrofrog:
@stscieisenhamer - I added some more tests and fixed a few corner cases. The only main change is that I added the identifier argument to the decorator so that this is used for all FITS files/objects.

astrofrog:
@ChrisBeaumont - is there any reason not to switch to using this as the default for all FITS files? (I made a modification to make it the default, but just want to check if you have any objections?)

stscieisenhamer:
:+1:

astrofrog:
@ChrisBeaumont - do you have any thoughts on the above?

astrofrog:
I plan to merge this later today.

stscieisenhamer:
Sorry, I had sent this earlier, but a local network disruption must have eaten it...

:+1:  to, in general, the session should record what actually happened, not what was intended to happen. Another scenario that this would help with is the situation where someone has a local reader but has made it the default for convenience sake. Having the session store what factory was actually used would prevent the confusing situation where User1 and User2 get drastically different results with no immediate indication as to why.

ChrisBeaumont:
Sorry for missing this thread . Setting this as the default fits loader sounds good to me. It'd be nice to have the feature to interactively determine how extensions are parsed into datasets and components, but this is nevertheless a step in the right direction (fewer failed loads)

astrofrog:
btw I decided to wait until I get #736 to work

stscieisenhamer:
A thought on the UI/data selection/identification issue: Considering the following points:
* FITS as general containers have Tables, which, in and of themselves, can be considered containers and be selected on.
* Other containers exists, such as HDF5, future ASDF, etc. which all have the same issues

and, the kicker:
* Another seemingly-unrelated-but-may-actually-be enhancement: redefinition of data already read in.

This last item has come up in casual conversation. The basic situation is that you have already read in data, but you want to change how that data is dealt with. Use cases would involve changing/defining WCS's, change units, change component attributes, merging/unmerging components/data sets, etc.
If one takes the philosophy of making getting data into glue really easy, and take care in well-implementing the last bullet, there would be no need to try to come up with a, possibly complicated and specific-to-format, data reader UI because any data manipulation is handled in-context within the app.
Just some rambling thought.

astrofrog:
@stscieisenhamer - yes, these are good points, though I can foresee situations where one still only just wants to read e.g. one HDU. However, I agree that some of the more complex operations (e.g. merge dataset X with dataset U and J but not K and Y) could be done in-app.

PR #741, opened 2015-09-07
WIP: Make subset and default data colors customizable (and re-arrange config)

astrofrog:
This also introduces a ``glue.core.config`` file. One of the things that's been bugging me for a while is the fact that core in ``glue.core`` needs to access code outside ``glue.core``, specifically in ``glue.config``.

What I'd like to do is move all the config related to ``glue.core`` in ``glue.core.config`` and to then import all that into ``glue.config``, so that this presents no changes to the user, but all the config related to ``glue.core`` can be accessed without going to the top-level.

More generally, I think that various sub-packages should have their own ``config.py`` file that then gets imported into ``glue.config`` - this helps maintain isolation of components. I'm going to try implementing that as part of this PR.


astrofrog:
Hmm, I just saw there is already a ``settings`` registry, so it probably makes sense to use that instead. However, this is not really used much at the moment (except for plotly). I'll have a think about whether it makes more sense to go with the registry approach or instead define parameters as constants in subpackage-level ``config.py`` files.

astrofrog:
This is more meant as a proof of concept, and we don't have to merge it, but basically all the registries needed by the core now live in ``glue.core.config``. The benefits aren't really clear though - what I was trying to do was to allow settings to live closer to the code they apply to. This all started because I thought that ``SUBSET_COLORS`` should live somewhere in ``glue.core`` not ``glue.config``. However, I could also simply add

```
from ..config import settings
settings.add('SUBSET_COLORS', [...])
```

However, I do think that e.g. ``fit_plugin`` should live in the spectral_tool plugin for example, so something similar to what is done here could work, but basically with the core config items in ``glue.config`` instead of ``glue.core.config`` (but keep the qt ones in ``glue.qt.config`` for instance).

astrofrog:
I'm going to close this since I don't think it really makes sense, and #742 is a better solution for now

PR #742, opened 2015-09-07
Make color cycle and default colors customizable

astrofrog:
This is an alternative to #741 for now, and simply uses the existing ``settings`` config registry to store the settings.

### DIFF
--- a/glue/config.py
+++ b/glue/config.py
@@ -557,3 +557,24 @@ def _default_search_order():
         search_order.append(os.environ['GLUERC'])
     search_order.append(os.path.join(config.CFG_DIR, 'config.py'))
     return search_order[::-1]
+
+
+###### Now define global settings ######
+
+GRAY = '#373737'
+BLUE = "#1F78B4"
+GREEN = "#33A02C"
+RED = "#E31A1C"
+ORANGE = "#FF7F00"
+PURPLE = "#6A3D9A"
+YELLOW = "#FFFF99"
+BROWN = "#8C510A"
+PINK = "#FB9A99"
+LIGHT_BLUE = "#A6CEE3"
+LIGHT_GREEN = "#B2DF8A"
+LIGHT_RED = "#FB9A99"
+LIGHT_ORANGE = "#FDBF6F"
+LIGHT_PURPLE = "#CAB2D6"
+
+settings.add('SUBSET_COLORS', [RED, GREEN, BLUE, BROWN, ORANGE, PURPLE, PINK], validator=list)
+settings.add('DEFAULT_VISUAL_COLOR', GRAY)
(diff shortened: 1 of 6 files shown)


astrofrog:
Ok, so at the moment, one can (with this PR) set the color cycle with e.g.

```
from glue.core import visual
from glue.config import settings

settings.SUBSET_COLORS = ['#a6cee3', '#1f78b4', '#b2df8a', '#33a02c']
```

Note that ``visual`` needs to be imported first because the settings are added in there, which is not good.

One option would be to simply define these in ``glue/config.py``, which is what I wanted to avoid ideally. Another option is to put it in ``glue/core/__init__.py`` and make sure ``glue.core`` gets imported before the user config is loaded.

astrofrog:
The example above is meant to be a colorblind-friendly map, which gives something like this:

screen shot 2015-09-07 at 10 47 00 pm

Unfortunately, this is the colorbrewer map with the most colors that is still colorblind-friendly (4 colors), so not sure what to do if the user wants e.g. 5.

astrofrog:
@ChrisBeaumont - at the moment, once a subset color is picked, it is fixed, but it might be desirable in some cases to read in a session file and select a different color cycle - so maybe it would make sense to actually just store colors as integers that are indices in the current color cycle. What do you think?

astrofrog:
Hmm, I think maybe the settings should live at the root of the package after all, either in ``config.py`` or in a new ``settings.py`` file. @ChrisBeaumont - any thoughts on this?

ChrisBeaumont:
Can you explains your hesitation about config.py? I've often wondered about
what should go in there and what should go in a more declarative file (eg
yaml)

I'm +1 on the general idea of moving ~all defaults to some kind of config
system.

ChrisBeaumont:
Maybe we just need to be better about not copying data out of settings or
config files, so that we don't have to worry as much about import order

astrofrog:
@ChrisBeaumont the hesitation about config.py is just that in general I think it would be clearer if code in ``core`` didn't have to import from anywhere outside ``core``, but don't think that's going to happen easily. Ideally the config items (e.g. default color) should live close to where they are used. Astropy does that, and it then basically exports those settings to a cfg file that the user can edit. I'd rather not use that system because it's actually reasonably complex though.

For now, maybe I'll just move the color setting to ``glue/config.py`` and then we can plan to discuss this in more detail in future - maybe it would make sense to have a chat on hangouts at some point?

astrofrog:
Ok, this should be ready to merge once tests pass.

ChrisBeaumont:
Sounds good. Yeah, let's have a more general discussion about user config
at some point

Issue #730, opened 2015-08-22
Resizing of axes causes 'bouncing' of axes


astrofrog:
This is a bit hard to describe in words but when resizing a window, sometimes it looks as though the axes itself are resized faster than the actual Qt window:

![resize_opt](https://cloud.githubusercontent.com/assets/314716/9423852/857b28b2-48d6-11e5-9327-2168d197e2ef.gif)

This looks a bit weird when using interactively. Now I originally thought this was related to how fast axes are drawn but actually this can't be the case since the axes are actually following the cursor *better* than the Qt window. Maybe there is a Qt setting related to making things smooth?

@ChrisBeaumont - do you know off the top of your head what could be causing this? If not, I can dig deeper into it.

astrofrog:
Note, the axes **are** a bit slow to update, but it's more that the axes and window aren't in sync

astrofrog:
This was fixed in #745

PR #745, opened 2015-09-09
Don’t use tight layout as it causes bouncing around of axes

astrofrog:
I looked into the issue described in https://github.com/glue-viz/glue/issues/730, namely that the axes appear to bounce around when being resized. This is due to the tight layout option, which causes the axes to be re-drawn after every draw with a different bounding box.

This PR disables this, and tries to instead optimize space used by setting the rect for the axes created. Of course, this isn't perfect either because the axes size is relative to the window so for larger axes there is a lot of wasted space. I am going to look into whether there is a better way to do that, but in the mean time, this does produce smoother axes resizing without any bouncing.

### DIFF
--- a/glue/utils/matplotlib.py
+++ b/glue/utils/matplotlib.py
@@ -176,3 +176,49 @@ def point_contour(x, y, data):
         return None
     xy = xy[0]
     return xy
+
+
+def FixedMarginAxes(ax_class, margins=[1, 1, 1, 1]):
+    """
+    Class factory to make an axes class that will preserve fixed margins when
+    figure is resized.
+
+    Parameters
+    ----------
+    ax_class : matplotlib.axes.Axes
+        The axes class to wrap
+    margins : iterable
+        The margins, in inches. The order of the margins is
+        ``[left, right, bottom, top]``
+    """
+
+    # Note that margins gets used directly in get_fixed_margin_rect and we
+    # don't pass it through the axes class.
+
+    def get_fixed_margin_rect(fig):
+
+        fig_width = fig.get_figwidth()
+        fig_height = fig.get_figheight()
+
+        x0 = margins[0] / fig_width
+        x1 = 1 - margins[1] / fig_width
+        y0 = margins[2] / fig_height
+        y1 = 1 - margins[3] / fig_height
+
+        dx = max(0.01, x1 - x0)
+        dy = max(0.01, y1 - y0)
+
+        return [x0, y0, dx, dy]
+
+    class ax_subclass(ax_class):
+
+        def __init__(self, fig, **kwargs):
+            rect = get_fixed_margin_rect(fig)
+            super(ax_subclass, self).__init__(fig, rect, **kwargs)
+
+        def draw(self, *args, **kwargs):
+            rect = get_fixed_margin_rect(self.figure)
+            self.set_position(rect)
+            super(ax_subclass, self).draw(*args, **kwargs)
+
+    return ax_subclass
--- a/glue/utils/matplotlib.py
+++ b/glue/utils/matplotlib.py
@@ -178,7 +178,7 @@ def point_contour(x, y, data):
     return xy
 
 
-def FixedMarginAxes(ax_class, margins=[1, 1, 1, 1]):
+def fixed_margin_axes(ax_class, margins=[1, 1, 1, 1]):
     """
     Class factory to make an axes class that will preserve fixed margins when
     figure is resized.
--- a/glue/utils/matplotlib.py
+++ b/glue/utils/matplotlib.py
@@ -178,47 +178,42 @@ def point_contour(x, y, data):
     return xy
 
 
-def fixed_margin_axes(ax_class, margins=[1, 1, 1, 1]):
+class AxesResizer(object):
+
+    def __init__(self, ax, margins):
+
+        self.ax = ax
+        self.margins = margins
+
+    def on_resize(self, event):
+
+        fig_width = self.ax.figure.get_figwidth()
+        fig_height = self.ax.figure.get_figheight()
+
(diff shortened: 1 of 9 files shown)


astrofrog:
@ChrisBeaumont - it would be great if you could try this out when you get a chance. I've now added a class factory that ensures that the margin between axes and the edge of the figure is constant in absolute terms. This allows us to minimize whitespace while avoiding the issues due to the tight layout option.

At the moment, this isn't perfect for the ``AxesCache`` but we can probably get it to work.

astrofrog:
(I will fix the test failures if we go with this approach)

astrofrog:
@ChrisBeaumont - I'm pretty happy with this approach now which works even for the ``AxesCache``. Could you review this and let me know what you think?

astrofrog:
Fixes #730

ChrisBeaumont:
Will take a look tmrw morning

ChrisBeaumont:
Cool, this seems to work well for me. Do you know how this works if a plot has multiple axes on it? I'm trying to think and I don't think there are any standard viewers that do so, so It's probably not a blocker either way.

astrofrog:
@ChrisBeaumont - thanks for reviewing! None of the viewers I am using this for have multiple axes. If a user defined axes in a custom viewer I am not applying this method to those to be safe, but I could document how to use the AxesResizer since it's pretty easy (alternatively, if the custom viewer can be used only with single axes, I can attach the AxesResizer to those too.

ChrisBeaumont:
IIRC it's not easy to build a custom viewer with multiple axes (the axes are auto-built and handed off to the user). It might be nice to:

 - Attach AxesResizers to the axes in custom viewers
 - Document how to modify the parameters of those resizers, in case users want to modify the margins of their custom plots

astrofrog:
@ChrisBeaumont - actually at the moment ``custom_viewer.create_axes`` is not a user-definable function unless their viewer inherits from ``custom_viewer`` and overrides it. We don't mention this route in the docs - so do we want to make ``create_axes`` a user-definable function?

If so, then we can also make it that to set the axes margins, the user/developer can specify ``custom_viewer.axes_margins`` to be a list of four values, which will then get used for the margins. What do you think?

astrofrog:
Sorry, just realized you said the same thing about the fact the user doesn't create the axes. I will implement the axes_margins parameter so the user can set them.

glue/utils/matplotlib.py
```diff
@@ -176,3 +176,44 @@ def point_contour(x, y, data):
...
+    Parameters
+    ----------
+    ax_class : matplotlib.axes.Axes
+        The axes class for which to fix the margins
+    margins : iterable
+        The margins, in inches. The order of the margins is
+        ``[left, right, bottom, top]``
+    """
+
+    resizer = AxesResizer(axes, margins)
```
ChrisBeaumont: it might be nice to store this resizer as an attribute on the axes somewhere, in case downstream code wishes to modify it.

glue/utils/matplotlib.py
```diff
@@ -176,3 +176,44 @@ def point_contour(x, y, data):
...
+        y1 = 1 - self.margins[3] / fig_height
+
+        dx = max(0.01, x1 - x0)
+        dy = max(0.01, y1 - y0)
+
+        self.ax.set_position([x0, y0, dx, dy]) 
+        self.ax.figure.canvas.draw()
+
+
+def fix_margins(axes, margins=[1, 1, 1, 1]):
```
ChrisBeaumont: what do you think about the name `freeze_margins`? `fix` could be interpreted as implying that the default is `broken`.

glue/utils/matplotlib.py
astrofrog: Hmm, that might be a way to customize the margins then, if it's accessible as ``axes.resizer.margins``. I worry a bit about adding ``axes.resizer`` though in case it conflicts with Matplotlib in future for any reason. Do you think we should just do it anyway and cross that bridge if it happens?

glue/utils/matplotlib.py
astrofrog: Yes, sounds good

glue/utils/matplotlib.py
ChrisBeaumont: I don't think we should worry about future name clashes with resizer

glue/utils/matplotlib.py
astrofrog: Fixed

glue/utils/matplotlib.py
astrofrog: Done :)

astrofrog:
@ChrisBeaumont - I implemented your suggestions, thanks! I'll keep this open for another day in case you want to review again.

ChrisBeaumont:
I was thinking about this a little more. I guess the main advantage of tight_layout is that it auto-chooses sensible initial margins for an arbitrary plot (and potentially complicated ones with multiple axes, weird axis labels, etc). It's drawback (fixed in this PR) is that it auto-updates these margins, leading to jumpy panning/zooming. So maybe the ideal solution is one that uses tight_layout's internal mechanism to determine good initial margins, and then passes *these* values to freeze_margins. That's not a high priority, but something we could consider if we start to notice that the current hard-coded frozen margins don't look good in certain situations.

astrofrog:
@ChrisBeaumont - yeah, I was thinking of that too. That might be possible *if* we can avoid an initial draw before tight_layout figures out the best margins (otherwise we'll get jumpy behavior upon initialization)

ChrisBeaumont:
Do you actually notice jumpiness at init time? I mostly see this during pan or zoom, when a new tick label is created which changes the margins. I haven't noticed any jumpiness when you first open a window

astrofrog:
Ah, I will double check

PR #612, opened 2015-09-23
Table Viewer widget?

astrofrog:
It would be great to be able to explore the tabular data for a full dataset or subset.

@ChrisBeaumont - is what is in ``table_widget.py`` a first stab at working on this? Did you explore this in more detail and did you run into any issues? I need this so could look into it too.

### DIFF
--- a/glue/external/qt.py
+++ b/glue/external/qt.py
@@ -176,6 +176,7 @@ def _load_pyqt5():
             setattr(QtGui, widget, getattr(QtWidgets, widget))
     QtGui.QItemSelection = QtCore.QItemSelection
     QtGui.QItemSelectionModel = QtCore.QItemSelectionModel
+    QtGui.QSortFilterProxyModel = QtCore.QSortFilterProxyModel
 
     register_module(QtCore, 'QtCore')
     register_module(QtGui, 'QtGui')
--- a/glue/external/qt.py
+++ b/glue/external/qt.py
@@ -174,6 +174,7 @@ def _load_pyqt5():
     for widget in dir(QtWidgets):
         if widget.startswith('Q'):
             setattr(QtGui, widget, getattr(QtWidgets, widget))
+    QtGui.QItemSelection = QtCore.QItemSelection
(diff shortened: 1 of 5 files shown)


astrofrog:
@nmearl - just for info, this is the work in progress on the table viewer. It's still very alpha quality, but basically should allow selections to/from other widgets (it took a while to iron out a lot of kinks with performance and selection modes). Maybe I should copy this to the sprint repository for now though (note that since it looks like there is a basic table widget there currently that one might take precedence)

nmearl:
This looks great. I see that you're wrapping the rows in a subset state, and then just broadcasting the selection via the mode update?. My experience with subsets is pretty limited to array data and masking -- what would the selection broadcast data look like in the message received by other widgets? I'm imaging dealing with `subset = message.sender` in one of the overridden methods.

I guess a more general question would be: what does tabular look like in Glue's internals?

astrofrog:
@nmearl - internally, tabular and gridded data are actually the same - a table is just a 1D data object with as many components as columns. When you receive the subset it should be the same as receiving a subset from say a scatter plot. Basically what works with the scatter plot should work with this.

astrofrog:
@nmearl @ChrisBeaumont - I've now removed this again from the list of default widgets, and I think it would be worth merging this so that it can already be used (despite being experimental) as opt-in.

ChrisBeaumont:
I'm fine with merging this as an opt-in viewer. I think I pushed on this a bit further in the table branch, and you might find some worthwhile ideas there that you could pull in (sorry for not pointing that out earlier).

astrofrog:
@ChrisBeaumont - just took a look and it seems like we were implementing mostly different additional things, so it should be possible to reconcile the two. I'll merge this first and have opened an issue to remind us about the additional changes (https://github.com/glue-viz/glue/issues/795)

astrofrog:
I've copied over the fixes to qt_helpers here: https://github.com/glue-viz/qt-helpers/pull/6

PR #752, opened 2015-10-03
Improvements to widget properties

astrofrog:
Several changes in this pull request, following work on the 3-d viewer:

* ``CurrentComboProperty`` has been renamed to ``CurrentComboDataProperty`` and ``CurrentComboTextProperty`` has been added, to differentiate between cases where the item should be selected based on the text vs data in the combo box (note that for backward-compatibility, ``CurrentComboProperty`` is set to ``CurrentComboDataProperty``)
* Adds the ability to define a mapping function for value-based properties, which can be useful for e.g. log-space sliders
* Adds ``CurrentTabProperty``. The tests have been expanded to cover all widget properties.
* Significantly expands tests in ``test_widget_properties.py`` to test all functions/classes

@ChrisBeaumont - am I right in thinking that when one doesn't need an actual callback, but instead just needs the property to be set when called, then the widget properties should be used, whereas if an immediate callback is needed, the callback properties and connect functions should be used?

### DIFF
--- a/glue/qt/widget_properties.py
+++ b/glue/qt/widget_properties.py
@@ -83,8 +83,11 @@ class CurrentComboDataProperty(WidgetProperty):
         Update the currently selected item to the one which stores value in
         its itemData
         """
-        idx = widget.findData(value)
-        if idx == -1:
+        # Note, we don't use findData here because it doesn't work
+        # well with non-str data
+        try:
+            idx = _find_combo_data(widget, value)
+        except ValueError:
             raise ValueError("Cannot find data '{0}' in combo box".format(value))
         widget.setCurrentIndex(idx)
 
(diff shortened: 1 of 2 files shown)


glue/qt/widget_properties.py
```diff
@@ -60,32 +60,88 @@ def __set__(self, instance, value):
...
+        except ValueError:
+            raise ValueError("Cannot find data '{0}' in combo box".format(value))
         widget.setCurrentIndex(idx)
 
+CurrentComboProperty = CurrentComboDataProperty
+
+
+class CurrentComboTextProperty(WidgetProperty):
+    """
+    Wrapper around ComboBoxes
```
ChrisBeaumont: differentiate docstrings here?

ChrisBeaumont:
This looks good

> am I right in thinking that when one doesn't need an actual callback, but instead just needs the property to be set when called, then the widget properties should be used, whereas if an immediate callback is needed, the callback properties and connect functions should be used?

That sounds right. I guess we could make it so that callbacks could more easily be attached to WidgetProperties directly (ie maybe WidgetProperty should inherit from CallbackProperty)

glue/qt/widget_properties.py
astrofrog: Done!

PR #756, opened 2015-10-06
Save loaded plugins

astrofrog:
Also moves some of the generic restore_session code to Application.

I had another commit which raised a warning if plugins were missing on loading, but it will generate unnecessary noise in most cases. What we need to know really is whether the plugins are *really* needed to load the session (someone could have had an irrelevant plugin loaded). However, doing this is going to be a fair bit more work.

For now, I'd like to merge this, just for information purposes, so that if we need to diagnose issues with users having issues loading files, we can check manually which plugins were loaded.

@ChrisBeaumont - does this sound reasonable? The plan would be to then figure out (after 0.6) how to determine exactly which plugins are needed when loading the session.

### DIFF
--- a/glue/core/application_base.py
+++ b/glue/core/application_base.py
@@ -95,6 +95,28 @@ class Application(HubListener):
         with open(path, 'w') as out:
             out.write(state)
 
+    @staticmethod
+    def restore_session(path):
+        """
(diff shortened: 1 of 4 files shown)


ChrisBeaumont:
LGTM

PR #755, opened 2015-10-16
Bug when selecting attribute in RGB image

astrofrog:
To reproduce:

* Load RGB image (e.g. PNG)
* Open image viewer
* Change attribute, and nothing changes

### DIFF
--- a/glue/qt/widget_properties.py
+++ b/glue/qt/widget_properties.py
@@ -76,7 +76,10 @@ class CurrentComboDataProperty(WidgetProperty):
         """
         Return the itemData stored in the currently-selected item
         """
-        return widget.itemData(widget.currentIndex())
+        if widget.currentIndex() == -1:
+            return None
+        else:
+            return widget.itemData(widget.currentIndex())
 
     def setter(self, widget, value):
         """
@@ -88,7 +91,10 @@ class CurrentComboDataProperty(WidgetProperty):
         try:
             idx = _find_combo_data(widget, value)
         except ValueError:
-            raise ValueError("Cannot find data '{0}' in combo box".format(value))
+            if value is None:
+                idx = -1
+            else:            
+                raise ValueError("Cannot find data '{0}' in combo box".format(value))
         widget.setCurrentIndex(idx)
 
 CurrentComboProperty = CurrentComboDataProperty
@@ -103,7 +109,10 @@ class CurrentComboTextProperty(WidgetProperty):
         """
         Return the itemData stored in the currently-selected item
         """
-        return widget.itemText(widget.currentIndex())
+        if widget.currentIndex() == -1:
+            return None
+        else:
+            return widget.itemText(widget.currentIndex())
 
     def setter(self, widget, value):
         """
@@ -112,7 +121,8 @@ class CurrentComboTextProperty(WidgetProperty):
         """
         idx = widget.findText(value)
         if idx == -1:
-            raise ValueError("Cannot find text '{0}' in combo box".format(value))
+            if value is not None:
+                raise ValueError("Cannot find text '{0}' in combo box".format(value))
         widget.setCurrentIndex(idx)
 
 
--- a/glue/qt/widget_properties.py
+++ b/glue/qt/widget_properties.py
@@ -237,6 +237,8 @@ def connect_current_combo(client, prop, widget):
     """
 
     def _push_combo(value):
+        # NOTE: we can't use findData here because if the data is not a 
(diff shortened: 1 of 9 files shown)


astrofrog:
@ChrisBeaumont - I'm a bit confused about this bug because I'm pretty sure it would have worked in the past, but this should fix it.

astrofrog:
Weirdly, this does not work if I now load two images and then flick between the datasets in the same image viewer, in that ``_update_attribute`` raises an exception. I need to look into it.

astrofrog:
@ChrisBeaumont - could you review this? Basically this fixes the scenario where you load two different datasets (in my case an RGB image and a FITS file), then drag dataset 1 onto the the main area and make an image viewer, then drag dataset 2 onto the main area (this should cause the data in the combo box to change, which it didn't before), then select dataset 1 from the data combo box again (which caused various errors).

Is this now the intended behavior, or should one not be able to add multiple datasets to an image viewer?

I'll add a regression test if this looks good to you otherwise.

glue/qt/widget_properties.py
```diff
@@ -237,7 +237,7 @@ def connect_current_combo(client, prop, widget):
     """
 
     def _push_combo(value):
-        idx = widget.findData(value)
+        idx = _find_combo_data(widget, value)
```
astrofrog: This will go away once #771 is merged and this is rebased.

astrofrog: As indicated in #755, I think we actually need to merge this first, so I'm actually just going to merge in the changes from 771 here to make things easier.

astrofrog:
Ignore for now - there are genuine test failures.

astrofrog:
@ChrisBeaumont - this is now ready for review. It grew beyond the original bug and fixes:

* The fact that the image did not update if the RGB component was changed
* The fact that the data and combo boxes were not updated correctly when showing multiple datasets in the same image viewer

I think that everything *should* pass now. Let me know what you think!

glue/qt/widget_properties.py
```diff
@@ -237,9 +247,15 @@ def connect_current_combo(client, prop, widget):
     """
 
     def _push_combo(value):
-        idx = widget.findData(value)
-        if idx == -1:
-            return
+        # NOTE: we can't use findData here because if the data is not a 
+        # string, PySide will crash
```
ChrisBeaumont: I've been bitten by `findData` like this too. We could envision fixing this method in our qt import/wrapper utility, so that findData behaves as we expect.

ChrisBeaumont: (well maybe, I'm not sure how much metaclass magic happens when you instantiate a QWidget, and whether that makes overriding methods tricky)

glue/qt/widgets/image_widget.py
```diff
@@ -218,15 +220,35 @@ def _connect(self):
...
+        add_callback(self.client, 'display_data', nonpartial(self._display_data_changed))
+        add_callback(self.client, 'display_attribute', nonpartial(self._display_attribute_changed))
+        add_callback(self.client, 'display_aspect', nonpartial(self.client._update_aspect))
 
         # sync data/attribute combos with client properties
         connect_current_combo(self.client, 'display_data', self.ui.displayDataCombo)
         connect_current_combo(self.client, 'display_attribute', self.ui.attributeComboBox)
         connect_current_combo(self.client, 'display_aspect', self.ui.aspectCombo)
 
+    def _display_data_changed(self, *args):
```
ChrisBeaumont: We can get rid of `*args` here because of the `nonpartial`, right?

glue/qt/widgets/image_widget.py
```diff
@@ -218,15 +220,35 @@ def _connect(self):
...
+            self.ui.attributeComboBox.clear()
+            return
+
+        with self.client.artists.ignore_empty():
+            self.set_attribute_combo(self.client.display_data)
+        self.client.add_layer(self.client.display_data)
+        self.client._update_and_redraw()
+        self.update_window_title()
+
+    def _display_attribute_changed(self, *args):
```
ChrisBeaumont: ditto

glue/qt/widget_properties.py
astrofrog: I've made an issue to remind me to do this:

https://github.com/glue-viz/glue/issues/774

It will require some more in-depth testing hence why I'm postponing it for now.

ChrisBeaumont:
LGTM

ChrisBeaumont:
LGTM

PR #758, opened 2015-10-18
Default to linear stretch and min/max limits

astrofrog:
I think it makes sense to default to min/max and a linear stretch. The current defaults often over-clip:

**Current**:

screen shot 2015-10-18 at 2 02 58 pm

**This PR**

screen shot 2015-10-18 at 2 01 35 pm

I was also wondering whether we should move the selection of the percentile, colormap, and scaling function to the options widget? We could then additionally add two fields to manually set the data limits.

@ChrisBeaumont - what do you think?

### DIFF
--- a/glue/clients/ds9norm.py
+++ b/glue/clients/ds9norm.py
@@ -123,8 +123,8 @@ class DS9Normalize(Normalize, object):
         self.stretch = 'linear'
         self.bias = 0.5
         self.contrast = 1.0
-        self.clip_lo = 5.
-        self.clip_hi = 95.
+        self.clip_lo = 0.
+        self.clip_hi = 100.
 
     @property
     def stretch(self):
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -55,6 +55,8 @@ v0.6 (unreleased)
 * Added two new settings ``settings.SUBSET_COLORS`` and ``settings.DATA_COLOR``
   that can be used to customize the default subset and data colors. [#742]
 
+* Improve defaults for image data limits.
+
 v0.5.3 (unreleased)
 -------------------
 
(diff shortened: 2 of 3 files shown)


glue/clients/ds9norm.py
```diff
@@ -123,8 +123,8 @@ def __init__(self):
         self.stretch = 'linear'
         self.bias = 0.5
         self.contrast = 1.0
-        self.clip_lo = 5.
-        self.clip_hi = 95.
+        self.clip_lo = 0.
+        self.clip_hi = 100.
```
ChrisBeaumont: Unfortunately IIRC these limits don't look nice on the W5 image we use in the getting started tutorial (or I would think many IR / optical images containing both stars and ISM). I don't know if there's a smarter way to choose the defaults based on the image contents?

astrofrog: It looks the same as what one would get in e.g. ds9 though, and I think it's going to be really hard to find a default that is esthetically pleasing for all datasets. I think that if we make the data limits and options to change the scale more obvious by putting them in the options widget, we could then include a step in the tutorial to show users how to change the stretch - what do you think?

astrofrog: Just to clarify, the contrast and bias would still get set by the mouse mode, but I'm suggesting to move the selection of the function and percentile to the options widget.

ChrisBeaumont: Moving the stretch to the options widget seems like a good idea. I remember considering this a few times and I ran into some implementation issue. But conceptually I think it makes sense, and maybe it's easier than I remember

astrofrog:
This has now been done while refactoring the viewers

PR #759, opened 2015-10-18
Improve histogram limit setting

astrofrog:
This implements the following improvements:

* fixes the failures with Numpy 1.10 (in principle)
* changes the automated limit calculation for histograms to ignore values <= 0 when xlog is True
* remembers the separate min/max settings for xlog=True/False
* remembers which components should be shown in log
* disable xlog checkbox for categorical components

This needs unit tests to test what has been implemented above.

### DIFF
--- a/glue/clients/histogram_client.py
+++ b/glue/clients/histogram_client.py
@@ -231,9 +231,11 @@ class HistogramClient(Client):
                 continue
 
             if self.xlog:
-                if np.any(data > 0):
-                    lo = min(lo, np.nanmin(data[data > 0]))
-                    hi = max(hi, np.nanmax(data[data > 0]))
+                positive = data > 0
+                if np.any(positive):
+                    positive_data = data[positive]
+                    lo = min(lo, np.nanmin(positive_data))
+                    hi = max(hi, np.nanmax(positive_data))
                 else:
                     lo = 1
                     hi = 10
--- a/glue/clients/histogram_client.py
+++ b/glue/clients/histogram_client.py
@@ -64,8 +64,10 @@ class HistogramClient(Client):
         self._figure, self._axes = init_mpl(figure=figure, axes=None)
         self._component = None
         self._saved_nbins = None
-        self._xlim = {}
+        self._xlim_cache = {}
+        self._xlog_cache = {}
         self._sync_enabled = True
+        self._xlog_curr = False
 
     @property
     def bins(self):
@@ -229,8 +231,12 @@ class HistogramClient(Client):
                 continue
 
             if self.xlog:
-                lo = min(lo, np.nanmin(data[data > 0]))
-                hi = max(hi, np.nanmax(data[data > 0]))
+                if np.any(data > 0):
+                    lo = min(lo, np.nanmin(data[data > 0]))
+                    hi = max(hi, np.nanmax(data[data > 0]))
+                else:
+                    lo = 1
+                    hi = 10
             else:
                 lo = min(lo, np.nanmin(data))
                 hi = max(hi, np.nanmax(data))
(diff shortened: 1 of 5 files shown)


glue/qt/widgets/histogram_widget.py
```diff
@@ -193,6 +182,20 @@ def component(self, component):
 
     @defer_draw
     def _set_attribute_from_combo(self, *args):
+        for d in self._data:
+            try:
+                component = d.get_component(self.component)
+            except:
+                continue
+            else:
+                break
```
astrofrog: @ChrisBeaumont - the above is ugly - is there a preferred way to retrieve a component from a component ID if the dataset is not known? Should this be a method on ``DataCollection``? (or can one component ID in principle refer to components in separate dataset?)

glue/qt/widgets/histogram_widget.py
```diff
@@ -193,6 +183,21 @@ def component(self, component):
...
     @defer_draw
     def _set_attribute_from_combo(self, *args):
+        if self.component is not None:
+            for d in self._data:
+                try:
+                    component = d.get_component(self.component)
+                except:
+                    continue
+                else:
+                    break
```
astrofrog: @ChrisBeaumont - the above is ugly - is there a preferred way to retrieve a component from a component ID if the dataset is not known? Should this be a method on DataCollection? (or can one component ID in principle refer to components in separate dataset?)

glue/clients/histogram_client.py
```diff
@@ -234,9 +216,38 @@ def _auto_nbin(self, calculate_only=False):
...
+                data = a.layer[self.component]
+            except IncompatibleAttribute:
+                continue
+
+            if data.size == 0:
+                continue
+
+            if self.xlog:
+                if np.any(data > 0):
+                    lo = min(lo, np.nanmin(data[data > 0]))
```
ChrisBeaumont: nit: data[data>0] is run twice, and is potentially expensive.

glue/qt/widgets/histogram_widget.py
ChrisBeaumont: There's no directly link from `CompoentID` to `Component`, so some loop like this seems unavoidable to me. In general a ComponentID can belong to >1 data objects in a DataCollection. I could se the argument for making this a utility method of DataCollection.

glue/clients/histogram_client.py
astrofrog: Fixed

glue/qt/widgets/histogram_widget.py
astrofrog: Just to be clear, if a componentID belongs to multiple datasets, the actual component is the same though, right? So choosing any of them should be ok?

astrofrog: Hmm, I guess that one can end up with a component ID referring to different data:

```python
In [2]: d1 = Data(x=[1,2,3])

In [3]: d2 = Data(y=[4,5,6])

In [4]: c1 = d1.find_component_id('x')

In [5]: c2 = d2.find_component_id('y')

In [6]: 

In [6]: d1.update_id(c1, c2)

In [7]: 

In [7]: print(d1[c2])
[1 2 3]

In [8]: print(d2[c2])
[4 5 6]
```

astrofrog: ``HistogramClient`` seems to be making the same assumption I made here, namely that a componentID refers to a unique component (even if present in multiple datasets).

astrofrog:
Since this PR fixes an immediate bug with Numpy 1.10, and I think what I did with the componentID is consistent with what is done in HistogramClient, I'm going to go ahead and merge this, and then open another issue to continue the discussion about ComponentIDs.

glue/qt/widgets/histogram_widget.py
ChrisBeaumont: When ComponentIDs are present in 2 datasets, it's because those two components have been linked together and are conceptually equivalent. The intent of this codeblock is to check if a component is categorical. There's technically nothing stopping you from linking a categorical component with a non-categorical one, but I don't think this would ever legitimately come up in practice. So I think the behavior here (grabbing the first component you find), should be reasonably well defined.

PR #762, opened 2015-10-18
Astropy ASCII tables

astrofrog:
This moves the Astropy table readers to a dedicated file. We now try and read in any file that cannot otherwise be read (priority=0) as a last attempt, which allows us to read a number of Astropy-friendly formats that could not be read before.

cc @teuben

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,8 @@ v0.6 (unreleased)
 * When a box has been drawn to extract a spectrum from a cube, the box can
   then be moved by pressing the control key and dragging it. [#707]
 
+* Refactored ASCII I/O to include more Astropy table formats. [#762]
+
 * When saving a session, if no extension is specified, the .glu extension is
   added. [#729]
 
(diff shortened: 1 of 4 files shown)


astrofrog:
@ChrisBeaumont - this gets rid of the ascii.glue format. I found that by re-arranging how the ASCII readers are dealt with, we can read in files without any issues (basically we always try and read with Table.read(..., format='ascii') **first**, which avoids errors related to ambiguities. Do you see any reason (other than performance) why we shouldn't do what's done in this PR?

ChrisBeaumont:
The ascii.glue format was initially added to sidestep some of astropy's format guessing (eg https://github.com/astropy/astropy/pull/1935#issuecomment-31972638). For that specific comment at least the original issues were caught by Glue's test suite, so in principle change that satisfies our tests should be fine. 

It's hard to know when it's safe to change our format guessing strategy, since we don't have test cases for all the various ways Table.read has broken on us in the past (many of which have been segfaults). But these changes at least look reasonable, so maybe we should try them out (and make sure to add new test cases as issues come up).

astrofrog:
@ChrisBeaumont - ok, thanks! If this passes, I'll go ahead and merge, and will try and include tests for more formats later (and every time users report an issue)

PR #766, opened 2015-10-20
Update docs url on help menu

### LINKED ISSUE #763: the help menu in the glue script is too old
1) There is no help->about to say what version of glue this is.

and related:

2) help->online documentation goes to    http://www.glue-viz.org/en/latest/
but this seems to refer to some old 0.1 version???
Instead http://www.glueviz.org/en/latest/ seems to be more up to date,
and point to 0.4.0.  
Oddly, http://www.glueviz.org/en/stable/ also has this version. So that's odd. 
Should one of them not point to 0.5.3 ?

ChrisBeaumont:
Closes #763


astrofrog:
Thanks!

PR #767, opened 2015-10-20
Fixed a bug that caused compressed image HDUs to not be read correctly

astrofrog:
Thanks to @Cadair for finding this :)


astrofrog:
The AppVeyor failure is unrelated and is fixed by #759

PR #768, opened 2015-10-20
Refresh plot after resizing image widget

astrofrog:
This fixes a bug that caused images to be heavily pixellated after a resizing, and only updating if the user clicked in the canvas.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,6 +49,8 @@ v0.6 (unreleased)
 * Improved reading of HDF5 files - all datasets in an HDF5 file are now read by
   default. [#747]
 
+* Fix a bug that caused images to not be shown at full resolution after resizing. [#768]
+
 * Fix a bug that caused the color of an extracted spectrum to vary if extracted
   multiple times. [#743]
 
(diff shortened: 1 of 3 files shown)


astrofrog:
I've disabled the regression test on OS X on Travis since for some reason the timer never appears to trigger the timeout. However, things work fine and the test works on my actual OS X desktop, so this is a peculiarity of Travis and might be related to multi-threading.

@ChrisBeaumont - if you have any ideas what this could be, let me know, but in the mean time I plan to merge things once Travis (Linux) and AppVeyor pass.

astrofrog:
This also fails on Windows and on top of that, we're now seeing some unrelated Travis issues linked to conda package updates.

ChrisBeaumont:
> I've disabled the regression test on OS X on Travis since for some reason the timer never appears to trigger the timeout. 

Which timer are you referring to here?

I assume I shouldn't be looking at the diff too closely at the moment, since it seems to be deleting lots of code.

astrofrog:
@ChrisBeaumont - yes, sorry, am in the middle of debugging. The issue seems to be that under certain circumstances, the QTimer for resize_timer doesn't fire off resize_end. I'm currently investigating...

astrofrog:
I've given up debugging for now, there are several issues clouding this, in particular the fact that when debugging, I found that the success of this test depends on whether other Qt tests were run before, so this makes things very confusing. For now, I'll skip the tests on Travis OSX and AppVeyor and get the fix in.

astrofrog:
I'm going to go ahead and merge this since the fix is important, and the regression test does work locally and on Travis with Linux. I can open another issue to keep track of the problem.

PR #769, opened 2015-10-20
New convenience method to get component from ID

astrofrog:
This is a continuation of the discussion from #759. However, it's still not clear to me whether component IDs should be uniquely tied to a specific data component. The following example illustrates that a component ID can point to two separate sets of data values:

```python
In [2]: d1 = Data(x=[1,2,3])

In [3]: d2 = Data(y=[4,5,6])

In [4]: c1 = d1.find_component_id('x')

In [5]: c2 = d2.find_component_id('y')

In [6]: 

In [6]: d1.update_id(c1, c2)

In [7]: 

In [7]: print(d1[c2])
[1 2 3]

In [8]: print(d2[c2])
[4 5 6]
```

In HistogramClient, ``_get_data_components`` actually will return all matches but then ``set_component`` will just choose the first arbitrarily.

@ChrisBeaumont - could you clarify what the intent was in terms of the component IDs, and to what extent they are unique?

### DIFF
--- a/glue/core/data_collection.py
+++ b/glue/core/data_collection.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 from .hub import Hub, HubListener
-from .data import Data
+from .data import Data, IncompatibleAttribute
 from .link_manager import LinkManager
 from .registry import Registry
 from .message import (DataCollectionAddMessage,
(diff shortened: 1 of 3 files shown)


ChrisBeaumont:
component ids should be treated as keys to fetch components. When you link two link two datasets together, you add a component ID from D1 into D2, usually with a derived component that computes whatever the component ID conceptually refers to. In other words, if a componentID is present in two datasets, Glue treats these two components as conceptually the same thing (e.g., both Right Ascension). In your example, ` d1.update_id(c1, c2)` is probably "wrong" in the sense that x and y aren't the same quantity, so they shouldn't be assigned the same ComponentID

astrofrog:
@ChrisBeaumont - ok, thanks. I think this PR is not appropriate because it treats component IDs as being uniquely linked to some data values. I'll close this for now.

PR #771, opened 2015-10-22
Fix PySide segfault

### LINKED ISSUE #770: Segmentation fault with PySide
One of the issues I ran into when debugging https://github.com/glue-viz/glue/pull/768 is that on PySide (on Mac), I get a segmentation fault when opening a new viewer.

To reproduce:

```
conda create -n test python=2.7 glueviz
conda remove pyqt sip
conda install pyside
```

then create a Python script with

```python
import numpy as np
from glue import qglue
qglue(x=np.random.random((100,100)))
```

Run it, then drag the dataset onto the main area and make an imagewidget. This gives a segmentation fault.

gdb shows the following error:

```
Program received signal EXC_BAD_ACCESS, Could not access memory.
Reason: KERN_INVALID_ADDRESS at address: 0x0000000000000008
0x0000000105d820fd in QVariant_isStringList ()
```

EDIT: need to also install the latest developer version of glue once the conda environment is set up

astrofrog:
Use _find_combo_data instead of findData, because PySide crashes if the data is not a string

Closes #770

### DIFF
--- a/glue/qt/tests/test_widget_properties.py
+++ b/glue/qt/tests/test_widget_properties.py
@@ -235,8 +235,9 @@ def test_connect_current_combo():
     assert combo.currentIndex() == 0
 
     # TODO: should the following not return an error?
-    t.a = 'c'
-    assert combo.currentIndex() == 0
+    with pytest.raises(ValueError) as exc:
+        t.a = 'c'
+    assert exc.value.args[0] == 'c not found in combo box'
 
 
 def test_connect_float_edit():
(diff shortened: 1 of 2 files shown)


astrofrog:
Ironically, the failure here is due, I think, to the bugs that #755 fixes :)

ChrisBeaumont:
:+1: as I mentioned elsewhere, we could think about patching the findData method directly, so we are less likely to hit this again in the future

astrofrog:
Fixed in #755

Issue #764, opened 2015-10-19
plugin manager can accumulate stale entries?


ChrisBeaumont:
I tried uninstalling and reinstalling the volume plugin. While the plugin was uninstalled, I still saw the vispy volume renderer as an option in the "manage plugins" window. How is that list computed? Does it remove entries appropriately?

astrofrog:
Ah yes, it's added to ``.glue/plugins.cfg``, need to think about how to deal with uninstalled plugins.

astrofrog:
Fixed in #776 - now only installed plugins are shown in the plugin manager

PR #776, opened 2015-10-22
Make sure we are testing PySide properly

astrofrog:
Conda now installs PyQt4 anytime Matplotlib is installed, so we have to uninstall it to make sure the tests are really running with PySide. This explains why we didn't catch some of the PySide failures sooner.


astrofrog:
This includes a fix for https://github.com/glue-viz/glue/issues/764

PR #797, opened 2015-11-20
Fixed compatibility with the latest stable version of ginga

astrofrog:
cc @ChrisBeaumont @jegpeek @ejeschke


astrofrog:
It turns out the tests weren't running correctly before because we were testing against our fork of ginga which wasn't changing, not the latest stable version of ginga.

PR #805, opened 2015-11-22
Split glue.core.data into glue.core.data, glue.core.component, and glue.core.component_id

astrofrog:
The ``data.py`` file was getting a bit long, so this splits things up more cleanly.


astrofrog:
This includes some other code cleanup related to imports.

PR #809, opened 2015-12-10
a quickie hack to have glueviz use integer indicies such that some pl…

mherkazandjian:
…aces such that the functionality is compatible with numpy indexing where float indicies are not allowed

The current development version of numpy (1.11.0.dev)  does not allow float indexing of arrays
such as x[0.1]

This kind of indexing is done in several places in the codebase of glueviz.

I implemented some fixes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,11 @@ notifications:
 
 env:
   matrix:
-    - PYTHON_VERSION=2.7 MPL_VERSION=1.4 ASTROPY_VERSION=1.0
-    - PYTHON_VERSION=3.4 MPL_VERSION=1.4 ASTROPY_VERSION=1.0
+    - PYTHON_VERSION=2.7 ASTROPY_VERSION=1.0
+    - PYTHON_VERSION=3.4 ASTROPY_VERSION=1.0
   global:
     - PYTEST_ARGS="--cov glue"
+    - MPL_VERSION=1.5
     - NUMPY_VERSION=1.9
     - IPYTHON_VERSION=4
     - NO_CFG_FILES=false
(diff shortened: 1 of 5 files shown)


astrofrog:
@mherkazandjian - thanks! Could you change ``np.int64`` to ``int`` since that should be sufficient?

mherkazandjian:
I chose int64 because for example if an image is too big, then maybe an int64 is more suitable. And this comes at a tiny extra cost. So I don't think it is a big deal.

astrofrog:
@mherkazandjian - actually, int in Python is 64-bit, so it should be the same - I just would prefer to avoid the overhead to the Numpy function versus using the built-in function.

astrofrog:
Though I guess maybe on some 32-bit Pythons, it will be a 32-bit int - but that seems like a corner case these days?

astrofrog:
@mherkazandjian - on second thoughts, you can leave it this way for now. I need to look into the Travis failures though, I think they are unrelated.

mherkazandjian:
I have not used 32-bit python. But i think numpy emulates int64 and handles it internally. Unfortunatly I do not have access to a 32bit OS or 32bit python installation. So unless you think it is necessary, lets keep it like this. Otherwise, i'll replace the int64 with int32.

astrofrog:
@mherkazandjian - could you actually change them to ``np.intp``? That seems to be the correct type for Numpy indices:

http://docs.scipy.org/doc/numpy/user/basics.types.html

If possible, could you also rebase this pull request on the latest master?

mherkazandjian:
ok, i replaced int64 with intp and rebased. the fix causes some tests to fail. it seems more changes need to be done.

astrofrog:
@mherkazandjian - actually the reason it fails now is because we are now properly testing against Numpy dev - your fixes have still reduced the number of failures.

astrofrog:
@mherkazandjian - just to check, do you plan to try and fix the remaining issues, or would you like me to finish this up?

mherkazandjian:
feel free to take over.

Issue #775, opened 2015-10-22
crazy idea for state handling 


ChrisBeaumont:
This is a pretty speculative / potentially unfeasible idea, but:

I've been playing around with react lately (facebook's library for building UIs in javascript). They have a nice philosophy for keeping UI's in proper sync. Each UI component implements a render() method that _completely_ configures the UI from scratch, based on the component's state. React then autodetects all state changes and calls render() as appropriate. 

Now, react has lots of internal fanciness that makes this efficient (the render method manipulates a mock UI where all method calls are cheap, and react converts that into an efficient, minimal set of real DOM manipulations. Figuring out how to do that with Qt could be really tricky). But it avoids a whole class of buts that Glue repeatedly encounters -- namely, we forget to attach the appropriate event listener to update a widget when state changes in some unforeseen way. It's a really nice pattern to have render() be the single place that renders the UI based on the current state, and relying on a separate system to run render when appropriate. 

It might be worth thinking about whether it's possible to nudge glue in this direction -- maintaining a more formal UI state datastructure, organizing display logic around correctly representing this state, and having a more generic (glue-agnostic) system determine when components need to be re-displayed. One concrete sign that we're moving in this direction is that most Qt connect() and emit() calls would be consolidated in one place, and _not_ in GlueApplication, ScatterWidget, ScatterClient, etc. All of that code would be focused solely on manipulating state and how to render that state (the widget classes would still use connect() to tie user input like mouseclicks to state manipulation, but they would not be directly connecting state change events to GUI update handlers -- that logic would be more generally handled elsewhere)

astrofrog:
This is a very interesting idea - just this week, while digging through #755 and #772, I've run into a maze of connections/callbacks which is getting pretty complex, so any way to simplify that would be nice.

So if I understand correctly, all UI components would then store their state in the same way - that it, just as an example, they would all store the state information in a particular attribute? If so, then I assume that the most efficient way to detect state changes would be to actually make the state object able to tell the main application when its state has changed? (rather than having the main application check through all the states).

Having everything determined by state then also makes it trivial to implement undo/redo and saving (it's already simple at the moment, but this would somehow formalize it and could provide it for free for *any* widget).

I guess that having this separation of state and UI then also would make it a lot easier to switch UI framework in future if needed.

[...]

PR #814, opened 2015-12-29
Create documentation about layer artists

astrofrog:
@ChrisBeaumont - while working on this, I ran across a couple of things in the API that would be nice to update:

* ``DataViewer._container``: this name is pretty obscure, and it should maybe instead be ``DataViewer._layer_artists`` or ``DataViewer._layer_artist_container``.

* Maybe the layer artist container should have a consistent name between clients and widgets (at the moment it's ``.artists`` vs ``_container`` for the same object, and ``artist_container`` in some kwargs). What about ``.layer_artists``, and make it public for both clients and widgets?

* I have a mental block when I run across instances of the ``layer`` variable which really is the layer *data*. Would you have any objections to me renaming ``layer`` to ``layer_data``? So then we'd never have just 'layer' - either layer artist or layer data.

What do you think?

Side note: at the moment, we don't really talk about the possibility of using clients in these docs - and that's because in principle one doesn't *have* to use a separate client class. Before writing docs on how to better separate client functionality and GUI, I think it would be good to converge on a solution in https://github.com/glue-viz/glue/issues/775 first, so adding details of clients to the page here is low priority for now in my view.

@PennyQ - just for info, I've started working on this, but it's by no means complete. I'll work on it more over the next couple of days (in my spare time).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,9 @@ v0.7 (unreleased)
 * The ``artist_container`` argument to client classes has been renamed to
   ``layer_artist_container``. [#814]
 
+* Added documentation about how to use layer artists in custom Qt data viewers.
+  [#814]
+
 v0.6 (2015-11-20)
 -----------------
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,9 @@ v0.7 (unreleased)
 
 * Python 2.6 is no longer supported. [#804]
 
+* The ``artist_container`` argument to client classes has been renamed to
+  ``layer_artist_container``. [#814]
+
 v0.6 (2015-11-20)
 -----------------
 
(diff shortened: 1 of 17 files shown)


ChrisBeaumont:
Doc changes LGTM.

> DataViewer._container: this name is pretty obscure, and it should maybe instead be DataViewer._layer_artists or DataViewer._layer_artist_container.

`DataViewer._layer_artist_container` sounds good

> I have a mental block when I run across instances of the layer variable which really is the layer data. Would you have any objections to me renaming layer to layer_data? 

I can't remember, but doesn't `layer` sometimes refer to a Subset as well? In that case `layer_data` sounds misleading to my ears (as I recall I think I started using `layer` as an alias for `subset or data`, but I agree this isn't very obvious)

astrofrog:
> I can't remember, but doesn't layer sometimes refer to a Subset as well? In that case layer_data sounds misleading to my ears (as I recall I think I started using layer as an alias for subset or data, but I agree this isn't very obvious)

Indeed it can refer to a subset. I'll try and think of whether there is a clearer name that would better encapsulate this, but otherwise I'll leave it as-is.

astrofrog:
I'll go ahead and merge this, and will continue working on the docs for the Qt data viewer in future (but what's added in this PR is already fine).

PR #815, opened 2015-12-30
Fix ginga plugin for new ginga API (set_follow_focus -> set_enter_focus)

astrofrog:
@ejeschke - is this the correct fix, or are the two methods not equivalent?


ejeschke:
@astrofrog, the code as you have modified it will work fine with older versions as well as newer versions of ginga.  It looks correct to me.

astrofrog:
@ejeschke - ok, thanks!

Issue #812, opened 2015-12-16
Create combo box that can show 'frequently used' or 'recent' attributes


astrofrog:
And this could then be used for list of data components/attributes.

PR #820, opened 2016-01-11
WIP: Implementation of a combo box with recent items

astrofrog:
### Preview

This looks like:

recent
### Issues

From a UI point of view, there are still some issues, because at the moment if the QComboBox is long it moves up when I click on it, so I have to keep scrolling up to the recent items which defies the point. I have a StackOverflow question open here:

http://stackoverflow.com/questions/34765193/preventing-long-qcombobox-from-moving-up

so we'll see if there are any good solutions.
### Technical details:

The way this works in order to be usable by glue as-is is that it overloads all the methods of QComboBox so that the index (as seen by public methods) is still the index of the original list. There may be better ways of doing this with a custom view or delegate (or model proxy) but I'm not 100% sure. @ChrisBeaumont - do you have any thoughts on this?

Fixes https://github.com/glue-viz/glue/issues/812

### DIFF
--- a/glue/qt/qtutil.py
+++ b/glue/qt/qtutil.py
@@ -396,6 +396,9 @@ def _custom_widgets():
     from glue.qt.link_equation import LinkEquation
     yield LinkEquation
 
+    from glue.utils.qt.qcombobox_recent import QComboBoxRecent
+    yield QComboBoxRecent
+
 
 def load_ui(path, parent=None):
     """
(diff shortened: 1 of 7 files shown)


astrofrog:
I'm going to close this, as I couldn't find a way to avoid the issues with long combo boxes, and it's more useful I think to show people how to use the keyboard to navigate attributes.

PR #827, opened 2016-01-21
Switch to absolute imports in most cases

astrofrog:
This switches all glue imports to absolute imports except in the following cases:

- local imports in ``__init__.py`` files
- local imports in ``test_*.py`` files and imports that only go up one level

The idea is that in the latter case, the tests will usually always be in the same place relative to the files the tests apply for, so in that case it makes sense to use relative imports.

However, in all other cases, when trying to refactor the code, I found that the relative imports really make the code much harder to understand (and harder to move files around).

With absolute imports, it will be much easier to check that certain parts of glue are not imported by other parts (for example nothing in ``glue.core`` should import from ``glue.qt``, etc.).

Incidentally, I'm guilty of switching from absolute to relative initially (in b999181f8f7120abaee928bfe7a94743780ff324, c7b369d3731bdd47e4d12ca2ff360fec8e0b51ae, 50402a6c6925b8b2078a679cae77a3635ee59406, and ccf77e29570a28b0dcdd276a05cc5d84164c305e) but I've now learned from my mistake :grin:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,12 @@ env:
     - PYTHON_VERSION=3.4 ASTROPY_VERSION=1.0
   global:
     - PYTEST_ARGS="--cov glue"
-    - MPL_VERSION=1.5.0
+    - MATPLOTLIB_VERSION=1.5.0
     - NUMPY_VERSION=1.10
     - IPYTHON_VERSION=4
     - NO_CFG_FILES=false
     - QT_PKG=pyqt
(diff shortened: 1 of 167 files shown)


astrofrog:
@ChrisBeaumont - let me know if you have any objections! :)

ChrisBeaumont:
No real objection from me -- I agree that relative imports are harder to grok when they go up more than 1 level, so the extra verbosity serves a purpose.

astrofrog:
@ChrisBeaumont - thanks! I'll now wait until the testing issue is resolved, then will merge.

PR #831, opened 2016-01-22
Move more utilities into glue.utils

astrofrog:
These are utilities that don't depend on any other part of glue, except for ``glue.external``. This will help make the code base a little more approachable, since everything in glue.utils can be treated as standalone functions/classes that can be understood independently.


astrofrog:
This is just some preliminary work towards re-organizing the viewers and other components which will come in a separate PR. As part of this though, I found that the Qt5 backend setting for matplotlib was a bit brittle because it only happened correctly if ``glue.clients`` was the first module where matplotlib was imported - a better approach is now included here which is to have an import hook that sets the Qt5 backend as soon as Matplotlib is imported, regardless of where it's imported from.

PR #834, opened 2016-01-25
ROI refactoring [rebased]

astrofrog:
Rebased version of #692. I'll finish this up here.

### DIFF
--- a/glue/utils/geometry.py
+++ b/glue/utils/geometry.py
@@ -2,7 +2,13 @@ from __future__ import absolute_import, division, print_function
 
 import numpy as np
 
-__all__ = ['polygon_line_intersections']
+__all__ = ['points_inside_poly', 'polygon_line_intersections']
+
+
+def points_inside_poly(x, y, vx, vy):
+    from matplotlib.path import Path
+    p = Path(np.column_stack((vx, vy)))
+    return p.contains_points(np.column_stack((x, y)))
 
 
 def polygon_line_intersections(px, py, xval=None, yval=None):
@@ -10,7 +16,9 @@ def polygon_line_intersections(px, py, xval=None, yval=None):
     Find all the segments of intersection between a polygon and an infinite
     horizontal/vertical line.
 
-    The polygon is assumed to be closed
+    The polygon is assumed to be closed. Due to numerical precision, the
+    behavior at the edges of polygons is not always predictable, i.e. a point
+    on the edge of a polygon may be considered inside or outside the polygon.
 
     Parameters
     ----------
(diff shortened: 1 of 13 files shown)


astrofrog:
I've now implemented polygonal selection for plots with one or two categorical components. What this does is the following:

### One categorical component, one numerical component

In this scenario, we interpret polygonal selection by checking for each category the range (or ranges for non-convex polygons) highlighted by the polygon:

categorical_selection

The ranges indicated in dark red are the actual final selection.

For efficiency, I implemented a ``MultiRangeSubsetState`` which can be used for discontinuous ranges (such as the various ranges covered by the intersection of a non-convex polygon with the category). Each ``MultiRangeSubsetState`` is then combined with a ``CategoricalROISubsetState`` for that specific category with ``AndSubset``, and the resulting subset state is then combined across categories with an ``OrSubset``.

What this means in practice is that categories not shown on the plot are not selected. In future, if we have the option to re-order the categorical components, we'll need to make sure this is properly taken into account here.

### Two categorical components

In this case, for each categorical component on the x axis, we check which categories fall inside the polygon on the y axis.

If any categories on the y axis do fall in the polygon, we then create a ``CategoricalROI`` for those categories, and combine it with ``AndState`` with a ``CategoricalROI`` for the specific category we are looking at on the x-axis. We then combine the subset states for all the categories on the x-axis with an ``OrState`.

### Notes

I believe that with these changes, the selection will be more intuitive. Without this addition, some points not in the polygon ended up being selected, which was counter-intuitive.

There are still failing tests, but I want to make sure that the above reasoning is ok before proceeding.

astrofrog:
(actually the tests were pretty easy to fix)

ChrisBeaumont:
@astrofrog this approach sounds mostly good to me. One thing I wonder about: sometimes it's nice to be able to draw a shape, and then click+drag that shape to get a feel for how small changes in selection impact another plot. It sounds like polygons are being converted into a more abstract quantity here -- do we still have the ability to drag pre-defined subsets as before? (this might be fine -- I can't remember if the shape on a plot was strongly tied to the subset state or not).

glue/utils/geometry.py
```diff
@@ -0,0 +1,82 @@
...
+
+    px = np.asarray(px, dtype=float)
+    py = np.asarray(py, dtype=float)
+
+    # Make sure that the polygon is closed
+    if px[0] != px[-1] or py[0] != py[-1]:
+        px = np.hstack([px, px[0]])
+        py = np.hstack([py, py[0]])
+
+    # TODO: vectorize this later, but need to get the logic for corner cases
```
ChrisBeaumont: yeah, it would be good to vectorize this or figure out a way to hook into MPLs C-level geometry utilities

astrofrog:
Note, for some reason the selection seems broken at the moment, working on it

astrofrog:
@ChrisBeaumont - the selection dragging still works :)

glue/utils/geometry.py
astrofrog: It should be easy to vectorize, but I was getting myself confused with corner cases :) I'll first write some tests (including corner cases), then will vectorize.

astrofrog:
@ChrisBeaumont - thanks for reviewing this! If you have no other comments/suggestions, I will merge this tomorrow/Tuesday. The control-click to drag selections still works, which I think was your main comment.

glue/utils/geometry.py
astrofrog: This is now vectorized in this PR. The matplotlib ``points_inside_poly`` which we rely on doesn't always behave intuitively for points on the boundaries, but this is not a new problem since we were relying on this function before.

PR #835, opened 2016-01-26
Reorganize viewers into glue.viewers

astrofrog:
This reorganizes a lot of the viewer-related files so that each viewer is compartmentalized in glue.viewers. The non-Qt and Qt code are still kept separate (the Qt code is in a ``qt`` sub-directory). Backward-compatibility with previous session files is preserved.

Remaining to-dos:

* [x] Viewers should self-register like plugins
* [ ] <s>Check that Mac app still works</s> (will have to be done later once this is merged)
* [x] Check why ``test_resize`` no longer works


astrofrog:
@ChrisBeaumont - this is the first main step in the reorganization of the code as we discussed before. However, after trying various things, I decided that there's probably no need to have a ``glue.components`` level in the hierarchy, and we can just have things like ``glue.viewers``, ``glue.importers``, etc. Let me know if you disagree.

If everything passes, I'll do some extensive testing locally, and if there are no issues I'll merge before continuing with the reorganization in a separate PR.

doc/developer_guide/api.rst
```diff
@@ -51,13 +51,13 @@ User Interface
 .. automodapi:: glue.qt.widgets.data_viewer
    :no-inheritance-diagram:
 
-.. automodapi:: glue.qt.widgets.scatter_widget
```
ChrisBeaumont: random question -- why `viewer_widget` instead of `viewer` or `widget`? No strong opinion from me

astrofrog: I wanted to include ``widget`` to show it's a Qt widget, and ``viewer`` as opposed to the options widget

ChrisBeaumont: ok I see - to distinguish from `options_widget`. Makes sense

ChrisBeaumont:
What's here looks good to me!

astrofrog:
@ChrisBeaumont - thanks for the review! :)

PR #840, opened 2016-01-28
Move more viewer-related files into glue.viewers

astrofrog:
This finishes moving files related to viewers (especially related to the custom viewer and to classes common to all viewers such as ``DataViewer``) into ``glue.viewers``.


astrofrog:
This is just a continuation of https://github.com/glue-viz/glue/pull/835, so there shouldn't be anything controversial.

PR #842, opened 2016-01-29
Set up ReadTheDocs

astrofrog:
ReadTheDocs now supports conda to set up the environment, so we can now easily install e.g. PyQt4 and other dependencies. This just adds the files for now, but we'll later need to switch the domain to point to ReadTheDocs and disable the S3 builds.


astrofrog:
It worked! http://glueviz.readthedocs.org/en/latest/

PR #844, opened 2016-01-29
Further reorganization of Qt files

astrofrog:
This further splits up the code from ``glue.qt`` into different places:

* ``glue.core.qt`` - this is a new module, and is intended for Qt classes that are tightly coupled with ``glue.core`` objects (e.g the layer artist model). Now this goes against our original rule of now Qt code in glue.core, but I think in this case it makes sense, and Travis is now set up to make absolutely sure that the Qt code is isolated (see https://github.com/glue-viz/glue/pull/843). However, we can still discuss this.

* ``glue.viewers`` - as in previous pull requests, things related specifically to data viewers

* ``glue.dialogs`` - these are fairly general dialogs that could be useful in a different glue app, such as a link editor. These are dialogs that usually change the underlying state of the data

* ``glue.app`` - as in previous pull requests, things that are specifically for the current Glue app that may not be generalized enough to fit in other categories.

There are a few small things left in ``glue.qt``, notably the icon management. The diff is quite hard to read, but the commit messages should hopefully be explicit enough.

As a bonus, there is now a mechanism for registering custom widgets for load_ui to avoid having to hard code it alongside load_ui.


astrofrog:
Once all this is done, I do plan to write some clear docs explaining the new organization of the code, to help direct newcomers who want to contribute!

astrofrog:
I'm going to go ahead and merge this so as to finish the rest of the re-organization today, but we can still discus whether the move of a few core things to ``glue.core.qt`` was a bad idea.

Issue #841, opened 2016-01-28
Consider providing temporary backward-compatibility for some classes/modules


astrofrog:
Following all the code reorganization, we should think whether any of the classes/functions that have been moved are likely to be used in user code and whether we should provide backward-compatibility with deprecation warnings for one version.

astrofrog:
``load_ui``

astrofrog:
Done - will add more if users request.

PR #845, opened 2016-02-01
Finish re-organizing files from glue.qt

astrofrog:
This finishes the reorganization of files from glue.qt and removes the glue.qt submodule for now (as indicated in https://github.com/glue-viz/glue/issues/841 we can then decide to add it back with some compatibility imports if we want to deprecate any of the imports over one version).

The reason for the large diff numbers is that I've removed the Adobe Illustrator file for the icons, replacing it with SVG icons.


ChrisBeaumont:
LGTM -- my only concern would be whether the inkscape -> png translation is adding any weirdness that the illustrator -> png translation didn't. The current PNGs in this branch look good -- were they regenerated from the new SVGs?

astrofrog:
@ChrisBeaumont - no, but I could re-generate them just to compare.

astrofrog:
Here's a view of the differences: https://github.com/astrofrog/glue/commit/7d242efb910e2ebda645439cda658944173ddc16 - I just noticed an artifact in glue_or.svg which I'll remove here. Some of the colors have changed slightly, so I'll see if I can tweak that too.

astrofrog:
Now looks better: https://github.com/astrofrog/glue/commit/1b192b2308c7e2900b9cbb0b30df40521e791a60

Issue #851, opened 2016-02-09
Proposed improvements to layer panel


astrofrog:
While working on the 3D viewers, I've been thinking a bit more about how to deal with cases where for example one wants to customize the properties of specific layers, in viewers that can show multiple kinds of layers.

Let's take the example of the 2D image viewer. In this viewer, we can show an image and overlay a set of points, which means that the viewer now contains an image layer and a scatter layer. So far so good. Now let's say that there are many points in the viewer, so I'd like to make the points translucent. I can go and edit the style of the layer by control-clicking on the layer and changing the transparency. This then propagates back to the main data properties and to all other viewers. However, let's say that in another scatter plot viewer, I was actually fine with not having any transparency. There is no way to decouple the transparency of both plots. Essentially, the 'edit style' for layers is the same as 'edit style' for the whole dataset.

**Proposal 1:** decouple data/subset and layer style - if the style is set at the data/subset level, it applies to all viewers. However, if the style is set at the layer artist level, then it applies only to that layer artist. By default, the layer artist style is set to the data/subset style.

The second problem I would like to solve is that at the moment, for a given layer, there are several places one can edit the style in incompatible ways. For an image layer, I can set the color via the edit style menu on the layer artist (which does nothing, unless it is a subset), and I can also set the colormap via the viewer itself. The stretch limits are also set in the viewer itself.

**Proposal 2:** all options which control the visual appearance of a layer should be set in the layers panel.

Finally, it's not obvious to people that they can control-click on layers in the list of layers

**Proposal 3:** either add a 'gear' symbol to the right of the layer in the list to show that it can be changed, or better, split the layer panel into 2 vertically, with the list on top, and the visual options underneath. The visual options panel would update depending on which layer artist was selected.

Another feature that would be nice to have is the ability to make the color and/or size of points dependent on data values. We've been experimenting this that in the 3d viewer but it would be nice to also have in the 2d version.

**Proposal 4:** once the above proposals are implemented, make the size and color of points controllable in a data-dependent way in the layer artist options. In future, we could also make this available at the data/subset level.

I'm going to try and work on a prototype for Proposals 1-3.

@ChrisBeaumont - do you have any thoughts on this?

astrofrog:
Done in #852

PR #852, opened 2016-02-10
Allow data viewers to define a widget for editing the layer artist style

astrofrog:
These are just the first steps towards something like #851. Very much work in progress, but opening this already to see if Travis or AppVeyor pick up issues as we go.

* [x] include the fixes to echo in that repository
* [x] include the fixes to qt-helpers in that repository
* [ ] <s>remove the edit style menu (except when no layer artist style widget is defined?)</s> - we can do that later

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -117,9 +117,9 @@ v0.7 (unreleased)
 
 * Properties of the VisualAttributes class are now callback properties. [#852]
 
-* ``glue.utils.qt.widget_properties.connect_value`` can now take an optional
-  value_range and log option, to scale the Qt slider values to a custom range
-  of values (optionally in log space). [#852]
+* Add ``glue.utils.qt.widget_properties.connect_value`` function which can take
+  an optional value_range and log option to scale the Qt values to a custom
+  range of values (optionally in log space). [#852]
 
 v0.6 (2015-11-20)
 -----------------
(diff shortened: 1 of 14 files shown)


astrofrog:
@ChrisBeaumont - at this point, I'd be interested in your feedback. Essentially what I've done here is made it so that data viewers can define an attribute ``_layer_style_widget_cls`` that points to a class that is used to create a layer style editor. If this is present, it then gets added below the layer list:

screen shot 2016-02-11 at 5 12 02 pm

This then makes it more clear that the layer properties can be edited, and transparency can be changed in a much more interactive way (useful for volume renderings in particular).

There are several steps from here:

* Implement a style editor for other layer types apart from scatter plots
* Have a toggle to specify whether the layer properties should propagate back to the data/subset style, or be specific to this artist. There might be cases where one wants to change the transparency only for a given layer artist.

What do you think about this so far?

ChrisBeaumont:
This looks nice! I like how it's more obvious how to change layer properties by always displaying the style widget. I'm also on board with letting classes provide their own widget classes here. Do you think this widget should also be the one that's displayed when you right click -> edit style a layer? (or can we remove that option at this point, since it's redundant?)

astrofrog:
@ChrisBeaumont - thanks for the feedback! I was thinking we could remove the 'edit style' once this is fully implemented, since it is indeed redundant.

astrofrog:
@PennyQ - just for information, this PR provides an infrastructure for data viewers to provide their own layer artist style editor, so I'm now going to work on trying to use this for the volume rendering viewer. I hope to have something done about this later today.

PennyQ:
@astrofrog Sounds great! Just keep going :)

PR #853, opened 2016-02-14
data factory for .npy and .npz files

adrn:
I think these are standard enough that it would be worthwhile to natively support loading Numpy save files. This implements loaders for both types of files. I haven't been able to test this in the GUI (how do I run a development version of the GUI?) but the tests pass.


astrofrog:
@adrn - thanks, this is a great idea! How do you have glue installed currently? Via the Mac app, conda package, or pip installed?

glue/core/data_factories/tests/test_numpy.py
```diff
@@ -0,0 +1,43 @@
...
+    data2 = np.array([("c",15.2352,-2.513), ("d",2.412,3.1341)],
+                     dtype=[('name','|S1'),('l','f8'),('b','f8')])
+
+    with tempfile.NamedTemporaryFile(suffix='.npz') as f:
+        np.savez(f, data1=data1, data2=data2)
+        f.seek(0)
+
+        data_loaded = df.load_data(f.name)
+
+        arr = data_loaded[0]
```
astrofrog: I think the reason for the failure on Travis is that the order of the datasets may not be deterministic. Indeed, on MacOS X with Python 3.4, the datasets are swapped:

```
In [2]: load('test.npz')
Out[2]: <numpy.lib.npyio.NpzFile at 0x101e14cc0>

In [3]: load('test.npz').files
Out[3]: ['data2', 'data1']
```

Maybe the easiest would be to add ``sorted`` in `` for groupname in npy_data.files:`` to make sure the order of the datasets is always the same?

adrn:
@astrofrog I think I installed it via conda

RE: 3.4 errors -- got it! Will give that a try

adrn:
@astrofrog any ideas on the windows failures? looks like it's a tempfile write permission issue?

astrofrog:
@adrn - you can just install the latest developer version as you would astropy. For instance, ``pip install https://github.com/glue-viz/glue/archive/master.zip`` should work. Maybe you can conda uninstall glueviz first (but all the dependencies will stay installed).

For Windows, I've seen issues in the past with ``tempfile`` - maybe instead you can use the pytest ``tmpdir`` argument (we use this in various places in astropy)

glue/core/data_factories/tests/test_numpy.py
```diff
@@ -0,0 +1,43 @@
...
+        for name in data.dtype.names:
+            assert_array_equal(data[name], data2[name])
+
+def test_npz_load(tmpdir):
+    data1 = np.array([("a",152.2352,-21.513), ("b",21.412,35.1341)],
+                     dtype=[('name','|S1'),('ra','f8'),('dec','f8')])
+    data2 = np.array([("c",15.2352,-2.513), ("d",2.412,3.1341)],
+                     dtype=[('name','|S1'),('l','f8'),('b','f8')])
+
+    with tempfile.NamedTemporaryFile(suffix='.npz', dir=tmpdir.strpath) as f:
```
astrofrog: Don't use ``tempfile`` at all - instead do:

```
with open(tmpfile.join('myfile.npz').strpath, 'wb') as f:
    ...
```

tempfile has some issues on Windows that I don't understand.

adrn:
Yea, weird, I thought I'd be able to get away with using it if we used tmpdir because it should be guaranteed to be writable. Ok I'll do it the old fashioned way :)

adrn:
@astrofrog cool -- looks like getting rid of tempfile was the way to go. Thanks!

astrofrog:
@adrn - thanks!

PR #871, opened 2016-02-26
WIP: generic limits widget

astrofrog:
This implements the logic of recomputing limits when changing attributes, using pre-sets to compute the limits, etc:

screen shot 2016-02-26 at 6 41 16 pm

However, this is proving to be a bit annoying layout wise, so an alternative would be to make this into a helper connect function but not actually a widget.


astrofrog:
Superceded by https://github.com/glue-viz/glue/pull/872

PR #872, opened 2016-02-29
Add a helper to link widgets related to setting limits

astrofrog:
For a number of widgets, we need to implement the logic of caching attribute limits for different data attributes, linking an autoscaling option to the limits, flipping the limits, etc. This PR adds a helper class to manage this in one place.

This is an alternative to https://github.com/glue-viz/glue/pull/871 - the advantage of the helper class is that it does not impose any particular layout (whereas the layout in #871 did not always fit in to particular widgets). It also allows some of the things, such as the scale mode and flip button, to be optional.

An example of using this is in https://github.com/glue-viz/glue-3d-viewer/pull/83 - in particular, note that the Python files there are significantly simplified.

The helper here can also be used inside Glue core in various places, such as the image layer artist editor, or histogram limits for instance.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,9 @@ Full changelog
 v0.8 (unreleased)
 -----------------
 
-* No changes yet
+* Added a new helper class ``AttributeLimitsHelper`` to link widgets related to
+  setting limits and handle the caching of the limits as a function of
+  attribute. [#872]
 
 v0.7 (2016-03-10)
 -----------------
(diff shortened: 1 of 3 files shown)


astrofrog:
@ChrisBeaumont - if you have a chance to take a look at this in the next couple of days, it would be great to have your feedback about this approach. Essentially the general question is how to link sets of widgets together via some logic - another common case would be a QSlider + QLineEdit, which could benefit from a similar class.

PR #896, opened 2016-03-24
Gracefully skip vector columns from FITS files

astrofrog:
Important for e.g. event files


astrofrog:
@eblur - once merged, this will allow event files to be read in!

PR #910, opened 2016-03-30
Update dendro.rst

PennyQ:
Remove the following part:

Enabling Dendrogram Visualization
---------------------------------

Because dendrogram visualization is still experimental, you
must enable it before using it. To do so, add the following
line to your :ref:`~/.glue/config.py <configuration>` file::

    from glue.qt.widgets import enable_dendrograms
    enable_dendrograms()


astrofrog:
Thanks @PennyQ!

PR #914, opened 2016-04-01
Glue labeled as Untitled Window on Linux, has no label on Windows 7

astrofrog:
My machine was RHEL6 (64-bit). I installed Glue under Anaconda (Python 2.7, Qt4). On the task bar, Glue is labeled as "Untitled Window", as shown below:

screenshot-1

On Windows 7, it appears like this:

untitled

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,11 @@ v0.8 (unreleased)
   setting limits and handle the caching of the limits as a function of
   attribute. [#872]
 
+v0.7.2 (unreleased)
+-------------------
+
+* Make sure main window title is set. [#914]
+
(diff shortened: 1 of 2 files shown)


astrofrog:
@pllim - good catch, code attached! Can you let me know if it fixes it?

pllim:
:+1: Looks like  this now on Linux. Pretty sure it will work on Windows too, but I can check for sure on Monday. Thanks!

untitled

astrofrog:
@pllim - thanks for confirming! I'll merge this now, but please open a new issue if it still doesn't work on Windows :)

pllim:
Confirmed working on Windows too.

PR #952, opened 2016-04-19
Fix QColorBox so that it resizes correctly

astrofrog:


astrofrog:
This is not the correct fix

PR #956, opened 2016-04-19
Add support for math and numpy modules in 'new component' window

astrofrog:
Also improve layout of window and add instructions.

To-do:

* [x] evaluate expression on-the-fly for single data element to check for syntax issues
* [x] emit errors/warnings inside window instead of with separate pop-ups
* [x] only allow to click OK once everything is set properly


astrofrog:
Just need to add a test for ``evaluate_test`` before I merge

PR #963, opened 2016-05-01
Fix bug which caused Component.numeric to not always work

astrofrog:
I can't remember which data file triggered this, but once I figure out I'll add a regression test.


astrofrog:
Ok, so the use case for this was that if in future ``Component`` included non-numerical data, then ``data[0]`` could end up being a string and that would fail, so this is a safer way to do it.

PR #964, opened 2016-05-01
WIP: Refactor histogram viewer

astrofrog:
Now that there are a number of convenience classes to help deal with attributes and limits caching, I thought it would be a good time to see if the histogram viewer can be simplified. In particular, I want to investigate whether we really gain much by having a separate client class.

Before I do this, it would be good for me to review the basic design we want to follow - in particular, one of the things I learned from the 3D viewers is that code to do with internal logic relating to the options widgets should be contained in a separate class (like `HistogramOptionsWidget` here).


astrofrog:
Superseded by https://github.com/glue-viz/glue/pull/1127

PR #971, opened 2016-05-02
Playback controls for slices

astrofrog:
This is a work in progress - I don't want to add a dependency to fontawesome, so once we are happy with this I'll make static PNGs or SVGs for the icons. At the moment this looks like:

screen shot 2016-05-02 at 4 58 06 pm

The first and last button go to the first and last slice
The second and second to last button step backwards or forwards once 
The third and sixth icon play and speed up in different directions
The fourth icon stops everything
The fifth icon plays at the default speed


astrofrog:
The latest version looks like:

screen shot 2016-05-03 at 3 19 22 pm

So it's a little bit simpler. On the right of the stop button, the buttons are play (and by clicking multiple times one can play faster), then step forward, and step to end. The ones of the left of stop are just the reverse equivalents.

Issue #959, opened 2016-04-21
Inconsistent naming of pixel axes


astrofrog:
If a FITS file is 2-dimensional, the pixel axes are called **Pixel x** and **Pixel y**. However, if I load a 4D cube, the pixel axes are called **Pixel Axis 0**, **Pixel Axis 1** and so on, *and* the order is reversed, which makes it very confusing to set up links.

astrofrog:
Fixed in #1029

Issue #958, opened 2016-04-21
Advanced use case of linking that it would be nice to support


astrofrog:
Here's a use case that a user asked about:

- Load up a PPV data cube
- Load up a moment 1/average velocity
- Make an image and histogram of the average velocity
- Link the pixel positions of the average velocity map with the two matching pixel coordinates for the cube

Now in principle, every pixel in the velocity map can be matched to a 1D spectrum in the cube, so selecting a region in the velocity map selects the corresponding region in the cube (projected along the velocity axis). So far so good.

Now the user selects pixels of the average velocity map from the histogram of average velocities. Since the selection isn't done along a linked component anymore, the selection doesn't propagate to the cube. However, one could imagine that since essentially pixels are being selected by the histogram selection, the selection could then propagate to the equivalent pixels in the other cube.

I think the solution for this to work is to implement the ability to define 1-to-1 links (not coordinate range links) and then link pixels from one data 1-to-1 with the other dataset.

See https://github.com/glue-viz/glue/issues/798 for more details.

PR #974, opened 2016-05-02
Join on multiple keys

astrofrog:
I realized today that we can solve the linking case described in https://github.com/glue-viz/glue/issues/958 by extending ``join_on_key`` to support linking on pairs, triplets, and n-tuples of components rather than just single components. The current PR implements a very rough (and very slow) version of that, just to test the concept. If things work well, then I can work on a way to make the calculation of the masks much faster. Here's an example in use:

We have a position-position-velocity cube and a second moment map (on the same celestial pixel scale).

```python
from glue.core import DataCollection, Data
from glue.core.data_factories import load_data
from glue.viewers.image.qt.viewer_widget import ImageWidget
from glue.viewers.histogram.qt.viewer_widget import HistogramWidget

data_linewidth = load_data('linewidth.fits')
data_cube = load_data('ppv_cube.fits')

dc = DataCollection([data_linewidth, data_cube])

# We can now link on pairs of components
data_linewidth.join_on_key(data_cube, ('Pixel x', 'Pixel y'), ('Pixel Axis 2', 'Pixel Axis 1'))

ga = GlueApplication(dc)

image = ga.new_data_viewer(ImageWidget)
image.add_data(data_cube)

histogram = ga.new_data_viewer(HistogramWidget)
histogram.add_data(data_linewidth)

ga.exec_()
```

Selecting in the linewidth histogram will now highlight pixels in the image.

Note that the fact the pixel axes have different naming conventions needs to be fixed (see https://github.com/glue-viz/glue/issues/959)

@ChrisBeaumont - do you have any thoughts on this approach so far?

### DIFF
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -195,7 +195,7 @@ class Data(object):
         There are several ways to use this function, depending on how many
         components are passed to ``cid`` and ``cid_other``.
 
-        **Single component for each dataset**
+        **Joining on single components**
 
         First, one can specify a single component ID for both ``cid`` and
         ``cid_other``: this is the standard mode, and joins one component from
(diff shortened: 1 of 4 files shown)


astrofrog:
Note that at the moment this is not going to be backward-compatible in terms of loading save files.

PR #936, opened 2016-05-03
When loading data into glue, merge window appears multiple times

astrofrog:
<s>Not 100% sure how to reproduce this, but seems to happen when loading data through the UI not command-line.</s> - fixed with code attached.


astrofrog:
The Travis failure is unrelated

PR #976, opened 2016-05-03
Try and get rid of sporadic failures due to travis command

astrofrog:


astrofrog:
Haven't seen these for a while so I think it's ok.

PR #1000, opened 2016-05-23
Fix bug that caused images to be over-pixellated when zooming in

astrofrog:
@robintw - this should fix the issue you were seeing in https://github.com/glue-viz/glue-geospatial/issues/11


robintw:
I've tested this, and can confirm it fixes it - thanks for the incredibly prompt fix! :-)

PR #1004, opened 2016-05-25
Fix bug with @memoize if a keyword argument was unhashable

astrofrog:


astrofrog:
The only Travis failure is with numpy-dev, which is a real Numpy regression (https://github.com/numpy/numpy/issues/7674)

PR #934, opened 2016-05-25
On Windows, drag and drop file onto Glue app raises error

astrofrog:
It seems to add forward slashes into path?


astrofrog:
Code attached - I'll try and add a regression test later.

PR #1014, opened 2016-06-08
Fix bug in ComponentIDComboHelper

astrofrog:
The numeric and categorical options were ignored in ``__init__``.

This needs a changelog entry and regression test.


astrofrog:
The failure is unrelated

PR #1017, opened 2016-06-10
WIP: fix issues with setting the matplotlib backend with the latest developer version of matplotlib

astrofrog:
Somehow, the existing mechanism breaks, so I need to investigate more


astrofrog:
I think it's working ok now

PR #1019, opened 2016-06-14
Added a new roadmap page to the developer docs

astrofrog:
We can add to this over time.

@PennyQ - do you have any thoughts/comments on this so far?


PennyQ:
**Support for big/complex data: an abstract data and computation interface**
If I understand correctly, our data is loaded in memory and stored in the Data object. And we would like to define three (or more) types of API for data access (to both data values and subsets, and support different formats and will return a uniformed array), data display (return image/rendering buffer for viewers to show) and data calculation (provides access to other calculation/analysis packages). But I feel the later two types (display and calculation) is built on the first API that they need to get the data first? In this way they are not in a flat layer but rather a layer with hierarchy.

**Support for big data: more efficient viewers**
I personally think ‘volume rendering’ for big data is more critical and challenging, since for scatter points users have more options to display (like Paraview), but 3D real-time rendering software for volume data is rare, and that’s something we should and could improve with big data.

**Glue in the browser**
This would be awesome!! Although I feel it will rewrite most of Glue code :(

philastrophist:
Yes, this! Big data support and irregular grid support are the only things stopping me from doing all my visualisations in glue! I currently use np.memmap but I'd prefer to use my own objects since the data I use can be irregular and they have handy methods. I've forked glue myself and am currently implementing my workarounds!

For example, I sometimes want to compute and then visualise probability distributions from MCMC chains/the chains themselves. The chains for each id sometimes have different lengths and at the moment they are stored in hdf5 and read into a tailor-made object for visualisation. So difficulties lie with irregular arrays, big data (sometimes the chains are too massive for memory) and accessing methods (like calculate pdf)  from custom objects.

astrofrog:
@PennyQ @philastrophist - thanks for the feedback! I'm going to go ahead and merge this now, and we can hopefully continue discussion about some of these topics on the mailing list!

Issue #990, opened 2016-05-17
Switch combobox items name and labels for image viewer?


PennyQ:
It may make more sense to use `x, y, slice` as the combobox label and real dataset column name in the combobox (as shown below), while I check the code in [image widget_viewer](https://github.com/glue-viz/glue/blob/master/glue/viewers/image/qt/viewer_widget.py) and couldn't find the lines to work on this, could you show me how to do it @astrofrog ? Thanks!

![image](https://cloud.githubusercontent.com/assets/13411839/15339430/69e66922-1c52-11e6-806c-a1c6508d7099.png)

astrofrog:
@PennyQ - if we do that we need to also do a bit of work to make sure we show the correct sliders when there are 4+ dimensions. This is all done I believe in:

https://github.com/glue-viz/glue/blob/master/glue/viewers/common/qt/data_slice_widget.py

astrofrog:
An I agree that listing it as:

x:
y:
slices:

makes more sense.

PennyQ:
@astrofrog Hmmm I didn't expect this would be a more complicated question, I thought it's just change the text for both labels and combobox, and my current problem is that I couldn't find the codes to assign the text to both. Would you prefer to show me where I can start or you could take care of it? :)

astrofrog:
Done in the re-written image viewer :)

PR #1024, opened 2016-06-19
Switch xyz label

PennyQ:
for #990 and see the figure below:
image

TODO: I don't have data with 4 dim so not sure if it works.

### DIFF
--- a/glue/viewers/common/qt/data_slice_widget.py
+++ b/glue/viewers/common/qt/data_slice_widget.py
@@ -22,7 +22,7 @@ class SliceWidget(QtGui.QWidget):
     slice_changed = QtCore.Signal(int)
     mode_changed = QtCore.Signal(str)
 
-    def __init__(self, label='', pix2world=None, lo=0, hi=10,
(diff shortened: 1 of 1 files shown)


astrofrog:
@PennyQ - thanks for working on this! Note that there is an unrelated commit here again.

I think you should easily be able to use Numpy to make a 4, 5, 6, etc. dimensional cube, save it as FITS, and read it in. I don't think it matters if the world coordinate system is not set, it should call the axes World Axis 1, World Axis 2, etc.

PennyQ:
Yeah, but the `SliderBar` doesn't get updated when the CurrentCombobox item is changed, for both `slice` and `4th dim` axis :(

I don't understand how the update works coz it seems the connect Function is assigned with an output value as shown below? Thanks! @astrofrog 
> mode.currentIndexChanged.connect(lambda x: self.mode_changed.emit(self.mode))

astrofrog:
I'll try and see how we can fix this.

astrofrog:
This has now been incorporated/superseded by https://github.com/glue-viz/glue/pull/1289

PR #1029, opened 2016-06-29
Performance improvements for selection

astrofrog:
This is a WIP, but I'm going to collect several performance improvements for selections.

PR #1023, opened 2016-06-17
Fix slow spectra

PennyQ:
This PR will fix the slow extraction of spectrum tool problem, as in #1021

PR #1032, opened 2016-07-04
Fix slow spectra [improved]

astrofrog:
@PennyQ - this is a continuation of your work in https://github.com/glue-viz/glue/pull/1023 (I have included your commits here, squashed to one commit). Basically in https://github.com/glue-viz/glue/pull/1029 I implemented a general solution for computing a mask from a ROI in an efficient way (inspired by your changes in #1023) and so I've now updated the spectrum tool to use this rather than duplicate the masking logic. I think it should be easier to read now.


astrofrog:
The failures are unrelated to this pull request

Issue #1037, opened 2016-07-14
ValueError: max() arg is an empty sequence of PV diagram extractor


PennyQ:
> Traceback (most recent call last):
>   File "/Users/penny/anaconda/lib/python2.7/site-packages/matplotlib/backends/backend_qt5.py", line 313, in keyPressEvent
>     FigureCanvasBase.key_press_event(self, key, guiEvent=event)
>   File "/Users/penny/anaconda/lib/python2.7/site-packages/matplotlib/backend_bases.py", line 1847, in key_press_event
>     self.callbacks.process(s, event)
>   File "/Users/penny/anaconda/lib/python2.7/site-packages/matplotlib/cbook.py", line 563, in process
>     proxy(_args, *_kwargs)
>   File "/Users/penny/anaconda/lib/python2.7/site-packages/matplotlib/cbook.py", line 430, in **call**
>     return mtd(_args, *_kwargs)
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/viewers/common/qt/mouse_mode.py", line 282, in key
>     self._finish_roi(self._last_event)
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/viewers/common/qt/mouse_mode.py", line 177, in _finish_roi
>     self._roi_callback(self)
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/plugins/tools/pv_slicer/qt/pv_slicer.py", line 32, in _extract_callback
>     self._build_from_vertices(vx, vy)
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/plugins/tools/pv_slicer/qt/pv_slicer.py", line 35, in _build_from_vertices
>     pv_slice, x, y, wcs = _slice_from_path(vx, vy, self.widget.data, self.widget.attribute, self.widget.slice)
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/plugins/tools/pv_slicer/qt/pv_slicer.py", line 186, in _slice_from_path
>     ind = _slice_index(data, slc)
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/plugins/tools/pv_slicer/qt/pv_slicer.py", line 226, in _slice_index
>     key=lambda x: data.shape[x])
> ValueError: max() arg is an empty sequence

astrofrog:
I think this is fixed

PR #1039, opened 2016-07-15
add float for slc instance checking

PennyQ:
this should fix the #1037


astrofrog:
I don't think this is the correct solution - we need to figure out why the slice item is not an int in the first place and fix it there.

PR #1059, opened 2016-08-01
Allow world coordinates to be shown on sliders

### LINKED ISSUE #1054: Show real velocity value instead of channel on 2D image viewer
Can we show real velocity value instead of channel here? I found when I know the regions of specific velocity range then I have to try the slider bar really adjust to it ;)

image

astrofrog:
@PennyQ - this fixes #1054

Basically, this will add a checkbox for each slider that allows the user to toggle between pixel and world coordinates:

screen shot 2016-08-01 at 3 39 32 pm screen shot 2016-08-01 at 3 39 43 pm

If the world coordinates are not well aligned with the pixel coordinates, a warning is shown:

screen shot 2016-08-01 at 3 39 52 pm

@PennyQ - can you test this out and let me know if it works for you?

### DIFF
--- a/glue/viewers/common/qt/data_slice_widget.py
+++ b/glue/viewers/common/qt/data_slice_widget.py
@@ -5,32 +5,39 @@ import os
 from functools import partial
 from collections import Counter
 
+import numpy as np
+
+from glue.core import Coordinates
 from glue.external.qt import QtGui, QtCore
 from glue.utils.qt import load_ui
 from glue.utils.qt.widget_properties import (TextProperty,
-                                       ValueProperty,
-                                       CurrentComboProperty)
+                                             ButtonProperty,
+                                             ValueProperty,
+                                             CurrentComboProperty)
 from glue.utils import nonpartial
 from glue.icons.qt import get_icon
 
 
 class SliceWidget(QtGui.QWidget):
+
     label = TextProperty('_ui_label')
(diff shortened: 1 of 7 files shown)


glue/viewers/common/qt/data_slice_widget.py
```diff
@@ -94,18 +102,50 @@ def __init__(self, label='', pix2world=None, lo=0, hi=10,
...
+            self._ui_slider.label_warning.hide()
+        self._ui_slider.label.setText(text)
+
+    def set_slider_from_label(self):
+        text = self._ui_slider.label.text()
+        if self.use_world:
+            # Don't want to assume world is sorted, pick closest value
+            value = np.argmin(np.abs(self._world - float(text)))
+            self._ui_slider.label.setText(str(self._world[value]))
+        else:
```
PennyQ: @astrofrog Can you tell me what does this value mean here?
`value = np.argmin(np.abs(self._world - float(text)))` and why there is a 
`self._world-float(text)`

astrofrog: self._world is an array of world coordinates, and text is the value that the user enters in the field. Since it's a string it needs to be converted to a float. So then np.abs(self._world - float(text)) finds how far away the value the user entered is from all the values in the array, and np.argmin finds the index of the world array with the smallest difference. So basically this is finding the closest value in the array.

PR #1063, opened 2016-08-07
Fix to Ginga plugin for colorbar/readout changes

ejeschke:
@astrofrog, this is a fix for [this issue](https://github.com/ejeschke/ginga/issues/366).

I think it might be reasonable to update the plugin to remove some of the bits for old (>1 year) versions of ginga.  If you agree I can submit another PR with those changes.

BTW, is there a way to know which version of Qt glue is using (4, 5 or pyside?)

### DIFF
--- a/glue/plugins/ginga_viewer/qt/viewer_widget.py
+++ b/glue/plugins/ginga_viewer/qt/viewer_widget.py
@@ -5,7 +5,13 @@ import os.path
 
 import numpy as np
 from ginga.misc import log
-from ginga.qtw import ColorBar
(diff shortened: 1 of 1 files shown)


astrofrog:
@ejeschke - thanks! I think it would be fine to remove the bits for old versions of ginga. I also think it would still make sense to extract the ginga plugin into a separate package so that we can make releases faster when either glue or ginga change API (as you mentioned in https://github.com/ejeschke/ginga/issues/231). Let's merge this in first and I can then pull this out into a third-party package.

PR #1066, opened 2016-08-08
Fit PV diagram into a stream of points

PennyQ:
In PV diagram I extracted pixels with max value along `position` axis, and get their world coordinate (including velocity as shown in the histogram below).

Now the data is stored in a .dat file and reloaded to Glue and get linked, could we create a new layer artist for these points data? I found `VertexROIBase` from `glue.core` but I don't know how to plot them. @astrofrog 
*Further: a graph layer artist that accepts edge info too!

image


astrofrog:
We definitely don't want to be writing the dat file. I think maybe the PV slicer tool could have a drop-down menu to choose between PV slice and maximum (or other kinds of calculations along the path).

What we really need to do is that when a PV slice or a profile is created, we should make a new Data object and add it to the data collection, then make a new viewer open up.

astrofrog:
I'm going to close this due to lack of activity but this is definitely something I think we should explore again once I've had a chance to refactor how the PV slicing is handled (i.e. with derived datasets)

PR #1070, opened 2016-08-09
Smart link data

PennyQ:
`Smart link` will find components that are considered as the same with one click, based on comparing their labels in our predefined `LINK_ARRAY`, this array will also get updated or learn from user's manual `glue` / `unglue` manual. (that's why it's smart!) 

image

### DIFF
--- a/glue/config.py
+++ b/glue/config.py
@@ -618,3 +618,6 @@ settings.add('DATA_ALPHA', 0.8, validator=float)
 settings.add('BACKGROUND_COLOR', '#FFFFFF')
 settings.add('FOREGROUND_COLOR', '#000000')
 settings.add('SHOW_LARGE_DATA_WARNING', True, validator=bool)
+from sets import Set
+settings.add('LINK_ARRAY', [Set(['GLON', 'Glon_deg']), Set(['GLAT', 'Glat_deg']), Set(['VHCO', 'Vel_kms'])])
+# TODO: how to save this settings to disk?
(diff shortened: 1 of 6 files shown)


glue/config.py
```diff
@@ -618,3 +618,6 @@ def _default_search_order():
 settings.add('BACKGROUND_COLOR', '#FFFFFF')
 settings.add('FOREGROUND_COLOR', '#000000')
 settings.add('SHOW_LARGE_DATA_WARNING', True, validator=bool)
+from sets import Set
+settings.add('LINK_ARRAY', [Set(['GLON', 'Glon_deg']), Set(['GLAT', 'Glat_deg']), Set(['VHCO', 'Vel_kms'])])
+# TODO: how to save this settings to disk?
```
PennyQ: @astrofrog How can I save this Link_array to disk? Now it will return default value after reopen glue :(

astrofrog:
@PennyQ - I like the idea of having auto-linking, but I think that we should handle this by having a registry for auto-linking functions so that people can register these via the ``config.py`` file. You can see examples of other kinds of registries in ``glue/config.py``. Can you try and set up a registry for auto-linking functions?

PennyQ:
@astrofrog I added the register for auto-linking but it's still a test PR code, two main questions are:
1)  how to save this link_array to disk and keep extending it?
2) remove links not work now -> how to get two individual components from result ListWidget (the right panel), it's set as `core.component_link.ComponentLink( )`

Also add test code for `link_editor.py`

PennyQ:
or we should leave the setting of link_array to users in `~/.glue/config.py`? then I could finish this pr more easily and add an instruction on http://www.glueviz.org/en/stable/customizing_guide/customization.html

astrofrog:
Closing this as we now have an auto-linking infrastructure in glue

PR #1085, opened 2016-08-19
Refactor toolbar code

astrofrog:
The main aim is to decouple the Matplotlib-specific stuff from other kinds of toolbars with custom modes. This still needs:

* [x] Do modes need a close method?
* [ ] Fix PV slicer for Ginga viewer
* [ ] Fix Spectrum extractor for Ginga viewer
* [x] Fix existing tests
* [x] Documentation
* [x] Tests
* [x] Spectrum tool missing
* [x] PV slicer view missing toolbar


astrofrog:
This still needs:

* [ ] Documentation
* [ ] Tests
* [ ] Make things in base mode classes abstract

astrofrog:
@pllim - just in case you are interested, this is the PR that refactors how tools for viewers are dealt with. I don't expect you to review all of this, but if you fancy taking a look at the new ``doc/customizing_guide/toolbar.rst`` file and letting me know if it seems clear enough, I would appreciate it. Thanks!

doc/customizing_guide/toolbar.rst
```diff
@@ -0,0 +1,205 @@
...
+
+The class-level variables set at the start of the class are as follows:
+
+* ``icon``: this should be set either to the name of a built-in glue icon, or
+  to the path to a PNG file to be used for the icon. Note that this should
+  **not** be a ``QIcon`` object.
+
+* ``tool_id``: a unique string that identifies this tool. If you create a
+  tool that has the same ``tool_id`` as an existing tool already implemented in
+  glue, you will overwrite the existing tool. This is never shown to the user.
```
pllim: But how would a developer know? Is it more fail-safe or possible for Glue to internally append some number (e.g., sequence number) to the given tool ID to avoid conflict?

doc/customizing_guide/toolbar.rst
```diff
@@ -0,0 +1,205 @@
...
+  at the bottom of the viewer whenever the button is active.
+
+* ``tool_tip``: this should be a string that will be shown when the user hovers
+  above the button in the toolbar. This can include instructions on how to use
+  the tool.
+
+* ``shortcut``: this should be a string giving a key that the user can press
+  when the viewer is active, which will activate the tool. This can include
+  modifier keys, e.g. ``'Ctrl+A'`` or ``'Ctrl+Shift+U'``, but can also just be
+  a single key, e.g. ``'K'``.
```
pllim: Does the key binding also overwrite one another like `tool_id`? Does Glue have a way to tell the user what key is actually bound to what after start up?

doc/customizing_guide/toolbar.rst
```diff
@@ -0,0 +1,205 @@
...
+
+When the user presses the tool icon, the ``activate`` method is called. In this
+method, you can write any code including code that may for example open a Qt
+window, or change the state of the viewer (for example changing the zoom or
+field of view). You can access the viewer instance with ``self.viewer``.
+Finally, when the viewer is closed the ``close`` method is called, so you should
+use this to do any necessary cleanup.
+
+The ``@viewer_tool`` decorator tells glue that this class represents a viewer
```
pllim: Personally, I like to either capitalize a package name or to mark it like `glue`, so it is obvious you are referring to the package, and not the sticky kind. But please ignore if you disagree.

pllim:
Thanks, @astrofrog . I left some comments on the doc.

Also, would it be overkill to add a line saying something like, "If you are interested to implement a custom tool, you may look at <link to code repo for tools> for the tool that closely resemble your desired functionality." ?

doc/customizing_guide/toolbar.rst
astrofrog: @pllim - yes, maybe it would be better to make it raise an error (I would prefer that to automatically changing the ID).

astrofrog:
@pllim - thanks for reviewing this! I agree with your comments and will fix all this :)

doc/customizing_guide/toolbar.rst
pllim: Error would be clearer than silently overwriting, so :+1: for error.

doc/customizing_guide/toolbar.rst
astrofrog: We deliberately refer to glue as lowercase for now, but I agree we should maybe change this in future.

doc/customizing_guide/toolbar.rst
astrofrog: Yes, I've now clarified this, and the tool tips now include the shortcut so the users can find out more easily.

doc/customizing_guide/toolbar.rst
astrofrog: Done

astrofrog:
@pllim - there aren't many tools developed yet outside the core repo, but once there are enough, I'll add a link as you suggest.

pllim:
Looks good, @astrofrog . I notice that you changed the tool IDs in the doc. Are there any syntax constraints in naming the IDs (e.g., no dashes, must start with `select` etc then separated by `:`, etc)? If so, they should be documented as well.

astrofrog:
No hard restrictions, I'm just trying to be tidy. If in future I add any requirements, I'll list them

astrofrog:
I'll fix the ginga slicing and spectral modes in a separate PR

PR #1090, opened 2016-08-23
[WIP] Add a colorbar to image viewer

### LINKED ISSUE #1087: Add colorbar to image viewer
As part of the JWST DADF sprint, I am looking into how to add a colorbar to the default image viewer in Glue. I traced the `imshow` commands to `glue/viewers/image/layer_artist.py`, where `self._axes.imshow(...)` is always directly appended to a list of `artists`. It is not obvious where to insert a colorbar as indicated in Matplotlib tutorial:

```python
fig, ax = plt.subplots()
cax = ax.imshow(data, ...)
cbar = fig.colorbar(cax, ...)
```

I don't see why a colorbar has to be an optional button to get displayed. I think it should just *always* be displayed with an image is shown.

pllim:
Fixes #1087. Colorbar values seem to update properly when I change the scaling.

Todo:
- [ ] How to implement colorbar layout properly so that the axes get repositioned without overlap when the viewer is resized?
- [ ] How to hook up colormap toolbar to change the color of the colorbar? Currently, when a different colormap is selected, the colorbar is still in black and white.

@astrofrog , I am starting to think that maybe this is not the right approach. Perhaps to properly do this, Glue needs a dedicated space for colorbar in the lower levels. My current maplotlib-fu and glue-fu are insufficient to accomplish anything more complicated than what is on this PR. As a comparison, in Ginga, colorbar is actually a separate plugin but it is just always there.

Somewhat better result with vertical colorbar:
screenshot

However, when viewer is maximized, colorbar still "eats" into the image axis:

Ugly attempt with horizontal colorbar:
screenshot

### DIFF
--- a/glue/viewers/image/layer_artist.py
+++ b/glue/viewers/image/layer_artist.py
@@ -178,12 +178,22 @@ class ImageLayerArtist(MatplotlibLayerArtist, ImageLayerBase):
         for v in views:
             image = self._extract_view(v, transpose)
             extent = get_extent(v, transpose)
-            artists.append(self._axes.imshow(image, cmap=self.cmap,
-                                             norm=self.norm,
-                                             interpolation='nearest',
-                                             origin='lower',
-                                             extent=extent, zorder=0))
+            cim = self._axes.imshow(image, cmap=self.cmap, norm=self.norm,
+                                    interpolation='nearest', origin='lower',
+                                    extent=extent, zorder=0)
+            artists.append(cim)
             self._axes.set_aspect(self.aspect, adjustable='datalim')
+
+        if hasattr(self._axes, 'colorbar'):
+            # Update existing colorbar.
+            self._axes.colorbar.update_normal(cim)
+        else:
+            # Add horizontal colorbar at the bottom.
+            self._axes.figure.subplots_adjust(bottom=0.15)
+            cbar_ax = self._axes.figure.add_axes([0.1, 0.05, 0.85, 0.05])
+            self._axes.colorbar = self._axes.figure.colorbar(
+                cim, cax=cbar_ax, orientation='horizontal')
+
         self.artists = artists
         self._sync_style()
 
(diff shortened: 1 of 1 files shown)


glue/viewers/image/layer_artist.py
```diff
@@ -178,12 +178,22 @@ def update(self, view, transpose=False, aspect=None):
...
-                                             interpolation='nearest',
-                                             origin='lower',
-                                             extent=extent, zorder=0))
+            cim = self._axes.imshow(image, cmap=self.cmap, norm=self.norm,
+                                    interpolation='nearest', origin='lower',
+                                    extent=extent, zorder=0)
+            artists.append(cim)
             self._axes.set_aspect(self.aspect, adjustable='datalim')
+
+        if hasattr(self._axes, 'colorbar'):
```
pllim: Ugly hack?

glue/viewers/image/layer_artist.py
```diff
@@ -178,12 +178,22 @@ def update(self, view, transpose=False, aspect=None):
...
+            # Update existing colorbar.
+            self._axes.colorbar.update_normal(cim)
+        else:
+            # Add horizontal colorbar at the bottom.
+            self._axes.figure.subplots_adjust(bottom=0.15)
```
pllim: Does not seem to have effect. Not sure what to do here.

astrofrog:
@pllim - thanks! Could you check whether you have the issue with adjust_subplots if you load an array without WCS information? (because if there is WCS information, it will use WCSAxes, otherwise it will use plain matplotlib, so it will allow us to check if the bug is with WCSAxes)

pllim:
@astrofrog , I see "World 0" and "World 1" axis labels when I open an image without any WCS. The colorbar is still blocking the X-axis. Does this mean Glue somehow still interpret the pixels as WCS? Or is this expected? Regardless, the lack of WCS does not fix the colorbar placement.

I'm pretty sure I didn't accidentally insert any WCS. Here is how I generated the image:

```python
from astropy.io import fits
import numpy as np
data = np.random.random((100, 100))
hdu = fits.PrimaryHDU(data)
hdu.writeto('no_wcs.fits')
```

```
$ fitsheader no_wcs.fits 
# HDU 0 in no_wcs.fits:
SIMPLE  =                    T / conforms to FITS standard                      
BITPIX  =                  -64 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                  100                                                  
NAXIS2  =                  100                                                  
EXTEND  =                    T             
```

PR #1093, opened 2016-08-23
Add option to disable maximizing window on startup

### LINKED ISSUE #1091: Do not want to start Glue in fullscreen mode
I see that a similar issue has been raised before (#469) but it was closed without resolution nor explanation.

pllim:
Commented out the line to maximize window. Fixes #1091.

### DIFF
--- a/glue/app/qt/application.py
+++ b/glue/app/qt/application.py
@@ -183,7 +183,7 @@ class GlueApplication(Application, QtWidgets.QMainWindow):
 
     """ The main GUI application for the Qt frontend"""
 
-    def __init__(self, data_collection=None, session=None):
+    def __init__(self, data_collection=None, session=None, maximized=True):
 
         self.app = get_qapp()
 
@@ -218,7 +218,7 @@ class GlueApplication(Application, QtWidgets.QMainWindow):
         lwidget.ui.layerTree.addAction(a)
         lwidget.bind_selection_to_edit_subset()
 
-        self._tweak_geometry()
+        self._tweak_geometry(maximized=maximized)
         self._create_actions()
         self._create_menu()
         self._connect()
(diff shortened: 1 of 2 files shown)


astrofrog:
Is there a way to make sure that the window does occupy the space available without being maximized though? On big screens, this might make a small window (and vice versa on small screens the window may be too big?)

pllim:
Ah, okay. The default on my desktop appears fine though. I'll dig around the code and see what happens. What kind of dimension do you have in mind? Would 80% of screen width and height be sufficient?

astrofrog:
Well ideally I still want it to fill the screen but without being 'fullscreen' if that makes sense? :-/

pllim:
p.s. Not very sure why the tests are failing... Segmentation fault?!

pllim:
@astrofrog , okay. It is now an option. Works when I tested it.

astrofrog:
Perfect, thanks!

Interestingly, the original code should just make the window as big as possible but not truly full screen (for that, we'd have had to use ``WindowFullScreen`` - which we don't want). In any case, I'm happy with this solution.

astrofrog:
@pllim - the segmentation faults are non-deterministic. I never managed to completely eradicate them, they are somehow caused I think by the tests that access the Qt IPython console.

pllim:
I have used `gdb` to trace seg fault in Python GUI before but that still requires a reproducible seg fault condition... Good luck with that...

pllim:
@astrofrog , sorry for mixing up "fullscreen" with "maximized screen". It is indeed not fullscreen, strictly speaking, but rather covering the whole screen except system toolbar and menubar. For my workflow that involves a gazillion of other terminal/GUI windows, that is still not desirable. But I'm glad this is merged, so I can choose to disabled "maximized" mode. Thanks!

pllim:
The keyword name `--no-fullscreen` might not be accurate since it is really supposed to be `--no-maximized-screen` but the latter is also very long to type. If you want me to submit a follow-up PR to change the keyword name, please let me know.

astrofrog:
@pllim - that might be better, just in case we introduce an actual full screen mode/option in future.

PR #1097, opened 2016-08-23
Wait cursor only shows when Glue is really doing something

### LINKED ISSUE #1092: "Please wait" cursor while trying to select input file makes me nervous
Currently, when I choose `File -> Open Data Set`, the cursor changes from "arrow" to "spinning circle" (usually used to indicate that something is busy). That is not the cursor that I expect to see when I am trying to choose a file. It makes me nervous, as I thought I might have accidentally opened a giant file that is eating up my memory.

pllim:
This fixes #1092 (the second commit).

### DIFF
--- a/glue/app/qt/application.py
+++ b/glue/app/qt/application.py
@@ -29,7 +29,7 @@ from glue.viewers.scatter.qt import ScatterWidget
 from glue.utils import nonpartial
 from glue.utils.qt import (pick_class, GlueTabBar,
                            QMessageBoxPatched as QMessageBox,
-                           set_cursor, messagebox_on_error, load_ui)
+                           set_cursor_cm, messagebox_on_error, load_ui)
 
 from glue.app.qt.feedback import submit_bug_report, submit_feedback
 from glue.app.qt.plugin_manager import QtPluginManager
@@ -675,7 +675,6 @@ class GlueApplication(Application, QtWidgets.QMainWindow):
 
     new_data_viewer = defer_draw(Application.new_data_viewer)
 
-    @set_cursor(Qt.WaitCursor)
     def _choose_save_session(self):
         """ Save the data collection and hub to file.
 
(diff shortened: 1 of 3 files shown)


pllim:
@astrofrog , I tried this with loading a big data file and the cursor appears to change properly. I didn't try saving/loading Glue session files. Perhaps you want to take this for a spin before merging.

glue/app/qt/application.py
```diff
@@ -733,11 +737,11 @@ def _reset_session(self, show=True):
...
-            dialog = QMessageBox.warning(self, "Confirm Close",
-                                         "Are you sure you want to reset the session? "
-                                         "This will close all datasets, subsets, and data viewers",
-                                         buttons=buttons,
-                                         defaultButton=QMessageBox.Cancel)
+            dialog = QMessageBox.warning(
+                self, "Confirm Close",
+                "Are you sure you want to reset the session? "
+                "This will close all datasets, subsets, and data viewers",
+                buttons=buttons, defaultButton=QMessageBox.Cancel)
```
astrofrog: In general, I prefer not to change the line length in cases like this when it actually makes the code less readable - would you mind disabling the line length PEP8 warnings or changing the line length to 100 or 120?

glue/app/qt/application.py
```diff
@@ -393,7 +393,8 @@ def _update_plot_dashboard(self, sub_window):
         if sub_window is None:
             return
 
-        layer_view, options_widget, title = self._get_plot_dashboards(sub_window)
+        layer_view, options_widget, title = self._get_plot_dashboards(
+            sub_window)
```
astrofrog: Here's another example where I don't think wrapping makes the code more readabale

astrofrog:
(merged manually since this required resolving conflicts)

glue/app/qt/application.py
pllim: Ok, no problem. I'll keep that in mind for the future.

pllim:
How did you "merge manually" and still get the PR to be listed as "merged" instead of "closed without merging"? I'm just curious. I have seen this across different repos. There must be some kind of git-fu that I have yet learned.

astrofrog:
@pllim - I think it's because I made the merge commit message look exactly like what GitHub normally does (I changed the default merge message)

pllim:
@astrofrog -- I added a note in my commit indicating I was not sure if a context manager should belong in a module named "decorators" (OCD speaking again!), but I couldn't find another module that is appropriate and didn't want to create a new `context_managers.py` with only this in it. FYI.

PR #1102, opened 2016-08-24
Move decorator to the proper module

pllim:
Move decorator from `main.py` to `decorators.py`.

Because I'm OCD like that.

### DIFF
--- a/glue/main.py
+++ b/glue/main.py
@@ -7,6 +7,7 @@ import optparse
 
 from glue import __version__
 from glue.logger import logger
+from glue.utils.qt import die_on_error
 
 
 def parse(argv):
@@ -88,37 +89,6 @@ def verify(parser, argv):
         err_msg = "Must provide a .glu file\n"
 
     return err_msg
-
-
-def die_on_error(msg):
-    """Decorator that catches errors, displays a popup message, and quits"""
-    def decorator(func):
-        def wrapper(*args, **kwargs):
-            try:
-                return func(*args, **kwargs)
-            except Exception as e:
-
-                import traceback
-
-                # Make sure application has been started
-                from glue.utils.qt import get_qapp
-                get_qapp()
-
-                from glue.utils.qt import QMessageBoxPatched as QMessageBox
-                m = "%s\n%s" % (msg, e)
-                detail = str(traceback.format_exc())
-                if len(m) > 500:
-                    detail = "Full message:\n\n%s\n\n%s" % (m, detail)
-                    m = m[:500] + '...'
-
-                qmb = QMessageBox(QMessageBox.Critical, "Error", m)
-                qmb.setDetailedText(detail)
-                qmb.show()
-                qmb.raise_()
-                qmb.exec_()
-                sys.exit(1)
-        return wrapper
-    return decorator
 
 
 @die_on_error("Error restoring Glue session")
--- a/glue/main.py
+++ b/glue/main.py
@@ -7,7 +7,7 @@ import optparse
 
 from glue import __version__
 from glue.logger import logger
-from glue.utils.qt import die_on_error
+from glue.utils.qt.decorators import die_on_error
 
 
 def parse(argv):
--- a/glue/utils/qt/decorators.py
+++ b/glue/utils/qt/decorators.py
@@ -1,12 +1,15 @@
 from __future__ import absolute_import, division, print_function
 
+import sys
 import traceback
 from contextlib import contextmanager
 from functools import wraps
 
+from glue.utils.qt import get_qapp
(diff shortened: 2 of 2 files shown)


astrofrog:
> Because I'm OCD like that.

Me too 😆  - I've been doing a lot of reorganization this year to tidy up things like this. Thanks!

astrofrog:
@pllim - just for info, the test failures looks real (not sure what's going on yet though, will look later this week unless you figure out the issue first!)

pllim:
Not very sure why Appveyor failed. I don't think my PR is that drastic...

pllim:
@astrofrog , I think it was caused by circular import. Hopefully I fixed it now.

pllim:
Hmm, now tests are timing out...

astrofrog:
@pllim - the reason the tests are timing out is because the following import:

```
from glue.utils.qt import QMessageBoxPatched as QMessageBox
```

needs to be inside the function. The reason for that is that the test patches QMessageBox, but that only works (with mock) if the function using that class accesses that class at the last minute. Currently, the class is imported into the global file namespace as soon as it's executed, and before it's patched. It *may* also be possible to update the patching to point to ``glue.utils.qt.decorators. QMessageBox`` but I'm not 100% sure. Note that there may also be patching in other tests, so you might want to search the whole code base for patching involving QMessageBox.

If you don't have time to look into this, I can also finish up the PR, so just let me know.

pllim:
@astrofrog , I moved the import inside the functions. Awaiting test results.

pllim:
Still getting errors... :cry: 

This one complains about a line that was added in Mar 2015, not sure why...
```python
===== ERRORS =====
______ ERROR collecting glue/plugins/ginga_viewer/qt/tests/test_client.py ______

Using @pytest.skip outside of a test (e.g. as a test function decorator) is not allowed. Use @pytest.mark.skip or @pytest.mark.skipif instead.

__ ERROR collecting glue/plugins/ginga_viewer/qt/tests/test_viewer_widget.py ___

Using @pytest.skip outside of a test (e.g. as a test function decorator) is not allowed. Use @pytest.mark.skip or @pytest.mark.skipif instead.
```
This one is directly tied to this PR but I am not sure what is the proper solution:
```
Traceback (most recent call last):

  File "/home/travis/miniconda/envs/test/bin/glue", line 11, in <module>

    load_entry_point('glueviz==0.9.0.dev219', 'gui_scripts', 'glue')()

  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/pkg_resources/__init__.py", line 565, in load_entry_point

    return get_distribution(dist).load_entry_point(group, name)

  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2589, in load_entry_point

    return ep.load()

  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2249, in load

    return self.resolve()

  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2255, in resolve

    module = __import__(self.module_name, fromlist=['__name__'], level=0)

  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/glueviz-0.9.0.dev219-py2.7.egg/glue/__init__.py", line 129, in <module>

    from .main import load_plugins

  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/glueviz-0.9.0.dev219-py2.7.egg/glue/main.py", line 10, in <module>

    from glue.utils.qt.decorators import die_on_error

ImportError: No module named qt.decorators
```

astrofrog:
@pllim - yes, the pytest version just got bumped in conda hence the errors. I have a fix that I'll include in a PR imminently.

astrofrog:
Ok, so the other failure is because one of the test configs is to make sure that the test suite runs normally if all ``qt`` sub-directories are removed (to enforce the separation of Qt code). I'll check what a reasonable solution would be in this case.

astrofrog:
I think the proper fix for this will take a little work - basically I think we should have a higher level ``@die_on_error`` that either uses the qt version or a non-qt version (basically just re-raising the exception in terminal) if qt is not present. Maybe the high level version could live in ``glue.utills`` (not ``glue.utils.qt``). If you don't have time to investigate this, I can take over this branch and continue in a separate PR (with your commits) - just let me know!

astrofrog:
@pllim - by the way, if you rebase it should get rid of the pytest issues

pllim:
@astrofrog, feel free to take over since you know how to best organize the top level decorator. Thanks!

astrofrog:
Continued in #1117

PR #1112, opened 2016-08-27
Implement Data.update_from_data

astrofrog:
This allows a dataset to be updated to match another dataset - importantly, it preserves component IDs for components that have matching labels. This can be used for example for refreshing a dataset where the number of elements has changed.


astrofrog:
@gahjelle - this is now merged, but note that I renamed the method to ``update_values_from_data``, so be sure to update your plugin.

PR #1101, opened 2016-08-24
Replace execfile() for Python 3 compatibillity

pllim:
First commit does PEP8 clean up on `main.py`. Second commit fixes #1100.

Note: I have no file to test the `exec()` with, so please run a test before you merge. Thanks.

PR #1114, opened 2016-08-27
Add regression test and changelog entry for #1101

astrofrog:
@pllim - just for info, you can actually execute any Python script (including 'hello, world!') with the ``glue -x script.py``, which dispatches to ``execute_script`` (which you fixed).


pllim:
Good to know. Thanks!

PR #1127, opened 2016-09-19
WIP: refactor built-in viewers

astrofrog:
This is a large chunk of work, and it might be necessary to split it up into several pull requests at a later time, but this will be the main PR to keep track of the work. This tries to go part of the way towards having a more stateful description of viewers (as described in https://github.com/glue-viz/glue/issues/775).

Things that are needed for each different viewer:

**Scatter viewer**
- [x] Make sure we use plot instead of scatter when there is no color coding, for performance
- [x] Choose better defaults for x and y axis
- [ ] Add labels in layer artist editor
- [x] Implement reset button
- [ ] Deal properly with categorical components
- [x] Fix log buttons
- [x] When layer is created for subset, copy settings from parent data layer
- [x] Limits should not change when showing a subset for fast scatter viewer

**Histogram viewer**
- [ ] Decouple view limits from histogram settings
- [ ] Don't update number of bins until user presses enter
- [ ] Add option to round selection to bins
- [ ] Add labels in layer artist editor
- [ ] Implement reset button
- [ ] Alpha doesn't have immediate effect
- [ ] Y limits flipping doesn't work
- [ ] Deal properly with categorical components

**Image viewer**
- [x] Image doesn't always appear
- [x] Limits aren't chosen automatically
- [x] Layer isn't visible by default
- [ ] Implement slicing
- [ ] Need to allow a dataset to be added multiple times to a viewer (since might want to show different components in each layer)
- [ ] Fix colors of image when using single-color mode (https://github.com/matplotlib/matplotlib/issues/7201)
- [ ] Have both a color-to-white and color-to-black mode

**Overall**
- [ ] Implement saving/restoring sessions
- [ ] Restore sessions from old files
- [ ] Improve performance by minimizing draws


astrofrog:
Superseded by https://github.com/glue-viz/glue/pull/1289

PR #1145, opened 2016-10-15
Refactor terminal code

astrofrog:
This requires IPython>=4 which I think will be reasonable for the next major version of glue


astrofrog:
Ipython or qtconsole requires setuptools >= 18, so need to make sure that gets installed.

PR #1153, opened 2016-10-28
Implement InequalitySubsetState for categorical components

astrofrog:
This makes it possible to create InequalitySubsetStates for categorical components using e.g. `d.id['a'] == 'string'`


astrofrog:
Marking this as a bug since this is likely how people intended things to work with the current implementation.

Issue #1152, opened 2016-10-25
Failed to restore session stored before Glue0.9


PennyQ:
So I had a session that was saved before glue9.0 , data is not included, but I got several errors when trying to open it under latest-dev version of Glue.

The first one is because we use `Subset` instead of `PatchedElementSubsetState`, so I also replaced them in my `.glu file`.

> Traceback (most recent call last):
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/utils/qt/decorators.py", line 53, in wrapper
>     return func(_args, *_kwargs)
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/app/qt/application.py", line 730, in _restore_session
>     ga = self.restore_session(file_name)
> ...
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/core/state.py", line 409, in _dispatch
>     typ = lookup_class_with_patches(rec['_type'])
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/core/state.py", line 120, in lookup_class_with_patches
>     return lookup_class(name)
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/utils/misc.py", line 81, in lookup_class
>     raise ValueError("Object '{0}' not found".format(ref))
> ValueError: Object 'glue_vispy_viewers.common.toolbar.PatchedElementSubsetState' not found

The second is as below, which I don't know how to solve it. @astrofrog Do you know the reason for it? thanks! All data is here https://1drv.ms/f/s!Avn4HRIa1XyWil0-SO9TKF2xGnM9

> Traceback (most recent call last):
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/utils/qt/decorators.py", line 53, in wrapper
>     return func(_args, *_kwargs)
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/app/qt/application.py", line 730, in _restore_session
>     ga = self.restore_session(file_name)
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/app/qt/application.py", line 772, in restore_session
>     ga = Application.restore_session(path)
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/core/application_base.py", line 128, in restore_session
>     return state.object('__main__')
>   File "/Users/penny/Works/Gluedev/glue/glue/glue/core/registry.py", line 91, in wrapper
>     return func(_args, *_kwargs)
[...]

PR #1160, opened 2016-11-01
Fix various issues with session loading

astrofrog:
@PennyQ - I think this should fix the issues you had loading the session in #1152 though I still need to add a fix so that ``PatchedElementSubsetState`` gets fixed too. Do you have the original session file before that change?


astrofrog:
The Travis failure is unrelated (intersphinx timeout)

PR #1171, opened 2016-11-16
Allow WCSCoordinates to init from only WCS

### LINKED ISSUE #1170: Allow WCSCoordinates to be initialized with only a WCS object
At the moment, WCSCoordinates **requires** a header, but we should change this so that a WCS object alone should be enough.

pllim:
Fixes #1170 and also some code clean-ups. @astrofrog , is this what you had in mind?

### DIFF
--- a/glue/core/coordinates.py
+++ b/glue/core/coordinates.py
@@ -85,6 +85,13 @@ class WCSCoordinates(Coordinates):
     standard.  This class does not take into account
     distortions.
 
+    Parameters
+    ----------
+    header
+        FITS header (derived from WCS if not given)
+    wcs
+        WCS object to use, if different from header
+
     References
     ----------
     * Greisen & Calabretta (2002), Astronomy and Astrophysics, 395, 1061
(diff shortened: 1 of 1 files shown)


pllim:
I see that this PR caused a test to fail on `WCSCoordinates(None)` but I don't know why one would want to make a nonsensical WCS.

astrofrog:
@pllim - the failing tests were actually checking an error was raised:

```python
    def test_pixel2world_invalid_input(self):
        coord = WCSCoordinates(None)
        x, y = {}, {}
        with pytest.raises(TypeError) as exc:
            coord.pixel2world(x, y)

    def test_world2pixel_invalid_input(self):
        coord = WCSCoordinates(None)
        x, y = {}, {}
        with pytest.raises(TypeError) as exc:
            coord.world2pixel(x, y)
```

but now the error is just raised earlier, at ``__init__`` time. Can you just update those tests to 

```python
        with pytest.raises(ValueError) as exc:
            coord = WCSCoordinates(None)
        assert exc.value.args[0] = '...'
```

? (and drop the rest of the content of the tests)

pllim:
I am re-submitting this as #1203 due to merge conflicts.

PR #1185, opened 2016-11-28
Significant performance improvements for WCS linking with cubes

astrofrog:
This includes a number of general performance improvements mainly related to avoiding repeating calculations in broadcasted arrays. The most immediate application is when linking e.g. a spectral cube and an image, selections in the image now propagate very fast to cubes.


astrofrog:
Travis is passing except for the Mac builds, which are delayed, but they were passing before, so I think it should be fine.

PR #1189, opened 2016-12-07
Improvements in linking

astrofrog:
### About

This re-thinks some aspects of data linking based on feedback from users. Some things that have confused users in the past are:

1. Due to the de-duplication of component IDs, if one links e.g. component ``A`` and ``C``, the link originally appeared as ``A <-> C``, but when opening the link editor later, the link was changed to ``A <-> A`` making it difficult to know what is what.

2. Following linking, some linked components appeared in the list of components of the link editor window. This would cause e.g. several instances of ``Pixel Axis 0 [z]`` to appear in the list, making it impossible to know which was which.

3. In the list of links, when linking e.g. 5 datasets all with components ``RA`` and ``Dec``, one would end up with many links of the form ``RA <-> RA`` with no indication of which data were linked. In principle, component IDs don't have to be attached to a dataset, but in 99% of use cases it will be.

This PR attempts to change this.

### Removing the merging of component IDs

First, this PR gets rid of the de-duplication of component IDs. This de-deduplication was discussed extensively in https://github.com/glue-viz/glue/pull/508 and there are some good reasons to do it, but unfortunately I think the cons outweigh the pros. Namely, the issue with de-duplication is issue 1 above (the original names of components is lost) and also that it's then impossible to un-glue all the links and return the data to the original state. Telling people that they've messed up the links and need to restart glue is not ideal.

### Introducing AutoDerivedComponent

A consequence of the removal of merging is that the link editor started showing all the linked components for each dataset. This PR introduces a trivial sub-class of ``DerivedComponent`` called ``AutoDerivedComponent`` which is used for derived components added via links (to differentiate it from user-created derived components). This then allows us to filter out the ``AutoDerivedComponents`` in lists of components, since these should never really be seen by the user. This addresses problem 2 above.

### Assigning parent data objects to component IDs

This may be a little controversial, but is needed to address problem 3. In 99% of use cases, component IDs are specific to a given dataset - that is, without the de-duplication, it's likely extremely rare that users set up data objects which share a component ID. In this PR, component IDs have a ``parent`` attribute that is the data they were first attached to. If a component ID is then added to a new dataset, the parent will still refer to the original dataset the component ID was added to. On the plus side, this then allows us to show this in the list of links:

screen shot 2016-12-07 at 19 53 24

Note that this uses rich text, so I've added a ``to_html`` method on ComponentIDs and on ComponentLinks. ComponentIDs without parent data are still allowed and can be used in ComponentLink with arithmetic expressions.

### Re-layout of the link editor

I've done some re-layout of the link editor to move the functions/gluing options in between the two columns of components and moved the list of links to the bottom (this ensures that even if the lines with the links are long, they are still readable). I've also made sure that horizontal scroll bars appear if needed in the list of links.

**Simple mode**

screen shot 2016-12-08 at 16 27 27

**Advanced mode**

screen shot 2016-12-08 at 16 27 38

@ChrisBeaumont @stscieisenhamer - it would be great if you could take a look at this given the discussion in https://github.com/glue-viz/glue/pull/508 - thanks!


glue/core/component.py
```diff
@@ -338,6 +338,10 @@ def __setgluestate__(cls, rec, context):
         return cls(None, rec['axis'], rec['world'])
 
 
+class AutoDerivedComponent(DerivedComponent):
```
ChrisBeaumont: small nit: I'm not sure if you want the name of this to be tied to its origin (added by machine) or its functional organization (should be invisible in certain contexts)

glue/core/component_link.py
```diff
@@ -166,16 +190,37 @@ def get_using(self):
...
+        return result
+
+    def to_html(self):
+        args = ", ".join([t.to_html() for t in self._from])
+        if self._using is identity:
+            result = "%s &#8596; %s" % (self._to.to_html(), self._from[0].to_html())
         else:
-            result = "%s <-> %s" % (self._to, self._from)
+            if self._inverse is None:
+                result = "%s &#8592; %s(%s)" % (self._to.to_html(), self._using.__name__, args)
```
ChrisBeaumont: small nit -- does unicode work here? Or are these HTML codes required?

astrofrog: I can't get quite the same result with unicode, so leaving this.

PR #1196, opened 2016-12-14
Properly deal with reassignment of Data.coords

astrofrog:
Re-compute world components if ``Data.coords`` is changed.

This will need to be rebased after https://github.com/glue-viz/glue/pull/1195 is merged and also will have some failures we need to investigate after https://github.com/glue-viz/glue/pull/1189 is merged.


astrofrog:
Still need to add a test to make sure all links based on the old world coordinates are removed

PR #1200, opened 2017-01-06
Travis: test with Python 3.6 by default

astrofrog:
Might not work yet since not all packages exist with Python 3.6


astrofrog:
🎉

PR #1203, opened 2017-01-10
Allow WCSCoordinates to init from only WCS

### LINKED ISSUE #1170: Allow WCSCoordinates to be initialized with only a WCS object

pllim:
Fix #1170. Supersedes #1171.

### DIFF
--- a/glue/core/coordinates.py
+++ b/glue/core/coordinates.py
@@ -202,9 +202,9 @@ class WCSCoordinates(Coordinates):
 
     Parameters
     ----------
-    header
+    header : :class:`astropy.io.fits.Header`
         FITS header (derived from WCS if not given)
-    wcs
+    wcs : :class:`astropy.wcs.WCS`
         WCS object to use, if different from header
 
     References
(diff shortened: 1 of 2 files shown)


glue/core/coordinates.py
```diff
@@ -199,18 +200,31 @@ class WCSCoordinates(Coordinates):
     standard.  This class does not take into account
     distortions.
 
+    Parameters
+    ----------
+    header
+        FITS header (derived from WCS if not given)
```
astrofrog: Can you add the type, i.e:

```
header : :class:`astropy.io.fits.Header`, optional
```

and same for ``wcs``?

glue/core/coordinates.py
```diff
@@ -358,10 +377,16 @@ def _get_ndim(header):
```
astrofrog: This should be an astropy WCS object, not just any object

pllim:
Thanks for reviewing, @astrofrog ! Your comments are addressed in the latest commit. And since I don't see any Sphinx test set up on Travis, I added `[skip ci]` to that commit (although it appears to run CircleCI anyway.

astrofrog:
@pllim - just so you know, there is a Sphinx text (it's the one that says ``DOC_TRIGGER=1`` - not very obvious!

pllim:
Ops. Just let me know if you want me to force push that last commit to re-trigger the entire CI suite.

astrofrog:
Thanks @pllim!

PR #1207, opened 2017-01-17
Added a decorator to avoid circular calling of class methods

astrofrog:
This is needed in PRs to refactor the viewers, but just thought I'd break it off as a separate PR.


astrofrog:
The test passed before the last skipped commit, so merging

PR #1216, opened 2017-02-03
Add properties to Subset that are the same as for Data (and other small fixes)

astrofrog:
This makes it so that e.g. Subset.visible_components works and just mirrors the Data object. This avoids things like:

```python
if isinstance(data, Subset):
    components = data.data.visible_components
else:
    components = data.visible_components
```

This also includes bug fixes for State-related functionality (since all these changes were made as part of the refactoring of the 3D viewers)


astrofrog:
The Travis failures are unrelated

PR #1217, opened 2017-02-06
Fix a bug which caused lower/upper values to be overwritten upon initialization

astrofrog:
This is just a bug in the developer version, so need for a changelog entry


astrofrog:
The Travis failures are unrelated, so I'll go ahead and merge this

PR #1231, opened 2017-02-09
Remove bundled version of WCSAxes and require Astropy >= 1.3

astrofrog:
WCSAxes is bundled with Astropy since 1.3


astrofrog:
The AppVeyor failure for Python 2.7 is unrelated, and will disappear once conda packages for Astropy and older Numpy versions are available.

PR #1246, opened 2017-02-18
Fixed RGB session loading

astrofrog:
Fixed loading of session files that have been produced after making an RGB visualization and switching back to monochromatic mode


astrofrog:
This bug is not present in the re-written viewers, so I'm going to close this.

PR #1248, opened 2017-02-22
Inherit .size attribute from Data in Subset

astrofrog:
We already inherit shape so this is just an additional convenience


astrofrog:
On second thoughts however this could be ambiguous because people might think Subset.size refers to the actual number of elements in the subset...

PR #1249, opened 2017-02-22
Improve naming of components during merging of datasets

astrofrog:
Prior to this, one of the components would have a name that didn't indicate which data it came from in ambiguous cases.


astrofrog:
The changelog entry needs to be moved

PR #1259, opened 2017-03-15
Remove unnecessary install dependency

rendinam:
Hello.  Would it be possible to remove the "install_requires" dependency on `glue-vispy-viewers`? It appears to not be a hard build, install, or runtime dependency for glue and the reference to `glue-vispy-viewers` here necessitates some gymnastics in packaging other glue plugins in astroconda, for instance.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,7 +115,6 @@ install_requires = ['numpy>=1.9',
                     'ipykernel',
                     'qtconsole',
                     'dill>=0.2',
-                    'glue-vispy-viewers>=0.7',
                     'xlrd>=1.0',
                     'h5py>=2.4']
 


astrofrog:
I agree that this makes packaging a little awkward - my main issue though is that I want this functionality to be available by default, but would like to avoid merging the packages. I wonder whether having some kind of conda meta-package would solve this - so for instance, the current main package would be packaged as **glue-core**, then **glue-vispy-viewers** would depend on **glue-core**, and finally **glueviz** would be a meta-package that depends on both **glue-core**, **glue-vispy-viewers**, and other glue plugins. What do you think?

rendinam:
After some examination, simpler option may exist.
Enforcing a tight coupling between **glueviz** and **glue-vispy-viewers** by having **glueviz** depend upon **glue-vispy-viewers** from within both `setup.py` and the conda recipe would pull in **glue-vispy-viewers** whenever **glueviz** is installed or built while allowing other plugins to continue to simply depend upon **glueviz**, ensuring that their host application remains a precondition. This arrangement would require the existing PR to be revoked and a new one filed against **glue-vispy-viewers** to remove the explicit **glueviz** dependency there.

After testing this, it appears to accommodate your desire to have **glue-vispy-viewers** functionality provided as default behavior while allowing for straightforward astroconda recipe maintenance.

If this sounds workable, I'll go ahead and pull the necessary levers.

astrofrog:
@rendinam I decided (for other reasons) to go down the road of making a glueviz metapackage. I'll open a PR to astroconda in the next hour.

PR #1277, opened 2017-04-06
WIP: Continued version of refactored viewers

astrofrog:
This is a rebased version of https://github.com/glue-viz/glue/pull/1127 (putting this in a separate branch as building it from scratch by cherry-picking and combining commits).

PR #1278, opened 2017-04-06
Refactor built-in histogram viewer

astrofrog:
In order to finish #1277 and #1127, I need to proceed in phases, so this PR is to refactor just the histogram viewer, since it is the one that needs the fewest conceptual changes compared to the built-in viewer.

Things to do for feature parity:

* [x] Select sensible number of bins for categorical components
* [x] Show categorical labels on x-axis

Other things to do:

* [x] Update test suite
* [x] Make sure that session loading is backward-compatible
* [x] Decide how to deal with number of bins as we change attributes
* [x] Fix ``test_update_subset_ignored_if_not_present``
* [x] Add changelog entry
* [ ] Updated documentation to describe programmatically modifying new viewers (will do in another PR once things are stabilized)

**Note** that while the diff indicates an increase in the number of lines overall, some of these are to add infrastructure common to all matplotlib viewers. If we count code lines (excluding tests and empty lines), the line count in ``glue/viewers/histogram`` goes from 737 to 372, i.e. almost half.


astrofrog:
Need changelog entry then this can be merged

PR #1289, opened 2017-05-05
Refactor scatter and image viewers

astrofrog:
So far this only refactors the scatter viewer but we need to also do the image viewer since it depends on the scatter layer artist.

Remaining to-dos before this is complete:

* [x] Support for colormaps
* [x] Different stretch functions
* [x] Selections
* [x] Option for pixel aspect ratio
* [x] Display of coordinate axes and coordinate axis limits
* [x] Slicing of cubes
* [x] Use WCSAxes for coordinates
* [x] Changelog entry
* [x] Overlay scatter plots on top of images
* [x] Fix standard viewer tests
* [x] Make it possible to still make RGB images with different attributes of a single dataset
* [x] Backward-compatibility for session files
* [x] Backward-compatibility for standalone image widget
* [x] Make sure the spectrum extraction and PV extraction work
* [x] Make sure x/y_att_world is changed if x/y_att is changed
* [x] Take out echo changes and open PR to echo repo
* [x] Tidy up TEMP and WIP commits

Things that could be done in a future PR:
* [x] Option to decouple color of layer from global color setting
* [ ] Port over any tests that existed before that aren't covered by standard tests (https://github.com/glue-viz/glue/issues/1322)
* [ ] Better default for limits and color/colormap
* [ ] Add backward-compatibility for previous viewer attributes (https://github.com/glue-viz/glue/issues/1318)
* [ ] Update example on programmatically starting glue and controlling viewers (https://github.com/glue-viz/glue/issues/1319)
* [ ] Write docstrings for viewer state properties so they appear correctly in docs (https://github.com/glue-viz/glue/issues/1320)
* [x] Fix aspect ratio when adding an image subset (it currently reverts back to 'equal' if it was on 'auto' before)
* [ ] Show colormap in layer color symbol (https://github.com/glue-viz/glue/issues/1321)


astrofrog:
The decrease in coverage is not ideal, but this is partially due to unused code that I'll remove in a cleanup soon.

Issue #738, opened 2015-09-02
Add new matplotlib colormaps


astrofrog:
Including viridis

astrofrog:
Done!

PR #1292, opened 2017-05-21
Generate names for mpl colormaps and register all of them

cmprince:
Addressing issue #738.

This adds *all* of the matplotlib colormaps to the registry. This may not have been the desired outcome! The colormap selection is now pretty crowded, but maybe that's OK? 

screenshot from 2017-05-21 01-23-22

Would it be better for the colormaps to be ordered in some way?

### DIFF
--- a/glue/config.py
+++ b/glue/config.py
@@ -386,7 +386,7 @@ class ColormapRegistry(Registry):
             # Change suffix of the reversed colormaps
             mpl_cmap_name = re.sub(r'_r$', ' (reversed)', mpl_cmap_name)
 
-            for abbr, colname in colmapper.iteritems():
+            for (abbr, colname) in colmapper.items():
                 # First check that the colname is not in the cmap name:
                 # Skips cmaps like "Oranges" and "Purples" so that
                 # the abbreviations "Or" and "Pu" do not get replaced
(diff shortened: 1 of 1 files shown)


astrofrog:
@cmprince thanks for the contribution! That is a lot of colormaps 😮

I think it might be nicer for users, rather than overloading them with all these choices by default, to have a menubar plugin (http://glueviz.org/en/stable/customizing_guide/customization.html#custom-menubar-tools) which pops up a window and allows users to select all the colormaps they want to add to the default registry. So that dialog would show all the colormaps from matplotlib for example, and users could select from checkboxes which ones they want to register (or not) in the colormaps registry. What do you think?

cmprince:
@astrofrog thanks for the feedback! I like the idea; the colormaps could also be clustered by theme (monochrome, diverging, perceptually neutral, etc.) this way. I'll take a look at the plugin and try to come up with something acceptable.

sumanthrao:
I'd love to contribute to the issue, shall I go ahead?

astrofrog:
@sumanthrao - I'm not sure if @cmprince has had a chance to work on this, so feel free to take a stab at it!

astrofrog:
I'm going to close this as it has gone stale but it would still be great to have a way of doing what I mentioned in https://github.com/glue-viz/glue/pull/1292#issuecomment-304827895 if anyone has time at some point.

Issue #1307, opened 2017-06-22
loading of unstructured numpy arrays


underchemist:
Hi,

I was just experimenting with glueviz and dumped some of my data into a .npy file only to find that glueviz had trouble loading it! When loading an numpy array saved to a .npy file I got the following traceback from loading the data in glueviz
```
Error message: 'NoneType' object is not iterable

Traceback (most recent call last):
  File "D:\Anaconda3\lib\site-packages\glue\dialogs\data_wizard\qt\data_wizard_dialog.py", line 38, in data_wizard
    result = gdd.load_data()
  File "D:\Anaconda3\lib\site-packages\glue\dialogs\data_wizard\qt\data_wizard_dialog.py", line 109, in load_data
    d = load_data(path, factory=fac.function)
  File "D:\Anaconda3\lib\site-packages\glue\core\data_factories\helpers.py", line 237, in load_data
    d = as_list(factory(path, **kwargs))
  File "D:\Anaconda3\lib\site-packages\glue\core\data_factories\helpers.py", line 330, in auto_data
    return fac(filename, *args, **kwargs)
  File "D:\Anaconda3\lib\site-packages\glue\core\data_factories\npy.py", line 39, in npy_reader
    for name in npy_data.dtype.names:
TypeError: 'NoneType' object is not iterable
```

I checked [glue/glue/core/data_factories/npy.py](https://github.com/glue-viz/glue/blob/master/glue/core/data_factories/npy.py) and saw the ValueError that mentions currently only structured numpy arrays are supported
 
``` python
if not hasattr(npy_data.dtype, 'names'):
        raise ValueError("Numpy save file loading currently only supports structured"
                         " arrays, e.g., with specified names.")
```

I didn't realize you could have unstructured or structured arrays in numpy so I looked into it and found a pretty basic workaround in my case that took my input of a 230x230x512 array of float32 values and copied it in a structured way
```python
import numpy as np
from np.random import rand
a = rand(230, 230, 512) # unstructured dummy data
b = np.array(a, dtype='float32, float32, float32') # structured copy of a
np.save('filesave1.npy', a)
np.save('filesave2.npy', b) # loadable by glueviz
```
However, I found all this not due to the ValueError I quoted above but because an unstructured multidimensional numpy array still has a names attribute, it is just Nonetype, which cannot be iterated on as it does further down in the loader function.
```python
for name in npy_data.dtype.names:
        comp = Component.autotyped(npy_data[name])
[...]

PR #1314, opened 2017-06-28
Unstructuredarrayfix

underchemist:
Upon finding I couldn't import numpy arrays automatically as seen here: #1307 
I set out to do as @astrofrog suggested and automatically convert an unstructured array to a structured array with a single field name of 'array'.

this includes test functions for unstructured arrays which are essentially copies of the already existing test_npy/npz_load functions

this is my first pull request so let me know what I'm missing, etc

thanks!

### DIFF
--- a/glue/core/data_factories/npy.py
+++ b/glue/core/data_factories/npy.py
@@ -4,7 +4,7 @@ from glue.core.data import Data, Component
 from glue.config import data_factory
 from glue.core.data_factories.helpers import has_extension
 
-__all__ = ['is_npy', 'npy_reader', 'is_npz', 'npz_reader']
+__all__ = ['is_npy', 'npy_reader', 'is_npz', 'npz_reader', 'from_unstructured_array']
 
 # TODO: implement support for regular arrays, e.g., not just structured arrays?
 
--- a/glue/core/data_factories/npy.py
+++ b/glue/core/data_factories/npy.py
@@ -31,9 +31,8 @@ def npy_reader(filename, format='auto', auto_merge=False, **kwargs):
     import numpy as np
     npy_data = np.load(filename)
 
-    if not hasattr(npy_data.dtype, 'names'):
-        raise ValueError("Numpy save file loading currently only supports structured"
-                         " arrays, e.g., with specified names.")
+    if isinstance(npy_data.dtype.names, type(None)):
+        npy_data = from_unstructured_array(npy_data)
 
     d = Data()
     for name in npy_data.dtype.names:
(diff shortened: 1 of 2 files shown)


glue/core/data_factories/npy.py
```diff
@@ -42,6 +41,22 @@ def npy_reader(filename, format='auto', auto_merge=False, **kwargs):
 
     return d
```
astrofrog: A simpler solution, rather than converting the plain array to an unstructured array, would be to do something more along the lines of:

```python
d = Data()
if npy_data.dtype.names is None:
    comp = Component.autotyped(npy_data)
    d.add_component(comp, label='array')
else:
    for name in npy_data.dtype.names:
         comp = Component.autotyped(npy_data[name])
         d.add_component(comp, label=name)
```

glue/core/data_factories/npy.py
```diff
@@ -31,9 +31,8 @@ def npy_reader(filename, format='auto', auto_merge=False, **kwargs):
     import numpy as np
     npy_data = np.load(filename)
 
-    if not hasattr(npy_data.dtype, 'names'):
-        raise ValueError("Numpy save file loading currently only supports structured"
-                         " arrays, e.g., with specified names.")
+    if isinstance(npy_data.dtype.names, type(None)):
```
astrofrog: This can be more easily written as ``npy_data.dtype.names is None``

glue/core/data_factories/tests/test_numpy.py
```diff
@@ -20,6 +20,31 @@ def test_npy_load(tmpdir):
...
 
+def test_unstruc_npy_load(tmpdir):
+    data = np.array([[152.2352, -21.513], [21.412, 35.1341]], dtype='f8')
+
+    with open(tmpdir.join('test.npy').strpath, 'wb') as f:
+        np.save(f, data)
+        f.seek(0)
+
+        data2 = df.load_data(f.name)
+        assert_array_equal(data, data2['array'])
```
astrofrog: You shouldn't need the ``seek(0)`` - instead, move the ``data2=`` and the following line out of the ``with`` statement (up one indent level) and this will ensure the file is closed. I would also set:

```python
filename = tmpdir.join('test.npy').strpath
```

before the ``with`` statement and use that in the ``with`` statement and for ``load_data``

glue/core/data_factories/npy.py
```diff
@@ -74,9 +89,8 @@ def npz_reader(filename, format='auto', auto_merge=False, **kwargs):
         d = Data(label=groupname)
         arr = npy_data[groupname]
 
-        if not hasattr(arr.dtype, 'names'):
-            raise ValueError("Numpy save file loading currently only supports structured"
-                             " arrays, e.g., with specified names.")
+        if isinstance(arr.dtype.names, type(None)):
+            arr = from_unstructured_array(arr)
```
astrofrog: Same comment as for the ``npy`` reader, just add the single component manually and avoid making a recarray.

glue/core/data_factories/tests/test_numpy.py
```diff
@@ -20,6 +20,31 @@ def test_npy_load(tmpdir):
...
+    with open(tmpdir.join('test.npz').strpath, 'wb') as f:
+        np.savez(f, data1=data1, data2=data2)
+        f.seek(0)
+
+        data_loaded = df.load_data(f.name)
+        arr = data_loaded[0]
+        assert_array_equal(data1, arr['array'])
+
+        arr = data_loaded[1]
+        assert_array_equal(data2, arr['array'])
```
astrofrog: As for the test above, store the filename in a separate variable, remove seek, and move the ``load_data`` onwards out of the ``with statement``.

astrofrog:
I finished this up in #1385 - thanks for starting this @underchemist!

PR #1331, opened 2017-07-05
With PyQt5, make sure we hide Qt warnings due to qtconsole

astrofrog:
This should get rid of these warnings:

```
QTextCursor::setPosition: Position '-1' out of range
QTextCursor::setPosition: Position '10' out of range
```

See https://github.com/jupyter/qtconsole/pull/224 for a possible fix in qtconsole

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,8 @@ v0.11.0 (unreleased)
 v0.10.5 (unreleased)
 --------------------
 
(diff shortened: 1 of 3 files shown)


astrofrog:
This causes segmentation faults when the tests end - I could work around that but I think maybe best to just wait for qtconsole to get fixed...

PR #1366, opened 2017-08-05
Fix spectrum and PV slicer tools and improve overall UI

astrofrog:
This also internally (for now) allows aggregation in the image viewer (used by the spectrum tool). Will need to expose that in future as an alternative to slicing.

Also fixes issues with the GlueLogger and Ipython terminal when resetting or opening sessions from glue.

Finally improves the overall UI, moving buttons to the toolbar and making it clearer what they do.


astrofrog:
The failures are now unrelated

PR #1405, opened 2017-09-15
Add ability for clients to set tab name

drdavella:
This adds a method that allows callers to set a custom tab name. It also provides a method that allows users to determine the index of a given tab.


astrofrog:
Thanks!

PR #1430, opened 2017-10-17
add_datasets priorizes Data with WCS when merging datasets

robelgeda:
WCS information is lost when the user selects to merge a dataset using the UI. This is because `data_collection.merge` uses the coordinates of the first `Data` object as the coordinates of the newly merged `Data` object. This fix is applied to `glue.app.application._choose_merge` such that the first `Data` object in the dataset has WCS information (if available). If multiple `Data` objects have WCS info, the first one in the list is selected.


astrofrog:
@robelgeda - thanks! I think it would make more sense to put this in DataCollection.merge (in case this is used not from a UI), specifically at this line:

```
        master.coords = data[0].coords
```

astrofrog:
@robelgeda - please also add an entry in the changelog, thanks! (v0.12)

robelgeda:
I left the edits in `application._choose_merge` because that is where the label of the merged data is auto generated. If the user does not provide a label, we would like the Data with the WCS inforamtion to be the default data label.

PR #1511, opened 2018-02-01
Customize plot appearance and export scripts

astrofrog:
This adds two main features:

* The ability to set axis labels, and label sizes via the GUI

* The ability to export scripts from the Matplotlib viewers

This is still a work in progress, and only the Scatter viewer can export a script at the moment.


astrofrog:
Things that would be nice to customize: whether to show the top and right spines, and also the tick spacing

PR #1514, opened 2018-02-06
Keyshortcuts

javerbukh:
There was a bug earlier that did not allow None as a valid viewer. None basically means that the shortcut is not for a specific viewer, but for the global application. Please push this asap as this is a very important fix for functionality!

### DIFF
--- a/glue/config.py
+++ b/glue/config.py
@@ -691,14 +691,23 @@ class KeyboardShortcut(DictRegistry):
         arg3: function()
             function to be run that corresponds with key
         """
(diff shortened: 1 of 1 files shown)


astrofrog:
@javerbukh - it looks like this needs to be rebased on the latest upstream master - also, it looks like you included some files that shouldn't be here (the ``bin/...`` files). Thanks!

astrofrog:
@javerbukh - this is now merged, thanks!

PR #1515, opened 2018-02-08
Enable click-and-drag relocation of ROIs

drdavella:
This PR allows ROIs in image viewers to be selected using the mouse, which activates the corresponding subset. ROIs can be relocated by clicking and dragging them to new locations. ~At the moment, in order for this to work properly, all ROI selection toolbar buttons must be deselected before clicking and dragging an ROI.~

Things that remain to be done:
- [x] Refactor the `RoiSelectionMixin` into a mouse mode that is always active
- [x] Disable the ROI selection mouse mode immediately after a region has been selected (???)
- [x] Provide an optional toolbar button for ROI selection that creates a new subset every time it is activated, rather than just operating on the current subset
- [x] Allow simple ROIs/subsets to be deleted with a right-click context menu (???)


drdavella:
@astrofrog this is probably about ready for a first-pass review. There is some API breakage in here which I didn't consider at the time (namely the move of most toolbar modes from `glue/viewers/common/qt/mouse_mode.py` to `glue/viewers/common/qt/toolbar_mode.py`).

astrofrog:
@drdavella - you will only see the following error:

```
E               ValueError: No net viewers should be created in tests
```

if you have objgraph installed. I just pushed a commit that makes the debugging a little easier - here's the object graph I get for some of the image tests:

objgraph-0mf2_vif

Basically the issue is there is a circular reference that goes viewer -> _default_mouse_mode -> RoiClickAndDrag -> viewer (via the ``_viewer`` attribute).

glue/core/qt/roi.py
```diff
@@ -123,8 +123,8 @@ def __init__(self, axes):
 
 class QtPolygonalROI(QtROI, roi.MplPolygonalROI):
 
-    def __init__(self, axes):
-        roi.MplPolygonalROI.__init__(self, axes)
+    def __init__(self, axes, _roi=None):
+        roi.MplPolygonalROI.__init__(self, axes, roi=_roi)
```
astrofrog: Why the ``_`` in the keyword argument name?

glue/viewers/common/qt/data_viewer.py
```diff
@@ -53,6 +53,8 @@ class DataViewer(ViewerBase, QtWidgets.QMainWindow):
     LABEL = 'Override this'
 
     _toolbar_cls = None
+    # This defines the mouse mode to be used when no toolbar modes are active
+    _default_mouse_mode = None
```
astrofrog: If this is intended to be a class, add a ``_cls`` suffix (and same inside the Toolbar class in places where it should be a class not an instance)

glue/viewers/common/qt/toolbar_mode.py
```diff
@@ -0,0 +1,565 @@
+"""
+ToolbarModes are CheckableTools that enable various MouseModes.
+
```
astrofrog: It might be a bit too late for this, but if this is basically a modified ``mouse_mode.py`` it would have been nice to have done a ``git mv`` to preserve the history across renaming. Is there a way you can edit the history to do this?

glue/core/qt/roi.py
drdavella: It's because `roi` was already being used as a module name (`from glue.core import roi`, also see the line immediately below). It might make more sense to change the import, though,  to be able to use `roi` as the keyword argument?

astrofrog:
Thanks! There are some small things I'd like to improve but I figured I can just open a PR :)

PR #1522, opened 2018-02-14
Support for 3d subset and a lasso 3d selection

maartenbreddels:
Not so sure about what to do with the to_mask for image data. But if it looks good I can take a look at adding unittests and that may clarify things.

### DIFF
--- a/glue/core/roi.py
+++ b/glue/core/roi.py
@@ -77,14 +77,27 @@ class Roi(object):  # pragma: no cover
     def contains3d(self, x, y, z):
         """Return true/false for each x/y/z pair.
 
-        :param x: Array of X locations
-        :param y: Array of Y locations
-        :param y: Array of Z locations
+        Parameters
+        ----------
 
-        :returns: A Boolean array, where each element is True
-                  if the corresponding (x,y,z) tuple is inside the Roi.
 
-        :raises: UndefinedROI exception if not defined
+        x : ndarray
+            Array of x locations
+        y : ndarray
+            Array of y locations
+        z : ndarray
+            Array of z locations
+
+        Returns
+        -------
+        ndarray
+            A Boolean array, where each element is True if the corresponding
+            (x,y,z) tuple is inside the Roi.
+
+        Raises
+        ------
+        UndefinedROI
+            if not defined
         """
         raise NotImplementedError()
 
--- a/glue/core/roi.py
+++ b/glue/core/roi.py
@@ -74,6 +74,20 @@ class Roi(object):  # pragma: no cover
         """
         raise NotImplementedError()
 
+    def contains3d(self, x, y, z):
+        """Return true/false for each x/y/z pair.
+
+        :param x: Array of X locations
+        :param y: Array of Y locations
+        :param y: Array of Z locations
+
+        :returns: A Boolean array, where each element is True
+                  if the corresponding (x,y,z) tuple is inside the Roi.
+
+        :raises: UndefinedROI exception if not defined
+        """
+        raise NotImplementedError()
+
     def center(self):
         """Return the (x,y) coordinates of the ROI center"""
         raise NotImplementedError()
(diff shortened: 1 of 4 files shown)


glue/core/roi.py
```diff
@@ -561,6 +575,44 @@ def move_to(self, xdelta, ydelta):
         self.vy = list(map(lambda y: y + ydelta, self.vy))
 
 
+class PolygonalProjected3dROI(PolygonalROI):
```
astrofrog: Could you add projected 3D ROIs for rectangle and circle too in a similar way?

glue/core/roi.py
```diff
@@ -74,6 +74,20 @@ def contains(self, x, y):
...
+        """Return true/false for each x/y/z pair.
+
+        :param x: Array of X locations
+        :param y: Array of Y locations
+        :param y: Array of Z locations
+
+        :returns: A Boolean array, where each element is True
+                  if the corresponding (x,y,z) tuple is inside the Roi.
+
+        :raises: UndefinedROI exception if not defined
```
astrofrog: Could you use Numpydoc format here? I'm trying to transition the docstrings to numpydoc progressively, so it would be good to avoid adding new code that's not in Numpydoc.

glue/core/roi.py
```diff
@@ -561,6 +575,44 @@ def move_to(self, xdelta, ydelta):
...
+        if not isinstance(y, np.ndarray):
+            y = np.asarray(y)
+        # work in homogeneous coordinates so we can support perspective
+        # projections as well
+        vertices = np.array([x, y, z, np.ones(len(x))])
+        # homogeneous screen coordinates
+        screen_h = np.array(np.dot(self.projection_matrix, vertices))
+        # convert to screen coordinates, and we don't care about z
+        x, y = screen_h[:2] / screen_h[3]
+        return self.contains(x, y)
```
astrofrog: The code here doesn't work if x, y, and z are not 1D - it would be nice if we allowed x, y and z to be n-dimensional arrays and returned an n-dimensional mark at the end (I guess the easiest might be to unravel)

glue/core/roi.py
```diff
@@ -561,6 +575,44 @@ def move_to(self, xdelta, ydelta):
...
+            raise UndefinedROI
+        if not isinstance(x, np.ndarray):
+            x = np.asarray(x)
+        if not isinstance(y, np.ndarray):
+            y = np.asarray(y)
+        # work in homogeneous coordinates so we can support perspective
+        # projections as well
+        vertices = np.array([x, y, z, np.ones(len(x))])
+        # homogeneous screen coordinates
+        screen_h = np.array(np.dot(self.projection_matrix, vertices))
```
astrofrog: I think np.matmul is now the recommended matrix multiplication function

glue/core/subset.py
```diff
@@ -1091,6 +1091,81 @@ def __str__(self):
...
+    @contract(data='isinstance(Data)', view='array_view')
+    def to_mask(self, data, view=None):
+
+        # TODO: make sure that pixel components don't actually take up much
+        #       memory and are just views
+        x = data[self.xatt, view]
+        y = data[self.yatt, view]
+        z = data[self.zatt, view]
+
+        # if (x.ndim == data.ndim and
```
astrofrog: I think you can basically ignore/remove this - it was an optimization that was useful when selecting a 2D region in an n-d cube in the image viewer, but since the 3D viewers don't (yet) support showing n-dimensional datasets, this isn't useful. The only thing you need to properly support cubes is to make ``contains3d`` work with n-dimensional arrays.

glue/core/roi.py
```diff
@@ -561,6 +575,44 @@ def move_to(self, xdelta, ydelta):
...
+    def contains3d(self, x, y, z):
+        if not self.defined():
+            raise UndefinedROI
+        if not isinstance(x, np.ndarray):
+            x = np.asarray(x)
+        if not isinstance(y, np.ndarray):
+            y = np.asarray(y)
+        # work in homogeneous coordinates so we can support perspective
+        # projections as well
+        vertices = np.array([x, y, z, np.ones(len(x))])
```
astrofrog: ``(len(x))`` -> ``x.shape``

astrofrog:
I forgot to mention, it would be good to add some simple tests of this, at least to make sure the code runs.

maartenbreddels:
Yes, I didn't want to do unittests before you saw it, but I'll get to it next week.

In vaex this would be multithreaded and sequential (good for memory mapping). With dask the sequential isn't an option I think. 
Maybe it is doable with a decorator to do this in chunks using pure Python.

astrofrog:
> In vaex this would be multithreaded and sequential (good for memory mapping)

Would it be worth splitting out the multi-threaded selection code into a lightweight standalone library that isn't tied to vaex? What I'm thinking of is a library where the main API is simply a set of ``contains`` functions corresponding to the different type of ROI selections in glue. If Vaex already implements something like this then it would be nice to re-use it!

glue/core/roi.py
maartenbreddels: That also didn't do it, but tensordot seemed more appropriate without needing to do transposes.

glue/core/roi.py
```diff
@@ -74,6 +74,33 @@ def contains(self, x, y):
...
         raise NotImplementedError()
 
+    def contains3d(self, x, y, z):
+        """Return true/false for each x/y/z pair.
+
+        Parameters
+        ----------
+
+
+        x : ndarray
```
astrofrog: Can you get rid of the empty lines on line 82 and 83?

astrofrog: Can you replace `ndarray` by:

```
:class:`numpy.ndarray`
```

Some older docstrings may just have `ndarray` but this doesn't create an intersphinx link

glue/core/roi.py
```diff
@@ -74,6 +74,33 @@ def contains(self, x, y):
...
+            Array of x locations
+        y : ndarray
+            Array of y locations
+        z : ndarray
+            Array of z locations
+
+        Returns
+        -------
+        ndarray
+            A Boolean array, where each element is True if the corresponding
```
astrofrog: Add back-ticks around ``True``

glue/core/roi.py
```diff
@@ -561,6 +588,70 @@ def move_to(self, xdelta, ydelta):
         self.vy = list(map(lambda y: y + ydelta, self.vy))
 
 
+def _project(projection_matrix, x, y, z):
+    """Projects 3d coordinates to 2d coordinates using a 4x4 matrix"""
+    if not isinstance(x, np.ndarray):
+        x = np.asarray(x)
```
astrofrog: I think you can get rid of `if not isinstance(x, np.ndarray):` as I think `np.asarray` doesn't do a copy if the input is already an array?

glue/core/roi.py
```diff
@@ -561,6 +588,70 @@ def move_to(self, xdelta, ydelta):
...
+    # homogeneous screen coordinates
+    screen_h = np.tensordot(projection_matrix, vertices, axes=(1, 0))
+    # convert to screen coordinates, and we don't care about z
+    x, y = screen_h[:2] / screen_h[3]
+    return x, y
+
+class Projected3dROI(Roi):
+    """"A region of interest defined in screen coordinates.
+
+    The screen coordiantes are defined by the projection matrix.
```
astrofrog: typo: coordinates

maartenbreddels:
Ok, thanks! I guess you squash this all, or do you want me to squash it in >1 commit?

astrofrog:
@maartenbreddels - no squashing required :)

astrofrog:
The Travis failures are unrelated - thanks!

PR #1524, opened 2018-02-14
Re-implement keyboard shortcuts using keyPressEvent instead of using an eventListener

javerbukh:
Now uses keyPressEvent instead of eventLister to handle key presses. Also creates a keyPressEvent method in widget.py in order to make sure matplotlib does not exclusively take all of the keyPressEvents.

### DIFF
--- a/glue/app/qt/application.py
+++ b/glue/app/qt/application.py
@@ -243,9 +243,6 @@ class GlueApplication(Application, QtWidgets.QMainWindow):
         self.new_tab()
         self._update_plot_dashboard(None)
 
-    def start_event_filter(self):
-        self.app.installEventFilter(self)
-
     def run_startup_action(self, name):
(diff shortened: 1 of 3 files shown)


astrofrog:
@javerbukh - it looks like this includes a number of unrelated commits - could you rebase against the latest upstream master?

javerbukh:
Done! Sorry about that

astrofrog:
@javerbukh - it looks like you accidentally committed a file/directory called ``glue/glue``. Do you have a clone of the glue repository inside glue itself? I'll fix this up, but in future I would recommend adding files manually to your commits using ``git add`` and avoiding e.g. ``git commit -a`` or ``git add *``.

astrofrog:
@javerbukh - yes, that might be cleanest, thanks!

PR #1525, opened 2018-02-15
Added a NewMode for the EditSubsetMode

astrofrog:
@drdavella - this adds a new mode in the selection toolbar and in EditSubsetMode which controls whether a new selection is create each time (I prefer to have this than a per-mode setting otherwise things will get confusing). In CubeViz we could then set EditSubsetMode.mode = NewMode at the start.

My plan is then (in a follow-up PR) to add a way to say that modes should always change back to the default after each selection, or instead to always 'stick' to what the user selected.


drdavella:
Sounds good to me. Just curious but does this preclude the possibilty of a "context-sensitve" toolbar mode? For example, when an ROI is selected, activating an ROI mode would edit the corresponding subset, but when no ROI is selected, the mode would create a new subset.

astrofrog:
> Just curious but does this preclude the possibilty of a "context-sensitve" toolbar mode? For example, when an ROI is selected, activating an ROI mode would edit the corresponding subset, but when no ROI is selected, the mode would create a new subset.

The default ReplaceMode is actually a mode which does exactly that (modify if selected, new if none is selected). To get this to work, I think you just need to make sure that if the user clicks outside all the ROIs, that this deselects any subset in the top left data collection widget.

drdavella:
Okay, great. Yes I had planned to do this in another PR (unless you were already working on that). In addition, as a helpful cue to users, I think it would be useful to make the yellow overlay region persist for as long as a region/subset is selected.

PR #1551, opened 2018-03-01
Fix error message that occurs when closing glue

drdavella:
Specifically, this fixes an error message that occurs when closing cubeviz applications.


astrofrog:
Thanks!

PR #1553, opened 2018-03-01
Fix click and drag of ROI when using NewMode

drdavella:
Prior to this fix, clicking and dragging an existing ROI in `NewMode` would create a new ROI instead of relocating the existing one.


drdavella:
@astrofrog, no rush but I just wanted to point out that this PR should be treated as completely separate from #1556. This one fixes an existing bug, whereas the other is a prototype for new functionality.

astrofrog:
Thanks!

PR #1554, opened 2018-03-01
Improve performance of slicing

astrofrog:
For now this PR doesn't do much but I plan to investigate whether we can use ``redraw_frame`` to avoid calling WCSAxes every time a slice is changed.


astrofrog:
Actually I found another bottleneck so things are already a lot faster with multiple subsets. I need to add a regression test.

PR #1556, opened 2018-03-02
WIP: Provide persistent ROI selection

drdavella:
Previously when clicking on an existing ROI, the yellow selection overlay was only present until the mouse button was released. Now, after clicking on an ROI, the selection overlay is present until clicking on another ROI, or until clicking on an area without any ROI. Clicking on an area without any ROI will also cause there to be no active subset.

This has some interesting consequences for creating and resizing ROIs. If an existing ROI is currently selected, then activating one of the ROI mouse modes will edit the corresponding subset. If no ROI is selected, then activating one of the mouse modes will cause a new subset to be created. This all assumes that `NewMode` is not the active subset mode.'

@astrofrog we should discuss whether this seems like reasonable behavior, and if so, what the interaction should be with `NewMode`. If we want to move forward with this, there are a few more things that need to be accounted for:

- [ ] It should only be possible to have a single active ROI selection across *all* active viewers
- [ ] In some cases it makes sense to display the active selection between different mouse modes. For example, if I have an active selection, and then I activate circle mode, the ROI selection should remain until I click to create a new selection (this is not currently the case).
- [ ] If we adopt this, the tool tip for mouse modes should be context-sensitive. Display one message when an ROI is selected, and another when no ROI is selected (thanks @brechmos-stsci)

This PR is based on top of #1553.

### DIFF
--- a/glue/viewers/image/qt/data_viewer.py
+++ b/glue/viewers/image/qt/data_viewer.py
@@ -68,6 +68,8 @@ class ImageViewer(MatplotlibDataViewer):
                                             origin='lower', interpolation='nearest')
         self._set_wcs()
 
+        self._active_roi = None
+
     @defer_draw
     def update_x_ticklabel(self, *event):
         # We need to overload this here for WCSAxes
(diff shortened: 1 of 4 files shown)


drdavella:
Closing. Will open a separate PR later.

PR #1603, opened 2018-03-14
When no ROI is selected, make sure no subset is selected

drdavella:
When clicking outside an existing ROI, this now means there is no active subset. This will be useful for future ROI improvements in glue, and also for functionality in cubeviz.


astrofrog:
@drdavella - thanks!

PR #1617, opened 2018-03-20
Do projection for 3D ROI in chunks

astrofrog:
This ports over some performance optimizations that were present in the 3D viewers - basically the projection step can take a lot of memory so it's better to do it in chunks to avoid using too much memory in one go.

@maartenbreddels - let me know if you have any ideas on doing this better!

### DIFF
--- a/glue/core/roi.py
+++ b/glue/core/roi.py
@@ -586,20 +586,6 @@ class PolygonalROI(VertexROIBase):
         self.vy = list(map(lambda y: y + ydelta, self.vy))
 
 
-def _project(projection_matrix, x, y, z):
-    """Projects 3d coordinates to 2d coordinates using a 4x4 matrix"""
-    x = np.asarray(x)
-    y = np.asarray(y)
-    z = np.asarray(z)
-    # work in homogeneous coordinates so we can support perspective
-    # projections as well
-    vertices = np.array([x, y, z, np.ones(x.shape)])
-    # homogeneous screen coordinates
-    screen_h = np.tensordot(projection_matrix, vertices, axes=(1, 0))
-    # convert to screen coordinates, and we don't care about z
-    x, y = screen_h[:2] / screen_h[3]
-    return x, y
-
 class Projected3dROI(Roi):
     """"A region of interest defined in screen coordinates.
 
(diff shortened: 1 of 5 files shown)


glue/core/roi.py
```diff
@@ -614,11 +600,57 @@ def __init__(self, roi_2d=None, projection_matrix=None):
...
+        # FIXME: this can cause a copy to be made if the data is not
+        # contiguous (for example for broadcasted pixel coordinates).
+        x = x.ravel()
+        y = y.ravel()
+        z = z.ravel()
+
+        chunk_size = 1000000
+        array_size = len(x)
+
+        n_chunks = max(array_size // chunk_size, 1)
```
maartenbreddels: I think this will fail for an array_size of 2000001, since that should be 3 chunks. `int(math.ceil(array_size / chunk_size))` should work (without the max).

glue/utils/tests/test_array.py
```diff
@@ -114,3 +116,35 @@ def test_unbroadcast():
...
+    assert_equal(array, 1)
+
+
+def test_iterate_chunks_invalid():
+
+    array = np.zeros((6, 3, 4, 5, 1, 8))
+
+    with pytest.raises(ValueError) as exc:
+        next(iterate_chunks(array.shape, chunk_shape=(6, 2, 1)))
+    assert exc.value.args[0] == 'chunk_shape should have the same length as shape'
```
maartenbreddels: See https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions
version 3 has the message argument

glue/core/roi.py
astrofrog: Thanks! I've now refactored the chunking to use dedicated functions which operate a little differently (basically I am trying to avoid having to change the arrays to 1D)

glue/utils/array.py
```diff
@@ -175,3 +176,78 @@ def broadcast_to(array, shape):
...
+    """
+    Given a data shape and a chunk shape (or maximum chunk size), iteratively
+    return slice objects that can be used to slice the array.
+    """
+
+    if chunk_shape is None and n_max is None:
+        raise ValueError('Either chunk_shape or n_max should be specified')
+    elif chunk_shape is not None and n_max is not None:
+        raise ValueError('Either chunk_shape or n_max should be specified (not both)')
+    elif chunk_shape is None:
```
maartenbreddels: `if chunk_shape is None !=  n_max is None:` if you want to be short

astrofrog:
> You wonder if something like this shouldn't go upstream, say numpy (if it doesn't already have this)

There is ``numpy.nditer`` but I never got it to behave exactly as I wanted.

glue/utils/tests/test_array.py
astrofrog: Good to know! I might just change it everywhere in a separate PR, will open an issue.

glue/utils/array.py
astrofrog: I think I'll leave it as-is for clarity, I'd rather be more verbose in this case.

maartenbreddels:
Sometimes after writing it yourself you start to understand it better :)

PR #1619, opened 2018-03-21
Add a mouse over pixel selection tool

Cadair:
This dosen't seem to work, i.e. the button is not there, I can't figure out how to make it register with the image viewer.

### DIFF
--- a/glue/viewers/image/qt/data_viewer.py
+++ b/glue/viewers/image/qt/data_viewer.py
@@ -26,6 +26,7 @@ from glue.viewers.image.composite_array import CompositeArray
 
 # Import the mouse mode to make sure it gets registered
 from glue.viewers.image.contrast_mouse_mode import ContrastBiasMode  # noqa
+from glue.viewers.image.pixel_selection_mode import PixelSelectionTool  # noqa
 
 __all__ = ['ImageViewer']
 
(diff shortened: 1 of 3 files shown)


astrofrog:
@Cadair - add it to the list of tools on line 64 in the ``image/data_viewer.py`` file

astrofrog:
@Cadair - thanks! Can you add an entry to the changelog under v0.13.0? (order doesn't matter but maybe near the bottom to avoid conflict with another PR which adds an entry near the top)

astrofrog:
Currently this doesn't require the user to click at the start, merely hovering over the image starts the selection. I would prefer a click to be required to start the selection, so maybe _on_move should default to False?

glue/viewers/image/qt/data_viewer.py
```diff
@@ -63,7 +64,7 @@ class ImageViewer(MatplotlibDataViewer):
     tools = ['select:rectangle', 'select:xrange',
              'select:yrange', 'select:circle',
              'select:polygon', 'image:contrast_bias',
-             'save:python']
+             'image:point_selection', 'save:python']
```
astrofrog: Can you put it just *before* contrast_bias?

Cadair: sure! How are you ordering them? they are alphabetical now right?

Cadair:
Thanks!

PR #1628, opened 2018-03-22
Use memory-mapping with HDF5 files whenever possible

astrofrog:
This seems to work, though I haven't had a chance to actually try it out with any large files as my disk is almost full...

I also noticed that the auto-merging isn't quite right at the moment, so I'll try and work on that soon, maybe in a separate PR.

@maartenbreddels - when you get a chance, can you review this?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,9 @@ v0.13.0 (unreleased)
 * Improve hiding/showing of side-panels. No longer hide side-panels
   when glue application goes out of focus. [#1535]
 
+* Use memory-mapping with contiguous arrays in HDF5 files, resulting
+  in improved performance for large files. [#1628]
+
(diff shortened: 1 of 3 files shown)


maartenbreddels:
Yes, I'll take a look next week!

maartenbreddels:
LGTM.
I wouldn't worry about the array size, for small files the performance isn't an issue anyway. The only issue is actually OSX, their mmap support is really bad, as is their diskcaching strategy. Maybe we should have an config option, or keyword argument for when you don't want mmapping (for benchmarking, or when an issue arrises)

astrofrog:
@maartenbreddels - just to check, what kind of issues do you see on MacOS X typically? Just performance, or actual crashes?

maartenbreddels:
Performance only.

PR #1635, opened 2018-03-29
Refactor spectrum tool

astrofrog:
This is a work in progress, but the aim is to add a profile viewer which is a proper data viewer and to use this instead of the current spectrum tool.

Things that still need to be done:

* [x] Add a way to have a vertical line that connects to a slice in an image viewer
* [x] Add a way to create moment maps based on an interval in the profile
* [x] Add a way to do interactive fitting
* [x] Add proper test suite
* [x] Add changelog entry/what's new entry
* [x] Show world coordinates on axis
* [x] Add a 'reference data' combo to avoid confusion
* [x] Don't show missing values in profile as zero
* [x] Don't enable navigation tool without sidebar present
* [x] Added tests for data linking/world coordinates


astrofrog:
I'll fix the failing Travis job (Matplotlib 1.5) in a follow-up PR that will make other improvements too

Issue #1569, opened 2018-03-08
Add 'Save Data' to File menu


astrofrog:
Currently it is only available in the toolbar. Generally speaking we should ideally have a one-to-one mapping between the toolbar and the File menu.

PR #1636, opened 2018-03-29
Improve UI for saving/exporting data

astrofrog:
This is to address https://github.com/spacetelescope/cubeviz/issues/324

This makes several changes:

* The button for saving data is now called 'Export Data/Subsets' to make it clear subsets can be exported, and using 'export' instead of 'save' makes it clearer that this will not overwrite the existing data files:

screen shot 2018-03-29 at 11 46 29

* 'Export Data/Subsets' is now also available via the 'File' dialog:

screen shot 2018-03-29 at 11 57 32

* Finally I have included a sentence at the bottom of the save dialog to make it clearer that the user will be prompted for a filename, and changed the button from 'OK' to 'Export':

screen shot 2018-03-29 at 11 46 33

@brechmos-stsci @kassin - does this seem ok?

Fixes https://github.com/glue-viz/glue/issues/1569

### DIFF
--- a/glue/app/qt/application.py
+++ b/glue/app/qt/application.py
@@ -664,7 +664,7 @@ class GlueApplication(Application, QtWidgets.QMainWindow):
         menu.addAction(self._actions['session_save'])
         menu.addAction(self._actions['export_data'])
         if 'session_export' in self._actions:
-            submenu = menu.addMenu("E&xport")
+            submenu = menu.addMenu("Advanced E&xporters")
             for a in self._actions['session_export']:
                 submenu.addAction(a)
         menu.addSeparator()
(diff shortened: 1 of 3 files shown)


brechmos-stsci:
If not merged, I would suggest using the work "Choose" everywhere you have "Pick".  

For wording, I wonder about 'Once you click "Export" you will be prompted to choose the filename.'

I wonder if the "Export" button should be "Next, Choose Filename" or something.  "Export" to me still feels like it is going to do the export.  I completely get you have the sentence above, but I wonder if the "Export" button should be more explicit.

astrofrog:
@brechmos-stsci - thanks for the comments! With the rewording of the last sentence as you suggest, I think that things are ok now as if users hesitate to press 'Export' they are likely to read the last sentence (I think?)

brechmos-stsci:
I think I am fine with Export.  We can always come back to it later... merge away (not that you need my permission :-) )

PR #1637, opened 2018-03-29
Split component arithmetic into its own dialog

astrofrog:
User testing has shown that the component/attribute manager can be confusing - instead we should have a dedicated dialog for arithmetic attributes and then a separate component manager that can be used to rename/reorder components.

Note that the button now mentions arithmetic attributes instead of 'add/edit components' which was unclear:

screen shot 2018-03-29 at 16 41 25

and in the dialog I refer to attributes instead of components:

screen shot 2018-03-29 at 16 43 30

@brechmos-stsci - could you take this for a spin with cubeviz and let me know if you are happy with this UI? For now you'll need to click the button in the screenshot above but we can also make it work from the drop-down in cubeviz once/if this PR is merged.


brechmos-stsci:
I think this is more clear now.  One suggestion I would make is could you add examples onto the dialog?  What  I had was:
image

I think it might be really helpful to have 3-4 examples of what people can do (based on the data available.

astrofrog:
@brechmos-stsci - would something like this be ok? Note that the text with the examples etc can actually be customized via a class-level attribute, so if needed we could potentially override these in cubeviz (but I'd like to get this as good as possible here first):

screen shot 2018-03-29 at 17 40 59

Currently the examples use the shortest attribute name that can be found for readability :)

brechmos-stsci:
That looks good.  Maybe use "FLUX" in the examples.  Actually what I did in mine was find the SCI/FLUX/DATA extension name and then dynamically set in the example expressions so they could see *exactly* what it should be like.

astrofrog:
@brechmos-stsci - at the moment the editor will dynamically get the name of the actual data attributes, so this is similar to what you are doing (it's just that in the example I chose the attribute names weren't interesting)

PR #1641, opened 2018-04-03
Improve mouse profile performance

astrofrog:
This improves the performance of computing profiles when using the single pixel extractor 'live'


astrofrog:
Merging this and will fix the PyQt4 failure in a follow-up PR.

PR #1642, opened 2018-04-04
Only deactivate default mouse mode if new tool is checkable

drdavella:
The default mouse mode should only be deactivated in cases where the new mouse mode corresponds to a checkable tool. This ensures that tools that do not enable persistent effects do not disrupt the functioning of the default mouse mode.

This fixes https://github.com/spacetelescope/cubeviz/issues/353.


drdavella:
Done!

Issue #1509, opened 2018-01-26
ComponentIDComboHelper with "None" option


brechmos-stsci:
It would be really nice to have an option for the helper to have a "None" option.  For example in cubeviz there is a combo box related to a viewer and the combo box shows the data's component options so we can switch between them. I would like to have another option for "None" (or some other word) so that we can have the ability to show nothing *or* the data components.

PR #1661, opened 2018-04-15
Make it possible to have a 'None' entry in ComponentIDComboHelper

astrofrog:
Fixes https://github.com/glue-viz/glue/issues/1509

@brechmos-stsci - do you have any thoughts on what the 'None' entry should show up as in the combo - should it be an empty string, or should it actually say 'None'?


astrofrog:
@brechmos-stsci - in the end I made it so that you can customize what you want shown for the None entry by simply setting ``none=`` to a string.

PR #1662, opened 2018-04-15
Fix compatibility with PySide2 and drop support for PySide and PyQt4

astrofrog:
PySide2 is now available from:

```
conda install -c conda-forge pyside2
```


astrofrog:
e31

PR #1663, opened 2018-04-15
WIP: Refactor PV slicing tool

astrofrog:
Still very early days... Remaining things to do in this PR:

* [x] Rebase once #1662 is merged
* [x] Extract slice for all attributes from current reference data
* [x] Automatically open an image viewer if needed
* [x] Add status bar documentation
* [ ] Rename tool to not be PV-specific
* [ ] Add tests
* [ ] Add changelog entry
* [ ] Fix the fact that rectangular selections are somehow propagated to PV slices
* [ ] Make subsets work properly
* [ ] When closing window with a slice and then updating the slice, do we re-open a new window?
* [ ] How do we deal with multiple datasets being present? do we extract all datasets as attributes in the new PV slice at resolution of reference data?
* [ ] Seeing C/C++ reference error when closing a PV slice window and then re-openng it with data and using the mouse over path
* [ ] Use APE 14 WCS to get nice labels on sliced data
* [ ] Don't show PV sliced data in link editor? (since links are implicit). Or show them with a link.
* [ ] Test that things work fine in a multi-attribute dataset

Things that could be done in future:

* [x] Extract slice for all visible layers (although only pixel-aligned ones)
* [x] Extract slice for reprojected data
* [ ] Preserve derived components
* [x] Make PVSliceData components be dynamically evaluated
* [ ] Allow thick slices
* [ ] Have a way of selecting whether a new slice is created or modifying the old one (for now we can just have one path per viewer and have that always update)

Other thoughts

* For now, for simplicity, don't use links between PV sliced data and parent data - allow subsets on main data to be viewer in slice but selections in slice shouldn't propagate back (except maybe via the non-slice axis)

### DIFF
--- a/glue/plugins/tools/pv_slicer/qt/pv_slicer.py
+++ b/glue/plugins/tools/pv_slicer/qt/pv_slicer.py
@@ -64,7 +64,7 @@ class PVSlicerMode(PathMode):
                 for pvdata in self.viewer.session.data_collection:
                     if isinstance(pvdata, PVSlicedData):
                         if pvdata.original_data is data:
-                            pvdata.original_data = self.viewer.state.reference_data
+                            pvdata.original_data = data
                             pvdata.x_att = self.viewer.state.x_att
                             pvdata.y_att = self.viewer.state.y_att
                             pvdata.set_xy(vx, vy)


astrofrog:
I'm currently stuck on an issue with Matplotlib getting stuck in an infinite loop in the following case:

* Set up an RGB image with three cubes
* Extract a path
* Update the path

When the path is updated, the second viewer then gets stuck in a drawing loop and never renders fully. Some things that seem to matter - if in the state class for the image viewer I return a random array instead of the fixed resolution buffer after the initial call, then there is no infinite loop. If I do the same in the PV SlicedData, it sometimes helps and sometimes not (randomly). This makes me think there is a race condition of some kind.

astrofrog:
Actually I think I found where the issue comes from - 

```

PR #1704, opened 2018-05-05
Fixed a bug that caused data combos to not update when a data object was renamed

### LINKED ISSUE #1682: Combo box for reference data in image viewer doesn't update when renaming
When renaming a dataset, the labels in the reference combo box don't change

astrofrog:
Fixes #1682


astrofrog:
Note: this doesn't yet actually fix the issue described in #1682!

PR #1713, opened 2018-05-09
WIP: abstract computation/views out of data viewers

astrofrog:
### About

This is something we've planned for quite a while - the idea is that we should avoid/minimize as much as possible direct data access in the viewers, and move computations on the data to the Data/Component classes so that in future we can make them remote.

### Implementation

At the moment this PR adds two important methods to the ``Data`` class:

* ``compute_statistic``: this is a method that given a component name, the name of a statistic (e.g. mean or median), the axes to operate over, and some other flags, will compute a statistic - this could be a single value or an array depending on ``axis``. This optionally takes the subset state to apply before computing the statistic.

* ``compute_histogram``: this is a method that given one or more components, as well as numbers of bins and ranges along each direction, will compute an n-dimensional histogram (again, this optionally takes a subset state)

The idea is that in a ``Data`` subclass, these could be replaced by methods that request this from a remote server.

In addition, this defines an abstract base class for components with a minimal set of properties and methods that a subclass should have. Note that components are still (for now) required to expose an API that looks nd-like (via shape, size, ndim, and __getitem__) **but** the underlying data doesn't have to be structured. So in essence, pixel coordinates might be 'virtual' in that they don't correspond to actual individual elements in memory. The ``__getitem__`` method could also be made to interface with a remote server.

### Example

This always just returns random data for all requests but is used to check that there are no points at which the whole data is accessed:

```python
import numpy as np

from glue.core import Data, DataCollection
from glue.viewers.image.qt import ImageViewer
from glue.viewers.histogram.qt import HistogramViewer
from glue.viewers.profile.qt import ProfileViewer
from glue.viewers.scatter.qt import ScatterViewer
from glue.core.component import BaseComponent
from glue.utils import view_shape


class RandomComponent(BaseComponent):

    @property
    def shape(self):
        return (10000, 10000, 10000)

    @property
    def ndim(self):
        return 3

    @property
    def size(self):
        return int(1e12)

    @property
    def numeric(self):
        return True

    @property
    def categorical(self):
        return False

    @property
    def datetime(self):
        return False

    def __getitem__(self, key):
        return np.random.random(view_shape((10000, 10000, 10000), key))

    def ravel(self):
        raise NotImplementedError()


class RandomData(Data):

    def compute_histogram(self, cid, range=None, bins=None, log=False, subset_state=None, subset_group=None):
        return np.random.random(bins) * 100

    def compute_statistic(self, statistic, cid, axis=None, finite=True, positive=False, subset_state=None, percentile=None):
        if axis is None:
            if statistic == 'min':
                return 0
            elif statistic == 'max':
                return 1
            elif statistic == 'mean' or statistic == 'median':
                return 0.5
            elif statistic == 'percentile':
                return percentile / 100
            elif statistic == 'sum':
                return 1e9 / 2
        else:
            final_shape = tuple(self.shape[i] for i in range(self.ndim) if i not in axis)
            return np.random.random(final_shape)


d = RandomData()
d.add_component(RandomComponent(), 'random')
dc = DataCollection([d])

ga = GlueApplication(dc)

image = ga.new_data_viewer(ImageViewer)
image.add_data(d)

histogram = ga.new_data_viewer(HistogramViewer)
histogram.add_data(d)

profile = ga.new_data_viewer(ProfileViewer)
profile.add_data(d)

scatter = ga.new_data_viewer(ScatterViewer)
scatter.add_data(d)

# show the GUI
ga.start()
```

### Main action items

* [ ] Implement ``Data.compute_histogram``
* [ ] Re-instate the ability to use a subset of points for the 2D histogram in scatter density
* [ ] Add support for weights in ``compute_histogram``
* [ ] Make sure tests pass with default implementation
* [ ] Build a proof-of-concept example of Data object which can communicate with a remote server 
* [ ] Write documentation page with the specification
* [ ] Check how things work in the 3D case
* [ ] Add a way to say 'give me X representative points'
* [ ] Figure out how to deal with ``roi_to_subset`` on ``Component``

### DIFF
--- a/glue/core/component.py
+++ b/glue/core/component.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import abc
 import logging
 import warnings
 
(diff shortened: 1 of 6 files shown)


astrofrog:
This is basically done:

http://docs.glueviz.org/en/latest/developer_guide/data.html

Or there are smaller issues about the unimplemented parts, so closing this.

PR #1723, opened 2018-05-16
HDU Header for components

robelgeda:
This will introduce header files from `Data.coords.header` into the `FITS (1 component/HDU)` data exporter. It also adds a function` make_component_header` that fills in component specific information into the header.


glue/core/data_exporters/gridded_fits.py
```diff
@@ -28,6 +44,8 @@ def fits_writer(filename, data, components=None):
     else:
         mask = None
 
+    data_header = data.coords.header if hasattr(data.coords, "header") else None
```
astrofrog: A more robust way here would be to check if ``data.coords`` is an instance of ``glue.core.coordinates. WCSCoordinates`` (rather than just checking whether ``header`` exists).

glue/core/data_factories/fits.py
```diff
@@ -114,14 +114,16 @@ def new_data(suffix=True):
             if is_image_hdu(hdu):
                 shape = hdu.data.shape
                 coords = coordinates_from_header(hdu.header)
+                units = hdu.header["BUNIT"] if "BUNIT" in hdu.header else None
```
astrofrog: I think you can write this as ``hdu.header.get('BUNIT')`` as that will return ``None`` if the key doesn't exist.

glue/core/data_factories/fits.py
```diff
@@ -205,7 +207,9 @@ def casalike_cube(filename, **kwargs):
         header = hdulist[0].header
     result.coords = coordinates_from_header(header)
     for i in range(array.shape[0]):
-        result.add_component(array[[i]], label='STOKES %i' % i)
+        units = header["BUNIT"] if "BUNIT" in header else None
```
astrofrog: Same comment as above regarding using ``.get``

glue/core/data_exporters/gridded_fits.py
```diff
@@ -9,6 +9,22 @@
...
 
 
+def make_component_header(component, header):
+    """
+    Function that extracts information from components
+    and adds it to the data header. The input header is
+    expected to come from Data.coords.header by default.
+    :param component: glue Component
+    :param header: astropy.io.fits.header.Header
+    :return:
```
astrofrog: I'm currently trying to switch over to the Numpydoc format (the same one used by astropy), so could you update this docstring accordingly?

glue/core/data_factories/fits.py
```diff
@@ -114,7 +114,7 @@ def new_data(suffix=True):
             if is_image_hdu(hdu):
                 shape = hdu.data.shape
                 coords = coordinates_from_header(hdu.header)
-                units = hdu.header["BUNIT"] if "BUNIT" in hdu.header else None
+                units = hdu.header.get('BUNIT')if "BUNIT" in hdu.header else None
```
astrofrog: Just to clarify, you can just do:

```
units = hdu.header.get('BUNIT')
```

and you don't need the if statement part.

robelgeda:
Ok thanks. Fixing now

astrofrog:
Before I merge this, could you add a test? Thanks!

robelgeda:
@astrofrog  I have added some tests

astrofrog:
@robelgeda - thanks! is the ``bunit.fits`` file actually needed? If not, could you remove it from the last commit?

robelgeda:
It is needed, it is a small 1D fits file with `BUNIT` inside of it

Issue #310, opened 2014-04-13
Asynchronous profile extraction


ChrisBeaumont:
Explore if this leads to a more responsive UI

PR #1736, opened 2018-05-20
Asynchronous profile computation

astrofrog:
This is a proof of concept for dealing with asynchronous computation of the profiles.

* [x] Use a single-shot ``QTimer`` to emit the layer computation signal so that we only show this for cases where the computation takes more than e.g. 0.5 seconds
* [x] Create computation started and ended signals, even if we don't really use them for now
* [x] Move some of the layer artist functionality to the generic matplotlib layer artist class so that it can be used for other matplotlib plot types
* [x] Add back caching for the state method that computes the profile (as or the image viewer)
* [ ] Try and implement this for the histogram
* [x] Avoid direct imports to timer and thread in layer artists

Fixes https://github.com/glue-viz/glue/issues/310

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,8 @@ Full changelog
 v0.14.0 (unreleased)
 --------------------
 
-* Compute profiles asynchronously to avoid holding up the UI. [#1736]
+* Compute profiles asynchronously to avoid holding up the UI,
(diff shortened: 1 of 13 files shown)


astrofrog:
I'll the histogram later as it'll be easier to do after the data abstraction layer is done.

PR #1770, opened 2018-06-15
Make sure we clean up things in viewers when closing viewers via the GUI

astrofrog:
This bypasses the ``close`` method so we need to do this in the ``closeEvent`` instead.


astrofrog:
The failures are unrelated and due to Plotly

PR #1788, opened 2018-06-25
Deprecate primary_components and visible_components

astrofrog:
There is a lot of duplication/overlap in the various component list properties, so deprecating two old ones that don't really make so much sense anymore.


glue/viewers/image/state.py
```diff
@@ -681,7 +681,7 @@ def __init__(self, layer=None, viewer_state=None, **kwargs):
     def _update_attribute(self, *args):
         if self.layer is not None:
             self.attribute_att_helper.set_multiple_data([self.layer])
-            self.attribute = self.layer.visible_components[0]
+            self.attribute = self.layer.main_components[0]
```
astrofrog: Double check this

PR #1851, opened 2018-08-28
make the BaseCartesianData.compute_histogram function accept weights

ngoldbaum:
This makes the signature of this function match the signature of `Data.compute_histogram`.


astrofrog:
Thanks!

PR #1870, opened 2018-09-05
add more useful error message for selection

eteq:
This makes a small change to yield a slightly more useful error message when you make an invalid choice: it shows the choices.

I'm not sure if there's somewhere upstream this belongs - it's in an "external" package, but I'm not clear what "echo" is a reference to...

### DIFF
--- a/glue/external/echo/selection.py
+++ b/glue/external/echo/selection.py
@@ -26,12 +26,13 @@ class SelectionCallbackProperty(CallbackProperty):
         self._force_next_sync = WeakKeyDictionary()
 
     def __set__(self, instance, value):
-        choices = self.get_choices(instance)
-        # We do the 'any' call in the following because 'value in choices'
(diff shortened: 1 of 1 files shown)


glue/external/echo/selection.py
```diff
@@ -26,8 +26,9 @@ def __init__(self, default_index=0, choices=None, display_func=None, **kwargs):
         self._force_next_sync = WeakKeyDictionary()
 
     def __set__(self, instance, value):
-        if value is not None and value not in self.get_choices(instance):
-            raise ValueError('value {0} is not in valid choices'.format(value))
+        choices = self.get_choices(instance)
+        if value is not None and value not in choices:
+            raise ValueError('value {0} is not in valid choices: {}'.format(value, choices))
```
astrofrog: You can't mix numbered and un-numbered formatting - can you put ``{1}`` instead? (just to be more explicit)

PR #1878, opened 2018-10-11
Edit subset trigger

maartenbreddels:
Assigning to `edit_subset` will cause more Messages, which I need to refresh the UI, now only a message is send after the new_subset_group call, but not the assignment to `edit_subset`.


astrofrog:
@maartenbreddels - can you remove the first commit?

maartenbreddels:
👍

PR #1896, opened 2018-11-08
Add numpy to setup_requires to workaround bottleneck bug

drdavella:
See https://github.com/kwgoodman/bottleneck/issues/195 for description of the packaging bug in `bottleneck` that causes this problem.


drdavella:
I thought this worked at one time, but it actually seems to make no difference, so I'm closing.

Issue #1734, opened 2018-05-20
Have a way to globally set font size


astrofrog:
It would be nice to have in the preferences panel a way to set the overall font size in the Qt application.

astrofrog:
Fixed in https://github.com/glue-viz/glue/pull/1905 - thanks @corytodd!

PR #1905, opened 2018-11-25
Add global widget font preference

corytodd:
This adds a font size feature that affects all Qt widgets' font size.
The default font size (8) has been preserved. The option is listed on the main
preferences page and uses the broadcast feature to update the UI
immediately on configuration save. This does _not_ affect the font size of
viewers or their axes.

#1734 is the only reference I've found for this feature. I fired up Glue for the 
first time today and thought the font was a bit small so here we are.


astrofrog:
@corytodd - thanks, this is great! Only one comment: the default font-size currently is platform dependent, and in fact the optimal font size on each platform is different. Therefore, rather than setting the default to 8, I think it should be set to whatever the default is here:

https://github.com/glue-viz/glue/blob/master/glue/utils/qt/app.py#L43

when the application starts up. Maybe you can just override the default value there?

corytodd:
@astrofrog absolutely, I'll look into that change and make sure the default font size is set appropriately.

corytodd:
Done!

Thanks to your and your team for this wonderful project by the way. The code is very easy to follow and it looks like a pleasure to work with.

astrofrog:
@corytodd - thanks for this great contribution! It works nicely - there are a couple of places where the font size doesn't seem to propagate, due to a Qt bug, but I can address that in a follow-up PR.

Issue #1910, opened 2018-12-20
Can't open demo data file  w5_psc.vot


aaronoppenheimer:
The "getting started" guide advises opening w5_psc.vot from the demo data zip file. When I try to open I see a failure, with this message:

```
Error message: Cannot convert fill_value nan to dtype int32

Traceback (most recent call last):
  File "C:\Users\Aaron.Oppenheimer\AppData\Local\Continuum\anaconda2\envs\glueviz\lib\site-packages\numpy\ma\core.py", line 479, in _check_fill_value
    fill_value = np.array(fill_value, copy=False, dtype=ndtype)
ValueError: cannot convert float NaN to integer

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Aaron.Oppenheimer\AppData\Local\Continuum\anaconda2\envs\glueviz\lib\site-packages\glue\dialogs\data_wizard\qt\data_wizard_dialog.py", line 36, in data_wizard
    result = gdd.load_data()
  File "C:\Users\Aaron.Oppenheimer\AppData\Local\Continuum\anaconda2\envs\glueviz\lib\site-packages\glue\dialogs\data_wizard\qt\data_wizard_dialog.py", line 119, in load_data
    d = load_data(path, factory=fac.function)
  File "C:\Users\Aaron.Oppenheimer\AppData\Local\Continuum\anaconda2\envs\glueviz\lib\site-packages\glue\core\data_factories\helpers.py", line 251, in load_data
    d = as_list(factory(path, **kwargs))
  File "C:\Users\Aaron.Oppenheimer\AppData\Local\Continuum\anaconda2\envs\glueviz\lib\site-packages\glue\core\data_factories\astropy_table.py", line 93, in astropy_tabular_data_votable
    return astropy_tabular_data(*args, **kwargs)
  File "C:\Users\Aaron.Oppenheimer\AppData\Local\Continuum\anaconda2\envs\glueviz\lib\site-packages\glue\core\data_factories\astropy_table.py", line 78, in astropy_tabular_data
    c = c.filled(fill_value=np.nan)
  File "C:\Users\Aaron.Oppenheimer\AppData\Local\Continuum\anaconda2\envs\glueviz\lib\site-packages\astropy\table\column.py", line 1265, in filled
    data = super().filled(fill_value)
[...]

PR #1911, opened 2018-12-20
Fix issue with reading in VO Tables

astrofrog:
Due to a change in either Numpy or Astropy (haven't had a chance to investigate which), reading of VO Tables with non-float columns was broken. This PR fixes this.

Fixes https://github.com/glue-viz/glue/issues/1910


astrofrog:
The Travis failure is unrelated

PR #1949, opened 2019-03-06
Added a class to represent a derived dataset produced by indexing

astrofrog:
With the abstract data interface, it's now pretty easy to create so-called *derived* datasets which are computed on-the-fly from other datasets. We can use this to reduce dimensionality, including making PV slices as proper datasets, and so on.

This PR adds a first class that simply slices along some dimensions to reduce the dimensionality.

This is still a work in progress as ``compute_histogram`` and ``compute_statistic`` need to be updated accordingly.

To try it out:

```python
from glue.core import DataCollection
from glue.core.data_factories import load_data
from glue.viewers.image.qt import ImageViewer
from glue.core.data_derived import IndexedData

data1 = load_data('/Users/tom/Data/PPV/l1448_13co_small.fits')
data2 = IndexedData(data1, (None, 10, None))

dc = DataCollection([data1, data2])

ga = GlueApplication(dc)
ga.new_data_viewer(ImageViewer, data=data1)
ga.new_data_viewer(ImageViewer, data=data2)
ga.start()

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,9 @@ v0.15.0 (unreleased)
 
 * Improve the user interface for the link editor. [#1934]
 
+* Added a new ``IndexedData`` class to represent a derived dataset produced
+  by indexing a higher-dimensional dataset. [#1953]
+
 v0.14.3 (unreleased)
 --------------------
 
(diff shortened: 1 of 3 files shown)


astrofrog:
The main thing to do now is make sure that world coordinates work.

astrofrog:
Which I will do in a separate PR at some point! This will be more complex as it relates to WCS slicing.

PR #1953, opened 2019-03-07
EXP: Updated pixel selection tool to make derived dataset

astrofrog:
This is an experimental pull request to see if we can make the pixel selection tool make a derived dataset instead of a subset. Currently the crosshairs are missing from the image viewer, which will need to be fixed. Also, we may want to keep the option to make a subset. Anyway this is just for testing for now.

This should be rebased once #1949 is merged

### DIFF
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -1211,6 +1211,11 @@ class Data(BaseCartesianData):
 
     def get_data(self, cid, view=None):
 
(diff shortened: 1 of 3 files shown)


astrofrog:
For now I've put this in https://github.com/glue-viz/glue-solar, so closing this. However in future I think once it's polished we can move it here.

PR #1963, opened 2019-03-22
Pass the log state to the histogram machinery

### LINKED ISSUE #1962: Density plots in scatter 2d not using the log parameter
When doing a 2D scatter plot, the "marker" mode gives the expected result but the "density map" and the "density map or markers" do not. They seem to do something fishy when the log attribute is switched on in one or both axis.

The density maps should take into account the log parameter of the axis.
markers
Density map
density
Density or markers
density_or_markers

cphyc:
This fixes #1962.

### DIFF
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -1619,13 +1619,13 @@ class Data(BaseCartesianData):
             return np.zeros(bins)
 
         if log is not None and log[0]:
-            xmin = np.log10(xmin)
-            xmax = np.log10(xmax)
+            xmin = np.log10(xmin) if xmin > 0 else np.log10(xmax) - 10
+            xmax = np.log10(xmax) if xmax > 0 else np.log10(xmin) + 10
             x = np.log10(x)
 
         if ndim > 1 and log is not None and log[1]:
-            ymin = np.log10(ymin)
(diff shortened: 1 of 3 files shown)


cphyc:
Using the same dataset as #1962 I know have the following plot
working_example

glue/core/data.py
```diff
@@ -1619,13 +1619,13 @@ def compute_histogram(self, cids, weights=None, range=None, bins=None, log=None,
             return np.zeros(bins)
 
         if log is not None and log[0]:
-            xmin = np.log10(xmin)
-            xmax = np.log10(xmax)
+            xmin = np.log10(xmin) if xmin > 0 else np.log10(xmax) - 10
+            xmax = np.log10(xmax) if xmax > 0 else np.log10(xmin) + 10
```
astrofrog: This might give strange results if both xmin and xmax are zero for some reason - I think it might be best to actually just return an array of NaN values if either xmin or xmax are negative rather than modify the limits. You can construct this array efficiently with:

```
broadcast_to(np.nan, bins)
```

could you update this here and below?

astrofrog:
I'll go ahead and merge this, thanks @cphyc!

PR #1984, opened 2019-04-16
Move as much as possible related to tools out of qt-specific directories

astrofrog:
This is so that we can use them in https://github.com/glue-viz/glue-jupyter/pull/92 without having to import Qt


astrofrog:
The remaining Travis failures are unrelated, I will investigate/fix in a separate PR.

PR #2052, opened 2019-07-31
Make sure autolinker is called when using qglue

astrofrog:
I need to check if there's any other places this is missing and should probably also add a button to do this manually if needed.


Cadair:
This doesn't seem to have worked...

astrofrog:
Just for the record, the change in https://github.com/glue-viz/glue/pull/2052/commits/076450ba282fc9e8c860ce99e25e40829d5bb0ba is needed following a change in Python 3.7.4 (specifically https://github.com/python/cpython/pull/13589). If time allows, I will investigate whether this is expected or a CPython bug.

Cadair:
:tada:

PR #2058, opened 2019-08-02
WIP: Provide kwarg to disable image cache

ejeschke:
This adds a keyword parameter to disable the use of the image cache.

### DIFF
--- a/glue/viewers/image/state.py
+++ b/glue/viewers/image/state.py
@@ -359,7 +359,7 @@ class BaseImageLayerState(MatplotlibLayerState):
         else:
             return view_shape(shape_slice, view)
 
-    def get_sliced_data(self, view=None):
+    def get_sliced_data(self, view=None, use_cache=True):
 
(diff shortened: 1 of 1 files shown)


ejeschke:
Needs a rebase against master and bit of work to see how this can work with the recent updates to Glue.

astrofrog:
@ejeschke - just to check, what is the motivation for this? Does the use of the cache cause issues with ginga?

ejeschke:
Yes. Ginga's slicing produces an index that works with numpy arrays, but it is not strictly a slice with start, stop, step--it is a numpy array itself.  Unfortunately, it seems that one cannot use that as a key (probably since numpy arrays are mutable).

In the version of Glue I wrote this for, this keyword seemed like a fairly simple workaround to avoid it trying to use the Ginga "slice" as a cache index.

ejeschke:
I may resubmit as a different PR, if I figure out the compatibility.

PR #2076, opened 2020-01-07
CI: Added Python 3.8, PyQt5 5.14 and PySide2 5.14 configurations

astrofrog:
This makes sure we are compatible with the latest Python and Qt versions


astrofrog:
There is a genuine issue on Windows with Python 3.8, so I'll need to investigate before this can be merged

astrofrog:
I've opened an issue for the Windows problem with qtconsole here: https://github.com/jupyter/qtconsole/issues/400 and have disabled that build on Azure for now.

Issue #1931, opened 2019-02-07
Make "replace selection" be default button when creating new subset


catherinezucker:
If you were clicked on e.g. "remove selection" in the boolean logic panel and then decided to "None/Create New Dataset" it wouldn't work because it doesn't auto switch to "replace selection." I think this should be the default!

catherinezucker:
Any update on this? Alyssa & I are struggling with this default option at AMNH this week, and think this is probably tripping up many users.

astrofrog:
@catherinezucker - FYI @specmicp made a pull request (now merged) which temporarily switches to replace mode when making a new selection.

catherinezucker:
AWESOME! Thank you @specmicp @astrofrog!!!

PR #2090, opened 2020-02-25
Fix bug #1931: when creating a new subset, use ReplaceMode

specmicp:
When a new subset is created, override the mode to use ReplaceMode.

The user does not have to remember to reset everytime the mode manually. The mode is only temporarily overriden for the creation of the new subset.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ v0.16.0 (unreleased)
 
 * Python 2.7 and 3.5 are no longer supported. [#2075]
 
+* Always use replace mode when creating a new subset. [#2090]
+
 * Fixed bugs in the profile viewer that occurred when using the profile viewer and
   setting reference_data to a different value than the default one. [#2078]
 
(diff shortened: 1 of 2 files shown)


astrofrog:
Thanks for the pull request! Unless you have a chance to do this, I will take a look at the test failures early next week to see what we need to fix before merging this.

specmicp:
The first error:  "ERROR at setup of TestGlueApplication.test_new_tabs" is I think a problem with upstream: https://github.com/jupyter/jupyter_client/issues/523

The solution to pass the test (at least on my machine) was to downgrade jupyter_client from 6.0.0 to 5.3.4 (available in pip)

specmicp:
the failing test glue/viewers/image/qt/tests/test_data_viewer.py is only failing with matplotlib 3.1.3, not matplotlib 3.1.2

Here is the changelog: https://matplotlib.org/3.1.3/users/whats_new.html

PR #2116, opened 2020-04-02
Don't sync color and transparency of image layers by default

astrofrog:
@nmearl - this should prevent the sync issues you were seeing


nmearl:
👍 Awesome! Thanks.

PR #2119, opened 2020-04-08
Added tests for glue.external.echo

astrofrog:
This adds the tests for echo here to make sure we haven't broken anything (since glue.external.echo has diverged a little from echo). In future we should unbundle echo and release it properly as a standalone package.


astrofrog:
Closing - I instead plan to add echo as a proper dependency

Issue #1883, opened 2018-10-19
Implement ability to show/hide axes


astrofrog:
In https://github.com/glue-viz/glue/pull/1876, @maartenbreddels added a new ``show_axes`` property for 2D viewers - we should honor this option in the Qt interface.

PR #2128, opened 2020-04-24
Hide Matplotlib Axes checkbox

jsub1:
This creates a checkbox that allows the user to hide 2D matplotlib axis. Allows to hide/show axis to included when exporting as script. Also adds test for hiding the axis and exporting.


jsub1:
For issue #1883

astrofrog (commented):
Thanks for working on this! I finally had a chance to take this for a spin, and it works well! However, two requests before we can go ahead and merge:

* When axes are not shown, we should also remove the white padding around the image. If you take a look at the ``init_mpl`` function you'll see it calls ``freeze_margins``. I think when you turn off the axes you might want to change that to freeze the margins to zero, and then change the frozen margins back to what they are currently if axes are turned back on.

* Currently if you load the L1448 cube, and show velocity on the x axis and say RA on the y axis, the ``MatplotlibImageMixin._set_wcs`` method is still getting called even when the axes are hidden. In principle we could avoid this call, or at least exit early inside that method if the axes are now shown - and then if the axes are turned on again, we could explicitly call ``_set_wcs``. This might make the slider even smoother for cases where the coordinate labels take a while to compute.

* Could you double check that in astropy.visualization.wcs, the ticks and so on are not being computed/drawn when the axes are hidden as is done here? I just want to make sure it's not that they are being computed but then not shown.

* It looks like there are some real code style issues and potentially real test failures?

Thanks!

astrofrog (changes requested):
Could you rebase and remove the last commit since that CI issue should now be fixed?

PR #2134, opened 2020-05-08
MNT: Ignore import error in pytest collection

pllim:
This is an automated update made by the `batchpr` tool :robot: to future-proof test collection behavior `pytest-doctestplus` - feel free to close if it doesn't look good or isn't relevant! You can report issues to @pllim.


astrofrog:
I don't think we need this here as there doesn't seem to be any import errors with the latest pytest :) Thanks for opening this though!

PR #2136, opened 2020-05-11
Add support for dask arrays and improve how image limits are determined with big datasets in mind

astrofrog:
This PR adds initial support for Dask datasets by adding a ``DaskComponent`` class, and also makes it so the image viewer defaults to using the first slice of a cube to determine the limits rather than a random set of pixels in the whole cube.

I'm not fully happy with this yet - it's very image-viewer-centric and doesn't really solve the issue of using other viewers with large Dask arrays. I'm also not a big fan of changing the current limits behavior for plain Numpy arrays which do support fancy indexing since the random sampling for limits has worked well until now.

I'll have a think to see if I can find a better method...

This requires https://github.com/glue-viz/echo/pull/26


astrofrog:
Closing in favor of #2137

PR #2139, opened 2020-05-12
[WIP] Initial setup for custom glue-solar splash screen

kakirastern:
## Description

To form part of GSoC 2020 glue-solar project. The goal is to enable a custom `glue-solar` splash screen in lieu of the default `glue` one if the `glue-solar` plugin has been detected, if such an operation is permissible.


kakirastern:
Looks like the logic used in the PR will not work with the current setup of `glue`, so mentors have recommended against it. Will consider the possibility of developing a custom viewer in the near future.

astrofrog:
@kakirastern - thanks! For now I'll close this, but an alternative if you really want to customize the splash screen would be to make it easier to customize the path to the image used to replace the current splash screen

kakirastern:
Thanks @astrofrog for the tip! Really appreciate it 👍

kakirastern:
Will definitely follow up with a PR based on your suggestion soon.

Issue #1348, opened 2017-07-14
Recursion limit when combining too many selections


astrofrog:
If a user makes a selection then refines it based on ~50-100 further logical selections, they will hit a recursion limit since the logical selections are stored as a tree. We should probably transform selections into a flat list to avoid this.

PR #2140, opened 2020-05-13
Non-recursive composite subsets

jsub1:
Fix for #1348. This PR adds a helper function that allows traversing the state tree in composite subset states without making recursive function calls.


astrofrog:
I'll try reviewing this soon, but in the mean time, could you fix the code style issues and also add a test that would normally have failed due to a recursion error?

PR #2144, opened 2020-05-20
Documentation for the new legend feature

### LINKED ISSUE #2143: Add entry to 'What's new' page for legend functionality
It would be great to add an entry to the 'What's new in glue v0.16?' page in the docs:

https://github.com/glue-viz/glue/blob/master/doc/whatsnew/whatsnew.rst

for the new legend functionality. In addition, I forgot to say that an entry should be added to the ``CHANGES`` file at the root of the repository.

@specmicp - would you like to take a stab at a 'What's new' entry (optionally including a figure)?

specmicp:
Resolves #2143 

I was looking at the documentation, would it be good to add a mention of the legend API in the customization guide for creating new viewer ? If so, where would be the best place ?


astrofrog:
@specmicp - thanks for this PR! In terms of docs under the section for making a new viewer, I think making a new dedicated page in this section would be good:

http://docs.glueviz.org/en/stable/#advanced-customization

(we already have a dedicated page for tools and toolbars for example). Would you like to add this here or in a follow-up PR?

specmicp:
@astrofrog That sounds like a good idea.

I think you can merge this PR. I'll do this suggestion in another PR to not delay this one too much, also in case another feature need to be added soon. Getting the python export script to look like the viewer in a simple enough way is quite tricky, although I think we got to a pretty good solution so it shouldn't change too much.

PR #2146, opened 2020-05-21
Legend background and foreground colors can be modified

### LINKED ISSUE #2145: Legend and background color
Viewers can have a custom background/foreground color (which can be changed in the preferences), for example white on black instead of black on white. We might want to consider making the legend follow this so that the background and text color match the rest of the figure:

Screenshot from 2020-05-20 15-01-34

cc @specmicp

specmicp:
Resolves #2145 

- Change in visual properties of the legend do not require to get the handles anymore
- Background and foreground can be changed
- Defaults for these properties are taken from the settings
- The legend is also set to be draggable

### DIFF
--- a/glue/viewers/matplotlib/state.py
+++ b/glue/viewers/matplotlib/state.py
@@ -34,7 +34,7 @@ class DeferredDrawSelectionCallbackProperty(SelectionCallbackProperty):
 
 VALID_WEIGHTS = ['light', 'normal', 'medium', 'semibold', 'bold', 'heavy', 'black']
 
-VALID_LOCATIONS = ['best',
+VALID_LOCATIONS = ['best (draggable)', 'best',
                    'upper right', 'upper left',
                    'lower left', 'lower right',
                    'center left', 'center right',
@@ -75,9 +75,10 @@ class MatplotlibDataViewerState(ViewerState):
     show_legend = DeferredDrawCallbackProperty(False, docstring="Whether to show the legend")
     legend_location = DeferredDrawSelectionCallbackProperty(0, docstring="The location of the legend in the axis")
     legend_alpha = DeferredDrawCallbackProperty(0.8, docstring='Transparency of the legend frame')
-    legend_title = DeferredDrawCallbackProperty("", docstring='Transparency of the legend frame')
-    legend_fontsize = DeferredDrawCallbackProperty(10, docstring='Transparency of the legend frame')
-    legend_frame_color = DeferredDrawCallbackProperty("#FFFFFF", docstring='Background color of the legend')
+    legend_title = DeferredDrawCallbackProperty("", docstring='The title of the legend')
+    legend_fontsize = DeferredDrawCallbackProperty(10, docstring='The font size of the title')
+    legend_frame_color = DeferredDrawCallbackProperty("#ffffff", docstring='Frame color of the legend')
+    legend_show_frame_edge = DeferredDrawCallbackProperty(True, docstring="Whether to show the edge of the frame ")
     legend_text_color = DeferredDrawCallbackProperty("#000000", docstring='Text color of the legend')
 
     def __init__(self, *args, **kwargs):
@@ -205,6 +206,7 @@ class MatplotlibDataViewerState(ViewerState):
         self.legend_title = state.legend_title
         self.legend_fontsize = state.legend_fontsize
         self.legend_frame_color = state.legend_frame_color
+        self.legend_show_edge = state.legend_show_edge
         self.legend_text_color = state.legend_text_color
 
     @defer_draw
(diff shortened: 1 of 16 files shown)


specmicp:
It mostly works and is reasonable, except for the edge of the frame.

There are several options:

1. The edge color is the foreground color
2. No edge
3.The edge color can be freely modified by the user
4. inherit from the axes

I would tend to prefer 4, but it is not really set in the settings either. Maybe it should ?

The default is pretty good on black on white, but not that great for any other... It requires a bit of playing around

specmicp:
In this commit, I also set the legend to be draggable so it can be drag around so the user can do whatever he wants. That's quite useful (at least for me), but it might mess up with other redraw... @astrofrog  what do you think ?

astrofrog:
@specmicp - thanks for fixing this! Just to make sure I understand, what is the difference between options 1 and 4? The axes frame color is set to the foreground color anyway, right?

astrofrog:
The dragging is very nice by the way! I guess at the end of the day it depends if it causes all drawing-related tests to fail, but so far from playing around with it it works nicely. If it's a problem for redraws, you could allow location to be set to 'draggable' or something like that, so that it's opt-in?

astrofrog:
One small issue - I think maybe we should call it 'box' color not 'frame' color, otherwise it might be confusing:

Screenshot from 2020-05-21 11-16-15

astrofrog:
Thinking about the actual frame color a bit more, I think maybe having it be the same as the text (option 1) for now might make sense, and in future we could consider having separate global options for the text and frame colors for both axes and legend? (I'm not convinced this is needed)

However, it might be nice to have a toggle for turning the frame on/off on the legend?

astrofrog:
Sorry for all the comments, but I'm also wondering whether as we add more of these properties for the legend it might be worth having a dedicated State class for the legend, so that then options would be set with ``viewer.state.legend.frame_color`` instead of ``viewer.state.legend_frame_color``. It might be tidier to do this - what do you think? (could be done as a separate PR though!)

specmicp:
@ astrofrog 

> Just to make sure I understand, what is the difference between options 1 and 4? The axes frame color is set to the foreground color anyway, right?
> One small issue - I think maybe we should call it 'box' color not 'frame' color, otherwise it might be confusing:

frame is the property name for the box in the matplotlib API, that's why I went with that name, but I can see now how it can be confusing.
I changed it in the glue interface, not sure yet if I should change it in the code

I also added the option to draw or not the edge, and set up the edge color to the font color. We can play with it a bit before deciding. I agree that it does not make sense to allow full customization of the legend, because it's going to clutter the options for sure.

specmicp:
> The dragging is very nice by the way! I guess at the end of the day it depends if it causes all drawing-related tests to fail, but so far from playing around with it it works nicely. If it's a problem for redraws, you could allow location to be set to 'draggable' or something like that, so that it's opt-in?

I saw some strange things when resizing the plot, so I added the drag-behavior as an option. It's also clearer that it's a feature also !

specmicp:
> Sorry for all the comments, but I'm also wondering whether as we add more of these properties for the legend it might be worth having a dedicated State class for the legend, so that then options would be set with `viewer.state.legend.frame_color` instead of `viewer.state.legend_frame_color`. It might be tidier to do this - what do you think? (could be done as a separate PR though!)

I though of the same idea, so I think it's a good thing indeed.

glue/viewers/matplotlib/state.py
```diff
@@ -34,13 +37,69 @@ def notify(self, *args, **kwargs):
...
                    'center left', 'center right',
                    'lower center', 'upper center']
 
 
+class MatplotlibLegendState(State):
+    """The legend state"""
+
+    visible = DeferredDrawCallbackProperty(False, docstring="Whether to show the legend")
+
+    loc_and_drag = DeferredDrawSelectionCallbackProperty(0, docstring="The location of the legend in the axis")
```
astrofrog, line 52: I think this would be clearer as just ``location`` as it was before - then the option can be ``draggable`` when it is draggable.

glue/viewers/matplotlib/state.py
```diff
@@ -34,13 +37,69 @@ def notify(self, *args, **kwargs):
 
 VALID_WEIGHTS = ['light', 'normal', 'medium', 'semibold', 'bold', 'heavy', 'black']
 
-VALID_LOCATIONS = ['best',
+VALID_LOCATIONS = ['best (draggable)', 'best',
```
astrofrog, line 40: What about just ``'draggable'`` for the first option?

glue/viewers/matplotlib/state.py
specmicp: yes,

I changed back that name.
To have the argument to pass to matplotlib API, the property 'mpl_location' need to be used now

glue/viewers/matplotlib/state.py
specmicp: Much better indeed !

specmicp:
@astrofrog I change according to your request, and corrected some bugs. So I think this PR is ready to be merged for more testing.

I will do a new PR soon, with more tests and the documentation of the API, if no big problem is discovered soon.

PR #2147, opened 2020-05-21
Improve efficiency of compute_statistic by minimizing data access

astrofrog:
This improves the efficiency of compute_statistic, especially in the context of the profile viewer, when subsets are applied, by first finding the minimal bounding box for the selection and then extracting data using this bounding box only. In simple tests, this can improve performance by 30x or more. This works especially well when loading CASA datasets since those are very sensitive to disk access.

This needs tests and a changelog entry, and the runtime errors need to be addressed.

cc @keflavich

### DIFF
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -1603,6 +1603,11 @@ class Data(BaseCartesianData):
 
         # TODO: generalize chunking to more types of axis
 
+        # In recent version of Numpy, using lists is not the same as using
+        # tuples, so we make sure we always use tuples to avoid confusion.
+        if isinstance(view, list):
+            view = tuple(view)
+
         if (view is None and
                 isinstance(axis, tuple) and
                 len(axis) > 0 and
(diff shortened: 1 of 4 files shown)


glue/core/data.py
```diff
@@ -1661,8 +1661,53 @@ def compute_statistic(self, statistic, cid, subset_state=None, axis=None,
...
             else:
                 mask = subset_state.to_mask(self, view)
+
                 if np.any(unbroadcast(mask)):
+
+                    # Find minimal subarray containing the masked area.
+                    subarray_slices = []
+                    for idim in range(mask.ndim):
+                        collapse_axes = tuple(index for index in range(mask.ndim) if index != idim)
+                        valid = mask.any(axis=collapse_axes)
```
keflavich, line 1671: This step can be arbitrarily expensive, no?  It will loop over all of the mask data?  Or is mask already sliced down at a previous step?

astrofrog: We already access np.any on the mask above so this won't change much, and we typically operate in chunks for big arrays.

astrofrog:
When we use the subcube approach, we need to pad out the result so the shape matches the result without optimization. I've added a regression test for this, but still need to push up a fix once I have time.

Issue #2148, opened 2020-05-25
Can't visualize subset in histogram


keflavich:
I've created some subsets on a CASA-produced "alpha" map.  I tried making a histogram, and got the message that I can't visualize the subset because it's made from something undefined for the data set.

This screenshot shows the error:
![image](https://user-images.githubusercontent.com/143715/82816302-e7257300-9e68-11ea-88e9-7d7d94ce4796.png)

keflavich:
This was loaded using "gs9"

PR #2149, opened 2020-05-25
Fix histogram viewer with dask arrays

astrofrog:
Fixes https://github.com/glue-viz/glue/issues/2148

@keflavich - can you confirm this works for you?


keflavich:
Yep, that fixed it.

PR #2150, opened 2020-05-28
Use QApplication

nilswagner:
# Pull Request Template

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

### DIFF
--- a/glue/app/qt/splash_screen.py
+++ b/glue/app/qt/splash_screen.py
@@ -29,7 +29,7 @@ class QtSplashScreen(QtWidgets.QWidget):
 
     def set_progress(self, value):
         self.progress.setValue(value)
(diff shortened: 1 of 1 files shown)


nilswagner:
Use QApplication instead of qapp
    def set_progress(self, value):
        self.progress.setValue(value)
        QtWidgets.QApplication.processEvents()  # update progress bar

astrofrog:
@nilswagner - thanks for the PR! Just to check, what prompted you to fix this? Does it fail with recent PyQt versions?

astrofrog:
I see now that this fixes issues with PySide, so I'll go ahead and merge it. Thanks!

PR #2158, opened 2020-06-18
New icons for x- and y-axis independent pan/zoom

rosteen:
Adds two new icons for buttons for single-axis pan/zoom functionality.


maartenbreddels:
Happy to merge this, but maybe also worth considering using the icons from:
https://materialdesignicons.com/icon/arrow-left-right-bold

These are all available from ipyvuetify, e.g. this should work:
```
import ipyvuetify as v
v.Icon(children=['mdi-arrow-left-right-bold'], color='#2d6683')
```

rosteen:
Our UI/UX person is working on a set of updated icons that (I think) use materialdesign. I made these to have something to use that looks consistent with the existing glue icons, to avoid features being blocked while waiting for the replacement set.

maartenbreddels:
Ok, good to hear. the CI failures should be unrelated, I'll let @astrofrog look at that.

astrofrog:
The CI failures are indeed unrelated, let's merge this as-is for now

PR #2164, opened 2020-07-25
Add NDData support as data factories to glue

kakirastern:
## Description
Add `astropy.nddata` support as data factories in order to offer an alternative means to load more specific data types, such as FITS files, but using the `NDData`, making the manipulation of `units`, `mask`, `uncertainty`, and `meta` of any FITS data more manageable in the context of `glue`.


astrofrog:
More generally, I wonder if we should just be doing ``NDData.read`` here and making that work properly for FITS in astropy. Alternatively, glue-solar could implement factories that delegate to ``<some NDData subclass>.read``

kakirastern:
Sure, let me make the changes soon

PR #2156, opened 2020-06-14
Update 1D Profile Viewer to use wcsaxes for plotting 1D spectrum of data cubes

kakirastern:
## Description

Tasks accomplished in this PR:
1. Update the existing `glue`'s "1D Profile" viewer to use Astropy's `wcsaxes` for plotting. 
2. Add a `slice` function option to the "1D Profile" to enable plotting a 1D spectrum of a data cube (e.g. a raster data cube) at a position picked out by `glue-solar`'s pixel extraction tool. 
3. Enable deletion of `IndexedData` type data so that now the dataset in the data collection generated by `glue-solar`'s pixel extraction tool can be deleted.

PR #2167, opened 2020-08-13
Update 1D Profile viewer to use wcsaxes for plotting and add sliders

kakirastern:
## Description
To replace https://github.com/glue-viz/glue/pull/2156. 

Tasks accomplished with this PR:

1. Update the existing glue's 1D Profile viewer to use `astropy`'s `wcsaxes` for plotting.
2. Add a `'slice'` function option to the 1D Profile to enable plotting a 1D spectrum of a data cube (e.g. an IRIS raster data cube) at a position picked out by glue-solar's pixel extraction tool.
3. Add sliders to 1D Profile viewer to update plot.
4. Enable deletion of `BaseData` type data so that now the dataset in the data collection generated by `glue-solar`'s pixel extraction tool can be deleted.


glue/viewers/image/qt/data_viewer.py
```diff
@@ -38,7 +39,9 @@ class ImageViewer(MatplotlibImageMixin, MatplotlibDataViewer):
 
     tools = ['select:rectangle', 'select:xrange',
              'select:yrange', 'select:circle',
-             'select:polygon', 'image:point_selection', 'image:contrast_bias',
+             'select:polygon', 'image:point_selection',
+             'image:pixel_extraction',
```
astrofrog, line 43: If we want to add this by default, we need to make sure the icon is different from point_selection (at the moment they might be the same?)

glue/viewers/matplotlib/state.py
```diff
@@ -273,7 +273,7 @@ class MatplotlibLayerState(LayerState):
 
     def __init__(self, viewer_state=None, **kwargs):
 
-        super(MatplotlibLayerState, self).__init__(viewer_state=viewer_state, **kwargs)
+        super(MatplotlibLayerState, self).__init__(viewer_state=viewer_state, preferred_cmap=None, **kwargs)
```
astrofrog, line 276: Revert?

glue/viewers/profile/layer_artist.py
```diff
@@ -79,6 +79,7 @@ def _calculate_profile_postthread(self):
 
         # Update the data values.
         if len(x) > 0:
+            x = np.arange(len(x))
```
astrofrog, line 82: Revert?

glue/viewers/profile/state.py
```diff
@@ -104,18 +113,9 @@ def _reset_x_limits(self, *event):
...
-            axis_view = [0] * data.ndim
-            axis_view[axis] = slice(None)
-            axis_values = data[self.x_att, tuple(axis_view)]
-            x_min, x_max = np.nanmin(axis_values), np.nanmax(axis_values)
-
         with delay_callback(self, 'x_min', 'x_max'):
-            self.x_min = x_min
-            self.x_max = x_max
+            self.x_min = -0.5
+            self.x_max = data.shape[self.x_att_pixel.axis] - 0.5
```
astrofrog, line 118: Is there a reason for changing the previous behavior here?

glue/core/visual.py
```diff
@@ -107,7 +107,9 @@ def preferred_cmap(self):
     @preferred_cmap.setter
     def preferred_cmap(self, value):
         if isinstance(value, str):
-            self._preferred_cmap = get_cmap(value)
+            for i, element in enumerate(colormaps.members):
+                if element[0] == value:
+                    self._preferred_cmap = element[1]
```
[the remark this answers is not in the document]
kakirastern: Well, if I do so I would get an error thrown since some native `glue` `cmaps` are not in the standard `matplotlib` library. Is there an alternative which would make this acceptable like a `try-except` statement?

glue/viewers/profile/state.py
kakirastern: Yes, it was because I wanted to conform to APE 14's pixel convention: https://docs.astropy.org/en/stable/wcs/wcsapi.html#pixel-conventions-and-definitions

glue/viewers/image/qt/data_viewer.py
kakirastern: The icon for the pixel extraction tool looks similar but not the same as the point selection tool, it was designed like so on purpose to reflect their similar functionality. 

Screenshot 2020-08-18 at 3 32 17 AM

glue/viewers/profile/layer_artist.py
kakirastern: Sure

glue/viewers/matplotlib/state.py
kakirastern: Sure

glue/app/qt/layer_tree_widget.py
```diff
@@ -197,14 +197,14 @@ class DeleteAction(LayerAction):
 
     def _can_trigger(self):
         selection = self.selected_layers()
-        return all(isinstance(s, (core.Data, core.SubsetGroup))
+        return all(isinstance(s, core.BaseData)
```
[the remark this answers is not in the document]
kakirastern: Oh, wait... I think I did this by accident. Should I add the `core.SubsetGroup` back?

glue/viewers/image/qt/data_viewer.py
kakirastern: > The icon added looks identical to https://github.com/glue-viz/glue/blob/master/glue/icons/glue_crosshair.png?

I think they are not really identical, except maybe there are two different pixel extraction tool icons right now with different icons, one from `glue` and the other from `glue-solar` (with the latter to be removed once the former can be upstreamed), so I think you might have been referring to this second icon.

kakirastern:
> There are also changes here that add the pixel extraction to the image viewer by default, which I'm not sure we want to do yet?

Yeah, I could revert the changes made regarding the pixel extraction tool, but I would like to add the image to the `glue` repo so we could use it in `glue-solar`... Would that be okay?

glue/viewers/profile/layer_artist.py
kakirastern: I tried reverting this, but then would get an error that no plot would show in the 1D Profile viewer.

kakirastern: Like so: 

Screenshot 2020-08-18 at 6 31 32 AM

kakirastern:
Removed pixel extraction tool code, but kept the new image for `glue-solar` use.

kakirastern:
@astrofrog PR ready for another review

glue/viewers/profile/layer_artist.py
Cadair: I think I can see why this is needed, as WCSAxes plots against array index, but this feels like the wrong place to put it?

glue/icons/icon_preview.html
```diff
@@ -47,6 +47,7 @@
     <td><img width="60" src='glue_contour.png'></td>
     <td><img width="60" src='glue_crosshair.png'></td>
     <td><img width="60" src='glue_lasso.png'></td>
+    <td><img width="60" src='glue_pixel_extraction.png'></td>
```
Cadair, line 50: anything pixel extraction related shouldn't be in this PR right? That code is still living in glue-solar for now?

Cadair:
I have noticed two major usability issues with this:

1) When you move the sliders the view changes, it would be good if you zoom in to a region of interest and then move the sliders, the view stays where you put it.
2) When the function isn't slice, the sliders shouldn't be visible. This will really confuse people if they are there and useless.

glue/icons/icon_preview.html
kakirastern: Sure, let me remove that from this PR

kakirastern:
So have just updated the `1D Profile` viewer modules to show spectra as profile plots. 

Regarding the comments:

> I have noticed two major usability issues with this:
> 
> 1. When you move the sliders the view changes, it would be good if you zoom in to a region of interest and then move the sliders, the view stays where you put it.
> 2. When the function isn't slice, the sliders shouldn't be visible. This will really confuse people if they are there and useless.

These will involve some deep changes and hence will need some time to accomplish.

kakirastern:
But I reckon we could enhance this PR with regards to https://github.com/glue-viz/glue/pull/2167#issuecomment-677732211 in a separate PR at a later stage.

kakirastern:
Turned out the code I wrote was only working in this case only if one tweaks the opacity slider. (And it was working fine for sequentially stacked IRIS Level 2 datacubes.) And I have just fixed the underlying bugs so it is working fine for the typical FITS files containing data cubes as well. See the following for a demo: 

Screenshot 2020-08-21 at 2 04 58 AM

astrofrog:
This is looking good! I did run into an issue with the controls in the lower left when changing the x axis:

Screenshot 2020-08-24 at 09 30 06

Do you see this too or shall I provide a data file to replicate this?

glue/viewers/image/qt/slice_widget.py
```diff
@@ -57,9 +58,11 @@ def sync_state_from_sliders(self, *args):
                 slices.append(self.viewer_state.slices[i])
         self.viewer_state.slices = tuple(slices)
 
+        if self.viewer_state.reference_data is not self._reference_data:
+            self._reference_data = self.viewer_state.reference_data
```
astrofrog, line 62: Is the if statement really needed? Mabye just write ``self._reference_data = self.viewer_state.reference_data`` directly?

glue/viewers/profile/layer_artist.py
astrofrog: I guess the question is why x needs updating? why is it not correct when returned from visible_data? Changing it here seems dangerous because there's no guarantee there aren't gaps in the data.

glue/viewers/profile/qt/options_widget.py
```diff
@@ -2,17 +2,14 @@
...
 from echo.qt import autoconnect_callbacks_to_qt
 from glue.utils.qt import load_ui, fix_tab_widget_fontsize
+from glue.viewers.profile.qt.slice_widget import ProfileMultiSliceWidgetHelper
+from glue.viewers.matplotlib.state import MatplotlibDataViewerState
 
 __all__ = ['ProfileOptionsWidget']
 
 
-WARNING_TEXT = ("Warning: the coordinate '{label}' is not aligned with pixel "
-                "grid, so the values shown on the x-axis are approximate.")
```
astrofrog, line 13: Why remove this?

kakirastern:
> This is looking good! I did run into an issue with the controls in the lower left when changing the x axis:
> 
> Screenshot 2020-08-24 at 09 30 06
> 
> Do you see this too or shall I provide a data file to replicate this?

Yeah, did you use the "spectrum tool" button? It is a known issue and I have a fix for it which I can implement it soon. I did meant to ask you but didn't get a chance to yet.

glue/viewers/profile/qt/options_widget.py
kakirastern: Actually @Cadair has been saying we don't need it if I remember correctly... I could add this back if that is not the case.

glue/viewers/image/qt/slice_widget.py
kakirastern: Sure

glue/viewers/profile/qt/options_widget.py
Cadair: WCSAxes makes this redundant right? If you plot a coupled axis it now shows both coords.

kakirastern:
> Yeah, did you use the "spectrum tool" button? It is a known issue and I have a fix for it which I can implement it soon. I did meant to ask you but didn't get a chance to yet.

Just patched this error. Actually I was baffled by why it was the case before as @Cadair never saw the extra sliders but I could. It was because he was using the drag-and-drop action to initiate the 1D Profile viewer. To tackle this error from using the spectrum tool button, I have just removed the code snippets that would cause such an issue in the commit https://github.com/glue-viz/glue/pull/2167/commits/da8253640dcbd25d06e0b4a2640cc02ff4606a0b.

glue/viewers/profile/qt/options_widget.py
astrofrog: Ah yes of course!

astrofrog: Well but wait it is still needed when not in slice mode, since you might be e.g. averaging over many slices with slightly different x axes

kakirastern: Okay, let me put it back 👌🏼

glue/viewers/image/qt/profile_viewer_tool.py
```diff
@@ -53,28 +53,3 @@ def activate(self):
...
-            if isinstance(profile_viewer.state.x_att, PixelComponentID):
-                for att in reference_data.pixel_component_ids:
-                    if att is not self.viewer.state.x_att and att is not self.viewer.state.y_att:
-                        if att is not profile_viewer.state.x_att:
-                            profile_viewer.state.x_att = att
-            else:
-                for att in reference_data.world_component_ids:
-                    if att is not self.viewer.state.x_att_world and att is not self.viewer.state.y_att_world:
-                        if att is not profile_viewer.state.x_att:
-                            profile_viewer.state.x_att = att
```
astrofrog, line 80: Why was this removed?

glue/viewers/profile/layer_artist.py
astrofrog: This question is still relevant

kakirastern: Let me check it again more closely

glue/viewers/profile/qt/options_widget.py
astrofrog: Yes just checked, and this warning is still needed in non-slice mode

glue/viewers/profile/qt/slice_widget.py
```diff
@@ -0,0 +1,134 @@
...
+        for i in range(self.data.ndim):
+            if self._sliders[i] is not None:
+                try:
+                    slices.append(self._sliders[i].state.slice_center)
+                except Exception:
+                    pass
+            elif self._sliders[i] is None:
+                slices.append(0)
+
+        self.viewer_state.slices = slices
```
astrofrog, line 134: I think in future we'll want to try and avoid duplication with the image viewer - I can investigate once we add this to the 3-d viewer (to visualize >3-d datasets)

glue/viewers/profile/state.py
```diff
@@ -171,6 +171,33 @@ def _update_priority(self, name):
...
+        return slices[::-1]
+
+    def _set_default_slices(self):
+        # Need to make sure this gets called immediately when reference_data is changed
+        if self.reference_data is None:
+            self.slices = ()
+        else:
+            self.slices = tuple([0] * self.reference_data.ndim)
+
+        print(self.slices)
```
astrofrog, line 199: Remove print statement

glue/viewers/profile/state.py
```diff
@@ -280,7 +316,13 @@ def update_profile(self, update_limits=True):
             data = self.layer
             subset_state = None
 
-        profile_values = data.compute_statistic(self.viewer_state.function, self.attribute, axis=axes, subset_state=subset_state)
+        if self.viewer_state.function == 'slice':
+            data_slice = list(copy.deepcopy(self.viewer_state.slices))
```
astrofrog, line 320: If slices is a list/tuple you can just do ``data_slice = self.viewer_state.slices[:]``

glue/viewers/profile/state.py
```diff
@@ -280,7 +316,13 @@ def update_profile(self, update_limits=True):
...
+            data_slice[pix_cid.axis] = slice(None)
+            profile_values = data.get_data(self.attribute, view=tuple(data_slice))
```
astrofrog, line 322: For subsets you might also need to call ``get_mask`` and then blank out any non-selected values in the profile (see also my comment about subsets in the main review)

glue/viewers/profile/viewer.py
```diff
@@ -1,18 +1,42 @@
...
 
+def get_identity_wcs(naxis):
+
+    wcs = WCS(naxis=naxis)
+    wcs.wcs.ctype = ['X'] * naxis
+    wcs.wcs.crval = [0.] * naxis
+    wcs.wcs.crpix = [1.] * naxis
+    wcs.wcs.cdelt = [1.] * naxis
+
+    return wcs
```
astrofrog, line 18: Avoid duplication with image viewer by moving this to glue.utils? (you can make a new glue.utils.wcs module)

astrofrog (changes requested):
This is almost good to go! A few usability issues:

* Currently the subset profiles don't seem to be right in slice mode, for instance if I make a small subset and push the sliders to the edge of the dataset then the subset profile should be empty/not shown:

Screenshot 2020-09-01 at 13 08 46

* Currently I can't get this to work with 2-d datasets - this used to work for non slice modes but appears broken for all modes now.

* The font size in the profile viewer doesn't seem to be the same as for the image viewer:

Screenshot 2020-09-01 at 13 03 51

I'll review the code in detail below.

glue/viewers/profile/qt/options_widget.py
kakirastern: I have been playing around with the different settings for a bit... Seems like there is a bug with the warning that would bring back the much dreaded extra sliders even when the function is not `slice`, especially when changing between the `x-axis` options. Let me see if there is a way around it.

glue/viewers/image/qt/profile_viewer_tool.py
kakirastern: Will add this back... Think I was having problems with these lines before

glue/viewers/profile/layer_artist.py
kakirastern: Found out why now... It's because the `x_att` and `x_att_pixel` have been mixed up. Am patching this now.

glue/viewers/profile/state.py
kakirastern: Sure

glue/viewers/profile/viewer.py
kakirastern: Sure

glue/viewers/profile/state.py
kakirastern: Sure

glue/viewers/profile/qt/slice_widget.py
kakirastern: Sure

kakirastern:
Have made the changes according to the suggestions given. PR is ready for another review.

glue/viewers/profile/state.py
astrofrog: Just to be clear, ``[:]`` does to a copy for a list/tuple (unlike Numpy arrays). But I'm ok with leaving it as is as it is more explicit.

glue/viewers/profile/layer_artist.py
```diff
@@ -149,15 +149,17 @@ def _update_visual_attributes(self):
...
             return
 
         changed = set() if force else self.pop_changed_properties()
 
-        if force or any(prop in changed for prop in ('layer', 'x_att', 'attribute', 'function', 'normalize', 'v_min', 'v_max', 'visible')):
-            self._calculate_profile(reset=force)
+        if force or any(prop in changed for prop in ('layer', 'slices', 'x_att', 'x_att_pixel', 'attribute',
+                                                     'function', 'normalize', 'v_min', 'v_max', 'visible')):
+            self._update_visual_attributes()
+            self._calculate_profile(reset=True)
```
astrofrog, line 162: For other viewers the approach is instead to do:

```
self._calculate_profile(reset=True)
force=True
```

which ensures the visual properties are updated next. Maybe switch to this approach?

glue/viewers/profile/state.py
```diff
@@ -152,12 +152,12 @@ def _reference_data_changed(self, *args):
...
-                    if self._display_world:
-                        self.x_att_helper.world_coord = True
-                        self.x_att = self.reference_data.world_component_ids[0]
-                    else:
-                        self.x_att_helper.world_coord = False
-                        self.x_att = self.reference_data.pixel_component_ids[0]
+                    # if self._display_world:
+                    #     self.x_att_helper.world_coord = True
+                    #     self.x_att = self.reference_data.world_component_ids[0]
+                    # else:
```
astrofrog, line 158: Remove if unused?

glue/viewers/profile/tests/test_state.py
```diff
@@ -41,14 +41,14 @@ def setup_method(self, method):
...
-    def test_basic_world(self):
-        self.viewer_state.x_att = self.data.world_component_ids[0]
-        x, y = self.layer_state.profile
-        assert_allclose(x, [0, 2, 4])
-        assert_allclose(y, [3.5, 11.5, 19.5])
+    # def test_basic_world(self):
+    #     self.viewer_state.x_att = self.data.world_component_ids[0]
+    #     x, y = self.layer_state.profile
+    #     assert_allclose(x, [0, 2, 4])
+    #     assert_allclose(y, [3.5, 11.5, 19.5])
```
astrofrog, line 51: Remove if unused

astrofrog (changes requested):
Looks good to me! I'm now going to be a bit annoying and ask if you could add a test - a lot of the code here isn't covered by the test suite since there isn't a test for the profile viewer in ``slice`` mode. There are other tests in ``glue/viewers/profile/qt/tests`` you could perhaps start from ?

glue/viewers/profile/tests/test_state.py
kakirastern: Sure

glue/viewers/profile/state.py
kakirastern: Yup, was hesitant before because I wasn't sure if I was in the right direction.

glue/viewers/profile/layer_artist.py
kakirastern: Sure

glue/viewers/profile/state.py
kakirastern: Okay... Agree that it is better to be more explicit here.

kakirastern:
> Looks good to me! I'm now going to be a bit annoying and ask if you could add a test - a lot of the code here isn't covered by the test suite since there isn't a test for the profile viewer in `slice` mode. There are other tests in `glue/viewers/profile/qt/tests` you could perhaps start from ?

Yup, sure. I could add some unit tests for the CI.

glue/viewers/profile/layer_artist.py
kakirastern: Tried changing to 

```python
    self._calculate_profile(reset=True)
    self._update_profile(force=True)
```
But error was thrown upon attempting to view via 1D Profile viewer as data cannot be added for some reason...

PR #2170, opened 2020-09-11
Adding polar and other plots

jsub1:
This PR adds the ability to change the projection of the Matplotlib 2D scatter plot. To make sure that selecting ROIs on the various projections, this PR also adds the ability for ROI subset states to apply an arbitrary transform before testing for inclusion in the ROI. There are also lots of tweaks and changes to the scatter code to make sure that things like script export, color-maps, scaling, vectors, error bars, limits, log, tick formatting, etc. behave when using the different projections. A number of tests are added as well. Some future improvements that can be made to this include:

* Curving the theta error bars when using a polar projection.
* Being smarter about units when axes are inherently angular. Currently the code naively assumes that the values of angular axis are in radians. However, if the selected attribute has an attached angular unit, we could convert it before plotting and ROI selection.

### DIFF
--- a/glue/core/roi.py
+++ b/glue/core/roi.py
@@ -49,6 +49,9 @@ def pixel_to_data(axes, x, y):
     xy = np.column_stack((np.asarray(x).ravel(), np.asarray(y).ravel()))
     return axes.transData.inverted().transform(xy)
 
+def pixel_to_axes(axes, x, y):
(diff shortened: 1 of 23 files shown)


astrofrog (approved):
Fantastic work! This works very nicely for me. There are some small improvements I think we can do (including the ones you mention) but for now I think we can merge and release this.

Issue #2183, opened 2021-03-08
Loading .csv data is slow


dstansby:
**Describe the bug**
I have a modestly sized (175MB, ~1 million lines) .csv file, which takes ~ seconds to be loaded by `pandas.read_csv`. Trying to import the same dataset into glue takes ~1 minute. It would be good if there were any performance gains to be had for importing csv data into glue.

**To Reproduce**
Steps to reproduce the behavior such as:
1. File > Import data > Import from file
2. Import a large .csv file

**Details:**

- Operating System: macOS 11.2.2
- Python version (``python --version``): 3.8.5
- Glue version (``glue --version``): 1.0.1
- How you installed glue: pip

**Additional context**

PR #2184, opened 2021-03-08
Prioritise pandas over astropy for reading tables

dstansby:
# Pull Request Template

## Description
As reported in https://github.com/glue-viz/glue/issues/2183, some basic benchmarking I've done shows that `astropy` is an order of magnitude slower reading tables when compared to `pandas`. In my particular case, the difference is 90 seconds vs. 4 seconds.

Given this large difference, I think it makes sense to try and use pandas where possible, and only fallback on astropy when that doesn't work.

### DIFF
--- a/glue/core/data_factories/tables.py
+++ b/glue/core/data_factories/tables.py
@@ -13,7 +13,8 @@ __all__ = ['tabular_data']
 def tabular_data(path, **kwargs):
     from glue.core.data_factories.astropy_table import astropy_tabular_data
     from glue.core.data_factories.pandas import pandas_read_table
-    for fac in [astropy_tabular_data, pandas_read_table]:
+    # Try pandas first, as it is an order of mangitude faster than astropy
+    for fac in [pandas_read_table, astropy_tabular_data]:
         try:
             return fac(path, **kwargs)
         except Exception:


dhomeier:
I would suggest to also track down where the astropy reader is being slowed down. With the default settings it should use at least the compiled C reader for basic or csv formats, which is generally at most a factor of 2-3 slower than pandas – with the optimised parser (`fast_reader=True`) being on par for larger tables.

These are some timing ratios from an ancient [benchmark notebook](https://github.com/dhomeier/astropy-notebooks/tree/master/io/ascii) just updated with astropy 4.3.1 and pandas 1.3.3; for a 1000000 line file respectively:
> fixed floating point
C to Python speed ratio: 3.19 : 1
C with fast_converter to C speed ratio: 2.13 : 1
Pandas to C with fast_converter speed ratio: 1.23 : 1
>
> exponential
C to Python speed ratio: 2.14 : 1
C with fast_converter to C speed ratio: 3.12 : 1
Pandas to C with fast_converter speed ratio: 1.04 : 1
>
> integers
C to Python speed ratio: 7.36 : 1
Pandas to C speed ratio: 1.29 : 1

Such a large performance lag as you observed would indicate that `io.ascii` is falling back at least to the pure Python reader, or possibly using the 'parallel' option, which is badly broken on many platforms.

image
image
image

dstansby:
Thanks for the reply - I'll try and file a performance issue upstream in astropy.

PR #2187, opened 2021-03-31
Fix FutureWarning in compute_statistic

rosteen:
# Pull Request Template

## Description

Resolves a `FutureWarning` thrown in `compute_statistic` (copied below):

```
/Users/pogle/miniconda3/envs/jdaviz1.1/lib/python3.9/site-packages/glue/core/data.py:1803: FutureWarning: 
Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of 
`arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result 
either in an error or a different result.
  full_result[result_slices] = result
```


astrofrog:
Thanks! I'm sorry I overlooked this PR and fixed this in https://github.com/glue-viz/glue/pull/2196

PR #2192, opened 2021-05-13
EXP: show how one would always create a new subset

astrofrog:
**This should not be merged as-is, it is just a proof of concept for discussion**


pllim:
Thanks! Couldn't have been able to figure this out myself.

I _think_ this is what the POs want... but I am not sure how they want this customizable... We can discuss more at spacetelescope/jdaviz#518 .

pllim:
I think this can be closed without merge.

PR #2199, opened 2021-06-04
Change default percentile to 95

pllim:
## Description

I want to accomplish spacetelescope/jdaviz#577 but not sure how to change the _default_ without digging all the way back to `glue-core`. Any advise will be appreciated. Thanks!

### DIFF
--- a/glue/core/state_objects.py
+++ b/glue/core/state_objects.py
@@ -316,10 +316,10 @@ class StateAttributeLimitsHelper(StateAttributeCacheHelper):
             return
 
         if use_default_modifiers:
-            percentile = 100
+            percentile = 95
             log = False
         else:
-            percentile = getattr(self, 'percentile', None) or 100
+            percentile = getattr(self, 'percentile', None) or 95
             log = getattr(self, 'log', None) or False
 
         if not force and (percentile == 'Custom' or not hasattr(self, 'data') or self.data is None):


astrofrog (commented):
The right way to do it is to subclass this state class here in jdaviz:

https://github.com/glue-viz/glue-jupyter/blob/8e3bb27fda607980317978026fc7a51d0aeb5afa/glue_jupyter/bqplot/image/state.py#L11

and redefine ``percentile`` with a default of 95:

https://github.com/glue-viz/glue/blob/d7dedb5f3e9f0f7f5349811d5d503c0164ceae01/glue/viewers/image/state.py#L492

which can be done by putting the default as the first argument in the CallbackProperty:

```
DDSCProperty(95, docstring
```

You then need to also subclass:

https://github.com/glue-viz/glue-jupyter/blob/8e3bb27fda607980317978026fc7a51d0aeb5afa/glue_jupyter/bqplot/image/layer_artist.py#L16

with the new state subclass for ``_layer_state_cls`` and redefine:

https://github.com/glue-viz/glue-jupyter/blob/8e3bb27fda607980317978026fc7a51d0aeb5afa/glue_jupyter/bqplot/image/viewer.py#L28

with the new subclass in the jdaviz image viewer. If you prefer, I could also do a PR to jdaviz, so just let me know.

PR #2225, opened 2021-08-12
Expose delay link manager

astrofrog:
This can be useful for other tools such as jdaviz

### DIFF
--- a/glue/core/data_collection.py
+++ b/glue/core/data_collection.py
@@ -137,6 +137,15 @@ class DataCollection(HubListener):
 
     @contextmanager
     def delay_link_manager_update(self):
+        """
+        Context manager to delay any updates to the link manager until the
+        context is exited.
+
+        This can be useful for improving performance if e.g. several datasets
+        or links are being added to the data collection, since otherwise the
+        link manaager updates its internal tree representation of the links
+        after each operation.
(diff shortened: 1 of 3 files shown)


glue/core/data_collection.py
```diff
@@ -126,15 +126,22 @@ def _sync_link_manager(self):
...
 
     @contextmanager
-    def _no_sync_link_manager(self):
+    def _ignore_link_manager_update(self):
         self._disable_sync_link_manager = True
         yield
         self._disable_sync_link_manager = False
 
+    @contextmanager
+    def delay_link_manager_update(self):
```
pllim, line 139: There is no docstring. Will this show up in API doc?

pllim (commented):
Maybe another PR, but I think Glue could use a "performance tips and tricks" section like `astropy`. There are some things we could have done to speed things up but they are not intuitive.

glue/core/data_collection.py
astrofrog: Added

pllim (approved):
LGTM!

pllim (commented):
p.s. Bonus would be an example code snippet showing the usage... but it does not block the merge of this PR.

Issue #2223, opened 2021-08-10
Fix reset limits for profile viewer


astrofrog:
Currently for the profile viewer (and by extension the specviz viewer in jdaviz) ``reset_limits`` does not do anything useful for the y axis when not in normalized mode. We should make sure that this works correctly.

pllim:
spacetelescope/jdaviz#335

pllim:
We can close this?  #2233

PR #2233, opened 2021-09-16
Fixes to profile viewer

astrofrog:
This fixes three issues with the profile viewer:

* The home button now correctly resets the y limits
* The visibility checkbox for profile viewer layers now works correctly
* The limits of the profile viewer are now no longer reset every time e.g. a subset is updated or a dataset is added - instead they are reset only when requested by the user or if the reference data changes (as for the image viewer)

### DIFF
--- a/glue/viewers/profile/qt/tests/test_data_viewer.py
+++ b/glue/viewers/profile/qt/tests/test_data_viewer.py
@@ -279,3 +279,42 @@ class TestProfileViewer(object):
         assert not viewer4.state.layers[2].visible
 
         ga.close()
+
+    def test_reset_limits(self):
+        self.viewer.add_data(self.data)
+        self.viewer.add_data(self.data2)
+        self.viewer.state.x_min = 0.2
+        self.viewer.state.x_max = 0.4
+        self.viewer.state.y_min = 0.3
+        self.viewer.state.y_max = 0.5
+        self.viewer.state.reset_limits()
+        assert self.viewer.state.x_min == 0
+        assert self.viewer.state.x_max == 4
+        assert self.viewer.state.y_min == 7
+        assert self.viewer.state.y_max == 23
+
+    def test_limits_unchanged(self):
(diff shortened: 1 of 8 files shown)


pllim:
This fixes #2223 , right?

glue/viewers/profile/qt/tests/test_data_viewer.py
```diff
@@ -279,3 +279,42 @@ def test_session_back_compat(self, protocol):
...
+        assert self.viewer.state.x_min == 0
+        assert self.viewer.state.x_max == 4
+        assert self.viewer.state.y_min == 7
+        assert self.viewer.state.y_max == 23
+
+    def test_limits_unchanged(self):
+        # Make sure the limits don't change if a subset is created or another
+        # dataset added - they should only change if the reference data is changed
+        self.viewer.add_data(self.data)
+        self.viewer.state.x_min = 0.2
```
pllim, line 300: I wonder if allclose check is good enough. You know floating point and all that.

glue/viewers/profile/state.py
```diff
@@ -135,7 +154,13 @@ def _layers_changed(self, *args):
         self._update_combo_ref_data()
 
     @defer_draw
-    def _reference_data_changed(self, *args):
+    def _reference_data_changed(self, before=None, after=None):
+
+        # A callback event for reference_data is triggered if the choices change
+        # but the actual selection doesn't - so we avoid resetting the WCS in
+        # this case.
+        if before is after:
```
pllim, line 162: Is this how multiverse is created? 😆

glue/viewers/profile/qt/tests/test_data_viewer.py
astrofrog: I think this is fine as it should literally return the exact same values, there is no math involved :)

glue/viewers/profile/state.py
astrofrog: Let's see!

PR #2235, opened 2021-09-25
Add ability to rotate ROIs and ROISubsetState

dhomeier:
## Description

WIP for implementing a rotation method on ROI shapes; starting with adding rotation angle as a property to the rectangular and elliptical ROIs.

Comments on implementation details:

- I've added the clockwise (do we have a specific preference for this? Perhaps change to anti-clockwise for consistency with `matplotlib.patches`?) rotation angle as `self.theta`, for now to `RectangularROI`, `EllipticalROI`, `PolygonalROI`, since that was already prepared in the ellipse class. This is understood as an "absolute" angle relative to the initialisation of the instance, so `rotate_to()` and `rotate_by()` allow changing this both as an absolute and additive modification.
- Rotation is always taken around the ROI centre, as rotating around another axis could in principle be effected by a combination of `move_to` and `rotate_to`. But an alternative axis point could still be added as an optional arg later. Again this differs partly from the matplotlib classes using the centre for `Ellipse`, but a corner anchor point for `Rectangle`.
- For `RectangularROI` and `EllipticalROI` the angle is needed as additional property to the size (width, height, semimajor axis etc.), so the latter always retain their `_x` and `_y` assignment even if no longer aligned with the original axes. For the rectangle that makes the meaning of `xmin`, `xmax` etc. rather confusing, but I could not think of a better way to initialise the ROI that is consistent with the current API (it would certainly be better to define it by width, height, centre now).
- As any rotated polygon is still a polygon in all generality, for `PolygonalROI` rotation is directly applied to the vertices. The angle is still stored so one might in principle restore the original orientation, but currently is not roundtripping, as the angle is not used in initialisation. I have made some attempts to add this to `VertexROIBase.__init__()`, but this would add a lot of overhead to the class and does not seem to be worth the effort (I think after adding or removing points there is also no way to restore the "original" polygon).

### DIFF
--- a/glue/core/roi.py
+++ b/glue/core/roi.py
@@ -117,14 +117,6 @@ class Roi(object):  # pragma: no cover
         """Translate the ROI to a center of (x, y)"""
         raise NotImplementedError()
 
-    def rotate_to(self, theta):
-        """Set rotation angle of ROI around center to theta"""
-        raise NotImplementedError()
-
-    def rotate_by(self, dtheta):
-        """Rotate the ROI around center by angle dtheta"""
-        raise NotImplementedError()
-
     def defined(self):
         """ Returns whether or not the subset is properly defined """
         raise NotImplementedError()
@@ -133,6 +125,14 @@ class Roi(object):  # pragma: no cover
         """ Returns a tuple of x and y points, approximating the ROI
         as a polygon."""
         raise NotImplementedError()
+
+    def rotate_to(self, theta):
+        """Set rotation angle of ROI around center to theta"""
+        raise NotImplementedError()
+
+    def rotate_by(self, dtheta):
+        """Rotate the ROI around center by angle dtheta"""
+        self.rotate_to(getattr(self, 'theta', 0.0) + dtheta)
 
     def copy(self):
         """
@@ -231,9 +231,6 @@ class RectangularROI(Roi):
         else:
             self.rotation = rotation(self.theta)
         self.invrot = self.rotation * [[1, -1], [-1, 1]]
-
-    def rotate_by(self, dtheta):
-        self.rotate_to(self.theta + dtheta)
 
     def transpose(self, copy=True):
         if copy:
@@ -681,9 +678,6 @@ class EllipticalROI(Roi):
             self.rotation = rotation(self.theta)
         self.invrot = self.rotation * [[1, -1], [-1, 1]]
 
-    def rotate_by(self, dtheta):
-        self.rotate_to(self.theta + dtheta)
-
     def __gluestate__(self, context):
         return dict(xc=context.do(self.xc),
                     yc=context.do(self.yc),
@@ -843,9 +837,6 @@ class PolygonalROI(VertexROIBase):
             self.vx, self.vy = rotation(dtheta) @ (dx, dy) + np.array(self.center()).reshape(2, 1)
         self.theta = theta
 
-    def rotate_by(self, dtheta):
-        self.rotate_to(self.theta + dtheta)
-
 
 class Projected3dROI(Roi):
     """"A region of interest defined in screen coordinates.
@@ -924,6 +915,9 @@ class Projected3dROI(Roi):
 
     def transformed(self, xfunc=None, yfunc=None):
         return self.roi_2d.transformed(xfunc, yfunc)
+
+    def rotate_to(self, theta):
+        return self.roi_2d.rotate_to(theta)
 
 
 class Path(VertexROIBase):
--- a/glue/core/roi.py
+++ b/glue/core/roi.py
@@ -6,7 +6,7 @@ from matplotlib.transforms import IdentityTransform, blended_transform_factory
 
 from glue.core.component import CategoricalComponent
 from glue.core.exceptions import UndefinedROI
-from glue.utils import points_inside_poly, iterate_chunks
+from glue.utils import points_inside_poly, iterate_chunks, rotation_matrix_2d
 
 
 np.seterr(all='ignore')
@@ -53,12 +53,6 @@ def pixel_to_data(axes, x, y):
 def pixel_to_axes(axes, x, y):
     xy = np.column_stack((np.asarray(x).ravel(), np.asarray(y).ravel()))
     return axes.transAxes.inverted().transform(xy)
-
-
-def rotation(alpha):
-    """Return rotation matrix for angle alpha (increasing anticlockwise) around origin.
-    """
-    return np.array([[np.cos(alpha), -np.sin(alpha)], [np.sin(alpha), np.cos(alpha)]])
 
 
 class Roi(object):  # pragma: no cover
@@ -127,11 +121,11 @@ class Roi(object):  # pragma: no cover
         raise NotImplementedError()
 
     def rotate_to(self, theta):
-        """Set rotation angle of ROI around center to theta"""
+        """Set rotation angle of ROI around center to theta (radian)"""
         raise NotImplementedError()
 
     def rotate_by(self, dtheta):
-        """Rotate the ROI around center by angle dtheta"""
+        """Rotate the ROI around center by angle dtheta (radian)"""
         self.rotate_to(getattr(self, 'theta', 0.0) + dtheta)
 
     def copy(self):
@@ -180,12 +174,12 @@ class RectangularROI(Roi):
 
     Parameters
     ----------
-    xmin, xmax :  float, optional
-        x coordinates of left and right border
-    ymin, ymax :  float, optional
-        y coordinates of lower and upper border
+    xmin, xmax : float, optional
+        x coordinates of left and right border.
+    ymin, ymax : float, optional
+        y coordinates of lower and upper border.
     theta : float, optional
-        Angle of anticlockwise rotation around center
+        Angle of anticlockwise rotation around center in radian.
     """
 
     def __init__(self, xmin=None, xmax=None, ymin=None, ymax=None, theta=None):
@@ -195,11 +189,6 @@ class RectangularROI(Roi):
         self.ymin = ymin
         self.ymax = ymax
         self.theta = 0 if theta is None else theta
-        if np.isclose(self.theta % np.pi, 0.0, atol=1e-9):
-            self.rotation = np.identity(2)
-        else:
-            self.rotation = rotation(self.theta)
-        self.invrot = self.rotation * [[1, -1], [-1, 1]]
 
     def __str__(self):
         if self.defined() and self.theta == 0:
@@ -226,11 +215,6 @@ class RectangularROI(Roi):
     def rotate_to(self, theta):
         """ Rotate anticlockwise around center to position angle theta (radian) """
         self.theta = 0 if theta is None else theta
-        if np.isclose(self.theta % np.pi, 0.0, atol=1e-9):
-            self.rotation = np.identity(2)
-        else:
-            self.rotation = rotation(self.theta)
-        self.invrot = self.rotation * [[1, -1], [-1, 1]]
 
     def transpose(self, copy=True):
         if copy:
@@ -253,16 +237,20 @@ class RectangularROI(Roi):
 
     def contains(self, x, y):
         """
-        Test whether a set of (x,y) points falls within
-        the region of interest
+        Test whether a set of (x,y) points falls within the region of interest
 
-        :param x: A scalar or iterable of x coordinates
-        :param y: A scalar or iterable of y coordinates
+        Parameters
+        ----------
+        x : float or array-like
+            x coordinate(s) of point(s).
+        y : float or array-like
+            y coordinate(s) of point(s).
 
-        *Returns*
+        Returns
+        -------
 
-            A list of True/False values, for whether each (x,y)
-            point falls within the ROI
+        inside : bool or `numpy.ndarray`
+            An iterable of True/False values, for whether each (x,y) point falls within the ROI
         """
 
         if not self.defined():
@@ -289,7 +277,7 @@ class RectangularROI(Roi):
         x = x[keep] - xc
         y = y[keep] - yc
         shape = (2,) + x.shape
-        x, y = (self.invrot @ [x.flatten(), y.flatten()]).reshape(shape)
+        x, y = (rotation_matrix_2d(-self.theta) @ [x.flatten(), y.flatten()]).reshape(shape)
         inside[keep] = (abs(x) <= self.width() / 2) & (abs(y) <= self.height() / 2)
         return inside
 
@@ -323,7 +311,8 @@ class RectangularROI(Roi):
             else:
                 corners = (np.array([-1, 1, 1, -1, -1]) * self.width() / 2,
                            np.array([-1, -1, 1, 1, -1]) * self.height() / 2)
-            return tuple((self.rotation @ corners) + np.array(self.center()).reshape((2, 1)))
+            return tuple((rotation_matrix_2d(self.theta) @ corners) +
+                         np.array(self.center()).reshape((2, 1)))
         else:
             return [], []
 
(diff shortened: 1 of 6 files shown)


dhomeier:
I've changed the sense of rotation to anticlockwise for consistency with matplotlib patches and projections, as well as `astropy.modeling`, erfa etc.

dhomeier:
Marking this as ready for review, as the core functionality should be implemented for all classes where rotation can be sensibly applied.
- `RangeROI` might be up for discussion whether this should be generalised to, or supplemented by, a diagonal range.
- For `Projected3dROI` it is unclear to me if and how spatial rotation _before_ projection to `roi_2d` might be applied.
- For the MplROIs all the selection methods only allow creating/modifying axes-aligned ROIs, so I think all one could do at this point is to support their initialisation with a rotated ROI. I don't see a method to _move_ them with an existing ROI, so this does not seem terribly useful at this point – or are there already Mpl events that could be used for rotation?

glue/core/roi.py
```diff
@@ -55,6 +55,12 @@ def pixel_to_axes(axes, x, y):
     return axes.transAxes.inverted().transform(xy)
 
 
+def rotation(alpha):
+    """Return rotation matrix for angle alpha (increasing anticlockwise) around origin.
+    """
+    return np.array([[np.cos(alpha), -np.sin(alpha)], [np.sin(alpha), np.cos(alpha)]])
```
astrofrog, line 61: This feels like something that belongs in ``glue.utils`` and should probably be called ``rotation_matrix`` or even ``rotation_matrix_2d`` in case we ever add 3-d rotations at some point.

glue/core/roi.py
```diff
@@ -163,21 +177,37 @@ class RectangularROI(Roi):
...
-    def __init__(self, xmin=None, xmax=None, ymin=None, ymax=None):
+    def __init__(self, xmin=None, xmax=None, ymin=None, ymax=None, theta=None):
         super(RectangularROI, self).__init__()
         self.xmin = xmin
         self.xmax = xmax
         self.ymin = ymin
         self.ymax = ymax
+        self.theta = 0 if theta is None else theta
+        if np.isclose(self.theta % np.pi, 0.0, atol=1e-9):
+            self.rotation = np.identity(2)
```
astrofrog, line 199: In general I think it would be good to use ``rotation_matrix`` instead of ``rotation`` to make it clear it's a matrix and not an angle.

glue/core/roi.py
```diff
@@ -163,21 +177,37 @@ class RectangularROI(Roi):
...
 
-    def __init__(self, xmin=None, xmax=None, ymin=None, ymax=None):
+    def __init__(self, xmin=None, xmax=None, ymin=None, ymax=None, theta=None):
         super(RectangularROI, self).__init__()
         self.xmin = xmin
         self.xmax = xmax
         self.ymin = ymin
         self.ymax = ymax
+        self.theta = 0 if theta is None else theta
+        if np.isclose(self.theta % np.pi, 0.0, atol=1e-9):
```
astrofrog, line 198: Could this logic be inside the ``rotation`` function.

glue/core/roi.py
```diff
@@ -163,21 +177,37 @@ class RectangularROI(Roi):
...
+            self.rotation = np.identity(2)
+        else:
+            self.rotation = rotation(self.theta)
+        self.invrot = self.rotation * [[1, -1], [-1, 1]]
```
astrofrog, line 202: Maybe ``inv_rotation_matrix`` and make private as just a convenience? Though this makes it sound like it's the inverse of the matrix when it's actually the rotation in the opposite direction. I think we might want to default to using the anticlockwise rotation in the first place - see my top-level review comment.

glue/core/roi.py
```diff
@@ -193,6 +223,15 @@ def move_to(self, x, y):
...
         self.ymax += dy
 
+    def rotate_to(self, theta):
+        """ Rotate anticlockwise around center to position angle theta (radian) """
+        self.theta = 0 if theta is None else theta
+        if np.isclose(self.theta % np.pi, 0.0, atol=1e-9):
+            self.rotation = np.identity(2)
+        else:
+            self.rotation = rotation(self.theta)
+        self.invrot = self.rotation * [[1, -1], [-1, 1]]
```
astrofrog, line 233: I don't think we should store three different variables that are for the same thing in case they get out of sync. Perhaps we should just store theta and then have the matrices be properties (and it might be premature optimization to worry about caching these)

glue/core/roi.py
```diff
@@ -271,12 +338,14 @@ def __gluestate__(self, context):
...
-                    ymax=context.do(self.ymax))
+                    ymax=context.do(self.ymax),
+                    theta=context.do(self.theta))
 
     @classmethod
     def __setgluestate__(cls, rec, context):
         return cls(xmin=context.object(rec['xmin']), xmax=context.object(rec['xmax']),
-                   ymin=context.object(rec['ymin']), ymax=context.object(rec['ymax']))
+                   ymin=context.object(rec['ymin']), ymax=context.object(rec['ymax']),
+                   theta=context.object(rec['theta']))
```
astrofrog, line 348: Here we have to take into account theta might not exist in old states/session files, and default to zero.

glue/core/roi.py
```diff
@@ -674,10 +814,29 @@ def contains(self, x, y):
...
 
+    # There are several possible definitions of the centre; this is easiest to calculate
+    # (and invariant under rotation?) - do not include starting vertex twice!
+    def center(self):
+        if not self.defined():
+            raise UndefinedROI
+        if self.vx[-1] == self.vx[0] and self.vy[:-1] == self.vy[0]:
+            return np.mean(self.vx[:-1]), np.mean(self.vy[:-1])
+        else:
+            return np.mean(self.vx), np.mean(self.vy)
```
astrofrog, line 825: Sometimes when drawing a region one might end up with a different density of points around the edge, so I wonder if a better definition would be to use the center of the bounding box of the unrotated vertices - this will then be more meaningful once we are able to use the UI to rotate the polygon as this will likely be done by showing the bounding box of the polygon and dragging it around to rotate. This does mean that we should retain a reference to the unrotated vertices, which I think is sensible anyway.

glue/core/roi.py
```diff
@@ -120,6 +126,14 @@ def to_polygon(self):
         as a polygon."""
         raise NotImplementedError()
 
+    def rotate_to(self, theta):
+        """Set rotation angle of ROI around center to theta"""
```
astrofrog, line 130: We should make it clear if we accept degrees or radians.

astrofrog (changes requested):
> I've added the clockwise (do we have a specific preference for this? Perhaps change to anti-clockwise for consistency with matplotlib.patches?) rotation angle as self.theta, for now to RectangularROI, EllipticalROI, PolygonalROI, since that was already prepared in the ellipse class. This is understood as an "absolute" angle relative to the initialisation of the instance, so rotate_to() and rotate_by() allow changing this both as an absolute and additive modification.

I would go with anti-clockwise, it's a pretty standard way of defining it.

> Rotation is always taken around the ROI centre, as rotating around another axis could in principle be effected by a combination of move_to and rotate_to. But an alternative axis point could still be added as an optional arg later. Again this differs partly from the matplotlib classes using the centre for Ellipse, but a corner anchor point for Rectangle.

I would always use the center of the bounding box of the unrotated ROI, as that is consistent with what we might want to do UI wise (this is what I think most vector graphics programs will do).

> For RectangularROI and EllipticalROI the angle is needed as additional property to the size (width, height, semimajor axis etc.), so the latter always retain their _x and _y assignment even if no longer aligned with the original axes. For the rectangle that makes the meaning of xmin, xmax etc. rather confusing, but I could not think of a better way to initialise the ROI that is consistent with the current API (it would certainly be better to define it by width, height, centre now).

What if we actually separated out the rotated and unrotated classes, so we'd have RotatedRectangularROI - and RectangularROI.rotated_by could return a RotatedRectangularROI for instance. This allows us to then be completely flexible with the API of the new classes without being constrained for 'historical reasons'.

> As any rotated polygon is still a polygon in all generality, for PolygonalROI rotation is directly applied to the vertices. The angle is still stored so one might in principle restore the original orientation, but currently is not roundtripping, as the angle is not used in initialisation. I have made some attempts to add this to VertexROIBase.__init__(), but this would add a lot of overhead to the class and does not seem to be worth the effort (I think after adding or removing points there is also no way to restore the "original" polygon).

If we go with the above suggestion of splitting out the classes, one could have RotatedPolygonROI which stores the original vertices and rotation and computes the rotated vertices as needed on the fly - and we could then have a method that would allow one to return an 'evaluated' PolygonROI.

> RangeROI might be up for discussion whether this should be generalised to, or supplemented by, a diagonal range.

Let's not worry about this - I think just adding rotation to the basic 2D shapes is enough.

>For Projected3dROI it is unclear to me if and how spatial rotation before projection to roi_2d might be applied.

As above let's not worry about this.

> For the MplROIs all the selection methods only allow creating/modifying axes-aligned ROIs, so I think all one could do at this point is to support their initialisation with a rotated ROI. I don't see a method to move them with an existing ROI, so this does not seem terribly useful at this point – or are there already Mpl events that could be used for rotation?

I think the right thing to do here will be to start investigating using the widgets in ``matplotlib.widgets`` to replace our own, and rotation is currently being added to them. But that's beyond the scope of this PR.

dhomeier:
Thanks for the review!
> I would go with anti-clockwise, it's a pretty standard way of defining it.

Changed in a0a1e7f
 
> > Rotation is always taken around the ROI centre, as rotating around another axis could in principle be effected by a combination of move_to and rotate_to. But an alternative axis point could still be added as an optional arg later. Again this differs partly from the matplotlib classes using the centre for Ellipse, but a corner anchor point for Rectangle.
> 
> I would always use the center of the bounding box of the unrotated ROI, as that is consistent with what we might want to do UI wise (this is what I think most vector graphics programs will do).

Agreed, at least the `matplotlib.widgets` selectors seem to work that way, too. For polygons it's not quite that obvious (see inline comment).

> > For RectangularROI and EllipticalROI the angle is needed as additional property to the size (width, height, semimajor axis etc.), so the latter always retain their _x and _y assignment even if no longer aligned with the original axes. For the rectangle that makes the meaning of xmin, xmax etc. rather confusing, but I could not think of a better way to initialise the ROI that is consistent with the current API (it would certainly be better to define it by width, height, centre now).
> 
> What if we actually separated out the rotated and unrotated classes, so we'd have RotatedRectangularROI - and RectangularROI.rotated_by could return a RotatedRectangularROI for instance. This allows us to then be completely flexible with the API of the new classes without being constrained for 'historical reasons'.

Not sure if that's really worth creating 3-4 new classes; I've tried to implement the rotations in such a way that they always return the same class (there was a point in case for turning any rotated rectangle into a polygon, but the current implementation should have somewhat better performance).
Personally I would probably prefer to be able to just use the accustomed classes and specify an optional rotation parameter, and there is perhaps a point in having them round-trip properly (quite literally with 180-degree rotations). Although that does not work completely, e.g. a 90-deg rotated ellipse(a, b) is a different object from the ellipse(b, a), and a 180-deg rotated rectangle has its xmin, xmax swapped...
I was rather thinking of accepting `xc, yx, width, height` as _additional_ optional parameters for `RectangularROI` analogously to `EllipticalROI`. Perhaps even to deprecate the `xmin, xmax` syntax at some point, although that could be problematic as they are certainly regularly used as positional arguments (which is however treacherous as well, since `__init__` takes `xmin, xmax, 
ymin, ymax` whereas `update_limits` takes `xmin, ymin, xmax, ymax`).
> 
> > For the MplROIs all the selection methods only allow creating/modifying axes-aligned ROIs, so I think all one could do at this point is to support their initialisation with a rotated ROI. I don't see a method to move them with an existing ROI, so this does not seem terribly useful at this point – or are there already Mpl events that could be used for rotation?
> 
> I think the right thing to do here will be to start investigating using the widgets in `matplotlib.widgets` to replace our own, and rotation is currently being added to them. But that's beyond the scope of this PR.

Yes, I think we should also wait until a decision between matplotlib/matplotlib#19864 and matplotlib/matplotlib#20839 is made.

glue/core/roi.py
dhomeier: The bounding box would at least save the worries if a starting node might be counted with double weight. But the bounding box has the disadvantage that it is not invariant under that rotation (unless we define "bounding box" to be aligned to x, y for the unrotated polygon only, and then rotate along with it). Having the centre change with rotations would make them non-commutative, and tumble the current concept of `rotate_to` and `rotate_by`.
Keeping the unrotated vertices is certainly good point either way, though this also leaves open what to do if vertices have been added or removed in between.

dhomeier: Having `center` depend on the density of vertices admittedly can be somewhat confusing, but it can be sensibly interpreted as a kind of centre of mass for the vertex collection...

glue/core/roi.py
dhomeier: Yes, reviewing how often they are used it's probably not worth caching them at all, certainly not both of them.

glue/core/roi.py
dhomeier: Probably worth a shot, although after only calling `rotation_matrix_2d` as needed, the extra tests for exact multiples of right angles will still be needed in most places, to decide if the whole (much slower) process of pre-selecting and back-rotating points is to be initiated at all.

glue/core/roi.py
dhomeier: Ideally quantities would be accepted as well (defaulting to radian if no unit is present), but units support in Glue at this point seems very limited, so I am not sure if it makes sense to introduce it at one single point.

glue/core/roi.py
dhomeier: Just noted that there is already a `rotation_matrix` in `astropy.coordinates.matrix_utilities`, but it defaults to `u.deg` and the full 3-d case.

dhomeier:
I've moved the rotation matrix to `.utils.geometry` and removed all instances storing its results (even when calling `contains()` many times, the main work will probably go into the actual matrix multiplication with the points).
Leaving the class design as is for the moment.

glue/core/roi.py
dhomeier: > Sometimes when drawing a region one might end up with a different density of points around the edge, so I wonder if a better definition would be to use the center of the bounding box of the unrotated vertices - this will then be more meaningful once we are able to use the UI to rotate the polygon as this will likely be done by showing the bounding box of the polygon and dragging it around to rotate. This does mean that we should retain a reference to the unrotated vertices, which I think is sensible anyway.

This is addressed by using `self.centroid` (see the "mock points" test).
Currently it is not keeping an explicit reference to the original vertices, just to the angle, which makes it possible in principle to restore them by rotating back to `theta=0`.
But I have still made the centre of rotation an optional input for now, since there may be legit reason to rotate around the mean position, bbox centre or even a corner or other point. This will produce confusing/inconsistent results when applying subsequent `rotate_by` operations around different centres though, so needs more discussion if this really is a good idea to allow.

glue/core/roi.py
dhomeier: Moved from `__init__` to `__contains__`, which is the caller for `rotation_matrix_2d`.

glue/core/roi.py
dhomeier: Removed as properties

glue/core/roi.py
dhomeier: There is one issue with using this on degenerate (1D) polygons, where `self.centroid()` returns [NaN, NaN] (mathematically correct, I guess, since it relates to the polygon area. Not sure if that should be considered;
could simply set it to the mean for those cases. That might actually be consistent enough.

astrofrog (approved):
I think this looks good generally. I've noticed a couple of issues but I will leave it up to you whether to deal with it here or in a follow-up PR:

* If I update an ROI attached to a subset state with:

```
s.subset_state.roi.rotate_by(1)
```

then the viewers don't update immediately, instead I need to e.g. pan/zoom to force a refresh. This is actually a pre-existing issue because it happens with ``move_to``. But ideally we should figure out whether we can somehow make sure a ``SubsetUpdateMessage`` is emitted for both translation and rotation.

* If I click and drag on an existing ROI, the selector (in yellow) is not aligned with the ROI (red) - but I think this happens on ``main`` already for me. However, if you could check that click and drag works for you with rotation enabled then that would be good.

Screen Recording 2022-06-14 at 14 10 34


Because neither of these are necessarily new issues, I am approving this - feel free to merge once you think it is good to go.

PR #2236, opened 2021-09-28
fix: deprecated use of imp and numpy

maartenbreddels:


maartenbreddels:
I think failures are unrelated.

dhomeier:
All the pyside tests have been failing in every single PR for weeks at least, so certainly unrelated.
docs-pyqt*[macos] seem to be stuck just now.

Issue #466, opened 2014-10-04
Migrate to numpydoc docstrings


astrofrog:
@ChrisBeaumont - When working on some other PRs, I was going to clean up a few docstrings to change the

```
        :param x: x pixel location in slice array
        :param y: y pixel location in slice array
```

syntax to the numpydoc syntax but I wanted to first check if that should be the goal in the long term, or whether you actually want everything to be in the syntax above? I have a preference for numpydoc for consistency with numpy/scipy/matplotlib/astropy but let me know what you think.

astrofrog:
I'm tagging this as effort-low since anyone can work on it for just a few files.

dhomeier:
Wow, this has been sitting around for a while!
For a heads up. both Numpy and Astropy docs now refer to https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard.

PR #2237, opened 2021-10-01
Update docstrings to numpydoc format

dhomeier:
## Description

This is beginning to address #466; I have tried to clarify or extend the docs where they seemed ambiguous or incomplete, but someone more familiar with the functionality should review the added `Parameters`, `Returns` etc. entries.

Note that #2235 includes updates for `roi.py` since a lot of changes and extensions to the docstrings were already added there.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -37,8 +37,8 @@ intersphinx_cache_limit = 10     # days to keep the cached inventories
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3.7', None),
     'matplotlib': ('https://matplotlib.org', None),
-    'numpy': ('https://docs.scipy.org/doc/numpy', None),
-    'astropy': ('http://docs.astropy.org/en/stable/', None),
+    'numpy': ('https://numpy.org/doc/stable/', None),
+    'astropy': ('https://docs.astropy.org/en/stable/', None),
     'echo': ('https://echo.readthedocs.io/en/latest/', None),
     'pandas': ('https://pandas.pydata.org/pandas-docs/stable/', None),
 }
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -41,6 +41,7 @@ intersphinx_mapping = {
     'astropy': ('https://docs.astropy.org/en/stable/', None),
     'echo': ('https://echo.readthedocs.io/en/latest/', None),
     'pandas': ('https://pandas.pydata.org/pandas-docs/stable/', None),
+    'PyQt5': ('https://www.riverbankcomputing.com/static/Docs/PyQt5/', None),
 }
 
 # Add any paths that contain templates here, relative to this directory.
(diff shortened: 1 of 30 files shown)


astrofrog:
@dhomeier - could you rebase? Sorry for not reviewing this sonner!

dhomeier:
No problem! I did a cursory search for new function's docstrings added since, or any that I've missed before.

dhomeier:
Beats me why the contracted form works in some cases like `~numpy.ndarray` but just breaks the links in others.

astrofrog (changes requested):
Looks good! However the docs failures are related to this PR, so could you investigate? (I think just some incorrect intersphinx references)

dhomeier:
> Looks good! However the docs failures are related to this PR, so could you investigate? (I think just some incorrect intersphinx references)

Can you point me to an example where they are failing? I think the last build here was https://readthedocs.org/projects/glueviz/builds/15921389/

astrofrog:
@dhomeier - ah I meant the Azure build, e.g. the tox ``py37-docs-pyqt513`` config

dhomeier:
Yeah, sorry, just found it – I had missed that among the other pyside + pyqt515 failures!

dhomeier:
> Warning, treated as error:
/home/vsts/work/1/s/.tox/py37-docs-pyqt513/lib/python3.7/site-packages/glue/core/data.py:docstring of glue.core.data.BaseCartesianData.compute_fixed_resolution_buffer:14:py:class reference target not found: glue.core.Data

Not quite an intersphinx issue then – that reference was indeed broken, missing the full class path. Still a mystery to me why the plain `~glue.core.Data` version did not raise (but the links in http://docs.glueviz.org/en/stable/api/glue.core.data.BaseCartesianData.html#glue.core.data.BaseCartesianData.compute_fixed_resolution_buffer are all dead, so maybe numpydoc never looked at them).

astrofrog:
@dhomeier - it looks like there might still be some issues as the Azure docs builds are still failing?

dhomeier:
Yes, I realised I need to `sphinx-build -Wn` as well to track them all down!

dhomeier:
Still a ton of warnings the likes of 
> docstring of glue.viewers.common.qt.toolbar.BasicToolbar.actionTriggered:2: WARNING: Inline emphasis start-string without end-string.

> glue/viewers/common/qt/toolbar.py:docstring of glue.viewers.common.qt.toolbar.BasicToolbar.create:: WARNING: py:class reference target not found: PyQt5.sip.voidptr

As they are pretty much inherited from `PyQt5.QtWidgets.QToolBar` they seem a bit out of control; so I hope docs-test will let them pass through.

astrofrog (approved):
Looks great, thanks! For the reference issues related to Qt I usually add exceptions in conf.py but you could always add that in a separate PR if you like

dhomeier:
Thanks, yes, seems a `('py:class', 'PyQt5.sip.voidptr')` would already have done it, but apparently the Azure builds don't care about that – and fortunately neither throw up the "Inline emphasis start-string without end-string" stuff, which is not a nitpick afaict.

PR #2259, opened 2021-12-20
Fix polar scatter serialization issue

Carifio24:
This PR fixes an issue with saving glue sessions that have polar plots in degree mode. This is accomplished by creating the `RadianTransform` class in `roi_pretransforms.py`, which is glue-serializable. This replaces the previous transform that was applying the conversion to degrees (an instance of `functools.partial`), which glue does not know how to serialize.


astrofrog (changes requested):
Looks good, thanks! Could you add a changelog entry?

Carifio24:
Added the changelog entry

astrofrog (approved):

PR #2265, opened 2022-01-28
BUG: future proof a problematic error-handling block

### LINKED ISSUE #2263: BUG: deprecation warning handling in 1.2.4 is problematic
**Describe the bug**
In testing yt, we treat warnings as errors within pytest jobs.

however with the latest glue-core release we're getting a warning, outside of our control:
```
DeprecationWarning: the load_module() method is deprecated and slated for removal in Python 3.12; use exec_module() instead
```
So the most natural workaround on yt's side would be to ignore specifically this warning, but for some reason I didn't manage to do that (see https://github.com/yt-project/yt/pull/3759)

I think this comes from the fact that glue-core is explicitly using `raise` in he following line, even if the parent exception is really a warning, preventing pytest to effectively ignore it
https://github.com/glue-viz/glue/blob/094e5ae8aff72434c08f0c307f1c7af4af9bea03/glue/config.py#L1009

**To Reproduce**

```toml
#pyproject.toml
filterwarnings = ["error"]
```

```python
# test.py
import glue
```

```
python3.10 -m pytest test.py
```

**Expected behavior**
Ideally, this warning shouldn't be raised at all (follow the recommendation it contains)

**Screenshots**

**Details:**

- Operating System: irrelevant
- Python version (``python --version``): 3.10 (this warning didn't exists on previous versions)
- Glue version (``glue --version``): 1.2.4
- How you installed glue: pip

**Additional context**

neutrinoceros:
Closes #2263

### DIFF
--- a/glue/config.py
+++ b/glue/config.py
@@ -1008,7 +1008,15 @@ def load_configuration(search_path=None):
         except IOError:
             pass
         except Exception as e:
-            raise type(e)("Error loading config file %s:\n%s" % (config_file, e), sys.exc_info()[2])
+            if isinstance(e, Warning):
+                # avoid hard-raising warnings if they were turned into errors
+                # see https://github.com/glue-viz/glue/issues/2263
+                warnings.warn(e, category=type(e))
+            else:
+                raise type(e)(
+                    "Error loading config file %s:\n%s" % (config_file, e),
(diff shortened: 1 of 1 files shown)


astrofrog (approved):
Thanks!

PR #2271, opened 2022-02-04
Update and reorganise sphinx `nitpick_ignore` list

dhomeier:
## Description

Following up on https://github.com/glue-viz/glue/pull/2237#issuecomment-1030030317 this is adding a `PyQt5.sip` entry which raised a warning locally, although it did not seem to affect the CI.
Also cleaned up the entire list a bit, including removing all generic and internal object exceptions – hopefully all further  warnings in our own docs can be fixed rather than ignored.
Apart from all the external `Qt*` references, what remains at this point are some `BasicToolbar` methods that are inherited from `PyQt5.QtWidgets.QToolBar`, and similar ones from matplotlib in `glue.viewers`; and thus also really outside of our own docs.
The module warnings for `organization.rst` appear to be due to no API documentation existing at that level, so why it would be nice to have actual references there, explicitly disabling the links also silences them.

As a note, as of Sphinx 4.1 (seems we have 4.4 everywhere) the entire list of external ignores could also be replaced by a more concise
```python
nitpick_ignore_regex = [('py:class', r'PyQt5\.QtCore\.Q[A-Z][a-zA-Z]+'),
                        ('py:class', r'PyQt5\.QtWidgets\.Q[A-Z][a-zA-Z]+'),
                        ('py:class', r'Qt\.[A-Z][a-zA-Z]+'),
                        ('py:class', r'QPalette\.[A-Z][a-zA-Z]+'),
                        ('py:class', r'QWidget\.[A-Z][a-zA-Z]+'),
                        ('py:class', r'Q[A-Z][a-zA-Z]+')]
```


astrofrog (approved):
Looks good! Feel free to merge once the CI passes.

I'd also be fine with requiring Sphinx 4.1+ if you want to use the regexps for the nitpicks, so I'll leave that up to you (could be done in a follow-up PR).

dhomeier:
> I'd also be fine with requiring Sphinx 4.1+ if you want to use the regexps for the nitpicks, so I'll leave that up to you (could be done in a follow-up PR).

Probably don't need to make it a strict requirement – since older versions will just ignore the unknown config field, if someone wants to build with pre-4.1 they will merely see the bunch of warnings again.

astrofrog:
Sounds good, feel free to merge once ready!

dhomeier:
Thanks for reviewing!

Issue #2269, opened 2022-02-03
Fix CI builds with PyQt 5.15


astrofrog:
In https://github.com/glue-viz/glue/pull/2268 I have disabled the PyQt 5.15 builds, but we should investigate why these aren't working and enable them again.

PR #2272, opened 2022-02-05
Enable PyQt 5.15 tests on macOS + Windows

dhomeier:
## Description

Experiment to investigate the failing tests from #2269; on Linux may be X11 related.
The collection error
```
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.
```
appears to be Linux only.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -48,13 +48,19 @@ jobs:
     # Test a few configurations on MacOS X
     - macos: py37-test-pyqt513
     - macos: py38-test-pyqt514-all
+    # PyQt 5.15 failing on Linux possibly due to X11-related problems (X11/lib/libxcb)
+    # - let's try on macOS
+    - macos: py39-test-pyqt515
     # FIXME: The PySide builds are broken, needs investigating
     # - macos: py37-test-pyside513
 
     # Test a few configurations on Windows
     - windows: py37-test-pyqt510
-    # FIXME: The following fails due to https://github.com/jupyter/qtconsole/issues/400
-    # - windows: py38-test-pyqt514-all
+    # The following failed due to https://github.com/jupyter/qtconsole/issues/400 - should be fixed upstream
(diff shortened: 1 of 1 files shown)


dhomeier:
`macos: py39-test-pyqt515` passing, `windows: py39-test-pyqt515` running past collection; probing a couple more combinations now.

dhomeier:
@astrofrog this looks like an incomplete installation on the Linux side – see https://forum.qt.io/post/666128 .
Is the `PyQt5` package itself broken or can/should we try to ensure installation of the library ourselves
(something like `apt-get install libxcb-xinerama0-dev`; however we tell azure about that)?

astrofrog:
You can install additional libraries on Linux in azure in the following section:

```

libraries:
      apt:
        - libxkbcommon-x11-0
```

I think it's expected that we have to install additional libraries on Linux, although annoying.

dhomeier:
> I think it's expected that we have to install additional libraries on Linux, although annoying.

Yes, I tried to run `python -m tox -e py38-test-pyqt515` manually on a RH system with anaconda, and it installed its PyQt5.15 package without any matching `libQt5Core.so`. pytest did not even get as far as finding `glue-core`...

dhomeier:
OK, must have been another `-dev` package or the `libxcb*` ones.
I am letting the whole set of combinations run once, then to decide which should stay in the standard configuration.

astrofrog (approved):
Looks good! Feel free to merge if this is ready to go

dhomeier:
Which pyqt515 versions should I keep? All 4x3 combinations seems a bit excessive... ;-)

PR #2276, opened 2022-02-11
Fix vector directions when overlaid on image

Carifio24:
When a scatter layer is overlaid on an image, vectors will have the wrong orientation if one (or both) of the axes have different pixel and world coordinate orientations. To fix this issue, this PR looks at the signs of the relevant entry of the cdelt coordinate increments. If they're opposite, it negates the corresponding vector components. The relation between x/y attributes and the wcs coordinate indices is the same one used [here](https://github.com/glue-viz/glue/blob/main/glue/viewers/image/viewer.py#L56).

I would be curious to know if @astrofrog or @catherinezucker (or anyone who's knowledgeable about WCS) has any additional thoughts/suggestions.

### DIFF
--- a/glue/viewers/scatter/layer_artist.py
+++ b/glue/viewers/scatter/layer_artist.py
@@ -277,6 +277,18 @@ class ScatterLayerArtist(MatplotlibLayerArtist):
                 vx = ensure_numerical(self.layer[self.state.vx_att].ravel())
                 vy = ensure_numerical(self.layer[self.state.vy_att].ravel())
 
+                # If this is being overlaid on an image viewer with a WCS,
+                # we need to check the relative orientations of the pixel/world axes
+                if getattr(self.axes, 'wcs', None) is not None:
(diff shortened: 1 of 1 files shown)


astrofrog (commented):
I'm not sure if using CDELT is sufficient, as there might be cases where an image is rotated using CROTA2 or a CD matrix, so then I'm not sure if the logic here still holds.

I'm trying to wrap my head around the use case here - what kind of table columns are you using to define the vector? Are they pixel or world coordinates?

Carifio24:
This PR definitely fell off of my radar. @astrofrog I had used the `get_cdelt()` method since the [astropy docs](https://docs.astropy.org/en/stable/api/astropy.wcs.Wcsprm.html?highlight=get_cdelt#astropy.wcs.Wcsprm.get_cdelt) say that it should work for CD or CROTA as well. I don't have a good set of examples of these on hand, but will certainly try this out if someone has any.

As far as defining the vectors, it looks like numerical components are what are allowed for the vector attributes of a scatter layer.

PR #2277, opened 2022-02-22
Profile viewer: reset y limits when changing collapse function

kecnry:
## Description

This PR resets the y-limits of a profile viewer when the collapse function is changed (similar to what is currently done for normalize - the plotted data have changed in y, so the existing y-limits are meaningless).


astrofrog:
@kecnry - we've fixed some of the CI issues recently so could you rebase to see if the failures here are real?

kecnry:
@astrofrog - rebased but many are still failing.  Is it possible the tests would need updating to account for this scenario?  Or is there a reason I'm missing where this is not the desired behavior?

PR #2279, opened 2022-02-28
Angle units in full-sphere projections

Carifio24:
This PR adds support for using degrees in the four full-sphere projections (Aitoff, Hammer, Mollweide, and Lambert) in the scatter viewer. Tick mark formatting has also been adjusted to match the angle units being used (tick mark locations have been left unchanged). I split the full-sphere projection test (in the Qt scatter viewer tests) into two separate tests, one for each angle unit. 

Since the `RadianTransform` ROI pre-transform originally only applied the transformation to the x component (as it was created with polar plots in mind), I added a `coords` parameter which allows specifying which coordinates are angular, as the full-sphere projections have two angular coordinates. This would cause a problem for older sessions that use a previous version of `RadianTransform`. To work around this, I've added a patch which will set coords to be `['x']` if it's missing (since this would be the appropriate value for polar plots).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@ v1.4.0 (unreleased)
 * Modify profile viewer so that when in 'Sum' mode, parts of profiles
   with no valid values are NaN rather than zero.
 
+* Add support for using degrees in full-sphere projections. [#2279]
+
 v1.3.0 (2022-04-22)
 -------------------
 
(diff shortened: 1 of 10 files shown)


astrofrog (approved):
This looks good but to check the CI failures this should be rebased once https://github.com/glue-viz/glue/pull/2284 is merged.

astrofrog:
@Carifio24 could you add a changelog entry for a new 1.4.0 version in the changelog?

Carifio24:
@astrofrog Just added the changelog entry

astrofrog:
I've rebased to fix a conflict

PR #2280, opened 2022-03-09
Ellipse: Do not throw error for theta=0

pllim:
## Description

`theta=0` is the same as no rotation, so throwing `NotImplementedError` for that case is confusing.


pllim:
Up to you. This patch is not critical for my work. I noticed the error and it didn't make sense to me.

astrofrog:
I'll merge this as no real downside even though theta shouldn't be set in the first place.

Issue #2285, opened 2022-03-29
TableViewer is buggy with incompatible subsets defined


jfoster17:
**Describe the bug**
The TableViewer throws and uncaught IncompatibleAttribute Error when an incompatible subset exists in the data collection and the underlying data set for the TableViewer is toggled to be invisible, as one might do when attempting to view another **valid** subset. 

**To Reproduce**
1. Created two unlinked datasets: d1 and d2
2. Define a subset on d2
3. Create a table viewer for d1
4. Toggle the visibility of d1 in the table viewer to generate the uncaught Error and cause the TableViewer to glitch.

**Expected behavior**
No error should be thrown.

**Details:**

- MacOS 12.3
- Python 3.8.13
- Glue version 1.2.5.dev58+ge8f03f18 (current dev)

**Additional context**
Pull request to fix is forthcoming...

jfoster17:
Fixed in #2286

PR #2286, opened 2022-03-29
Set disabled TableLayerArtists to be not visible

jfoster17:
# Pull Request Template

## Description

A fix for #2285, making sure that disabled TableLayerArtists are not visible, as the viewer will sometimes try to draw visible (but disabled) layers and throw an IncompatibleAttribute Error when it tries to do so. A regression test for this behavior is included.

### DIFF
--- a/glue/tests/helpers.py
+++ b/glue/tests/helpers.py
@@ -61,7 +61,6 @@ H5PY_INSTALLED, requires_h5py = make_skipper('h5py')
 PYQT5_INSTALLED, requires_pyqt5 = make_skipper('PyQt5')
 PYSIDE2_INSTALLED, requires_pyside2 = make_skipper('PySide2')
 
-
 HYPOTHESIS_INSTALLED, requires_hypothesis = make_skipper('hypothesis')
 
 QT_INSTALLED = PYQT5_INSTALLED or PYSIDE2_INSTALLED
--- a/glue/tests/helpers.py
+++ b/glue/tests/helpers.py
@@ -12,9 +12,14 @@ import pytest
 def make_skipper(module, label=None, version=None):
     label = label or module
     try:
-        mod = __import__(module)
+        if label == 'PyQt5':  # PyQt5 does not use __version__
+            from PyQt5 import QtCore
+            version_installed = QtCore.PYQT_VERSION_STR
+        else:
+            mod = __import__(module)
+            version_installed = mod.__version__
         if version:
-            assert LooseVersion(mod.__version__) >= LooseVersion(version)
+            assert LooseVersion(version_installed) >= LooseVersion(version)
         installed = True
     except (ImportError, AssertionError):
         installed = False
@@ -56,6 +61,7 @@ H5PY_INSTALLED, requires_h5py = make_skipper('h5py')
 PYQT5_INSTALLED, requires_pyqt5 = make_skipper('PyQt5')
 PYSIDE2_INSTALLED, requires_pyside2 = make_skipper('PySide2')
 
+
 HYPOTHESIS_INSTALLED, requires_hypothesis = make_skipper('hypothesis')
 
 QT_INSTALLED = PYQT5_INSTALLED or PYSIDE2_INSTALLED
@@ -65,6 +71,12 @@ SPECTRAL_CUBE_INSTALLED, requires_spectral_cube = make_skipper('spectral_cube',
 
 requires_qt = pytest.mark.skipif(str(not QT_INSTALLED),
                                  reason='An installation of Qt is required')
+
+PYQT_GT_59, _ = make_skipper('PyQt5', version='5.9')
+
+REQUIRES_PYQT_GT_59 = PYQT_GT_59 and not PYSIDE2_INSTALLED
+
+requires_pyqt_gt_59 = pytest.mark.skipif(str(not REQUIRES_PYQT_GT_59), reason='Requires PyQt >= 5.9')
 
 
 @contextmanager
--- a/glue/tests/helpers.py
+++ b/glue/tests/helpers.py
@@ -73,9 +73,8 @@ requires_qt = pytest.mark.skipif(str(not QT_INSTALLED),
 
 PYQT_GT_59, _ = make_skipper('PyQt5', version='5.10')
 
-REQUIRES_PYQT_GT_59 = PYQT_GT_59 and not PYSIDE2_INSTALLED
-
-requires_pyqt_gt_59 = pytest.mark.skipif(str(not REQUIRES_PYQT_GT_59), reason='Requires PyQt > 5.9')
+requires_pyqt_gt_59_or_pyside2 = pytest.mark.skipif(str(not PYQT_GT_59 and not PYSIDE2_INSTALLED),
+                                                    reason='Requires PyQt > 5.9 or PySide2')
 
 
 @contextmanager
--- a/glue/tests/helpers.py
+++ b/glue/tests/helpers.py
@@ -72,11 +72,11 @@ SPECTRAL_CUBE_INSTALLED, requires_spectral_cube = make_skipper('spectral_cube',
 requires_qt = pytest.mark.skipif(str(not QT_INSTALLED),
                                  reason='An installation of Qt is required')
 
-PYQT_GT_59, _ = make_skipper('PyQt5', version='5.9')
+PYQT_GT_59, _ = make_skipper('PyQt5', version='5.10')
 
 REQUIRES_PYQT_GT_59 = PYQT_GT_59 and not PYSIDE2_INSTALLED
 
-requires_pyqt_gt_59 = pytest.mark.skipif(str(not REQUIRES_PYQT_GT_59), reason='Requires PyQt >= 5.9')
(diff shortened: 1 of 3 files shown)


glue/viewers/table/qt/tests/test_data_viewer.py
```diff
@@ -460,6 +460,46 @@ def test_incompatible_subset():
...
+    gapp = GlueApplication(dc)
+    viewer = gapp.new_data_viewer(TableViewer)
+    viewer.add_data(d2)
+
+    #This subset should not be shown in the viewer
+    sg1 = dc.new_subset_group('invalid', d1.id['x'] <= 3)
+
+    assert len(viewer.state.layers) == 2
+    assert not viewer.state.layers[1].visible
+    assert viewer.state.layers[0].visible
```
dhomeier, line 484:
```suggestion
    assert len(viewer.layers) == 2
    assert not viewer.layers[1].visible
    assert viewer.layers[0].visible
```
Note that `data_by_row_and_column()`, including your fix, only operates on `viewer.layers` directly, not `viewer.state`, so is this perhaps what you want to test?
Thanks for looking into this!

dhomeier, line 484:
```suggestion
    assert len(viewer.layers) == 2
    assert not viewer.layers[1].visible
    assert viewer.layers[0].visible
```
Note that `data_by_row_and_column()`, including your fix, only operates on `viewer.layers` directly, not `viewer.state`, so is this perhaps what you want to test?
Thanks for looking into this!

jfoster17:
Thanks for your comments @dhomeier -- I don't *think* it matters if you call viewer.state.layers or viewer.layers since they should be kept in sync. I accidentally removed the calls to 

`gapp.show()`
 `process_events()`

when cleaning up this commit. Without this the layers don't try to be drawn and become invalid. With this restored you should be able to see this test failing without my change to `data_by_row_and_column()` and passing with my change.

jfoster17:
But I also did change to call viewer.layers directly...

glue/viewers/table/qt/tests/test_data_viewer.py
```diff
@@ -460,6 +460,49 @@ def test_incompatible_subset():
     assert refresh2.call_count == 0
 
 
+def test_table_incompatible_attribute():
+    """
+    Regression test for a bug where the table viewer generates an
+    uncaught IncompatibleAttribute error if one toggles the
+    visibility of a dataset iff an invalid subset exists. This
```
dhomeier, line 467:
```suggestion
    visibility of a dataset if an invalid subset exists. This
```
spelling nitpick
And I do not quite understand the comment about the "uncaught" error, since that code is actually in the case intercepting said `IncompatibleAttribute` error.

glue/viewers/table/qt/tests/test_data_viewer.py
```diff
@@ -460,6 +460,49 @@ def test_incompatible_subset():
...
+    """
+    app = get_qapp()
+    d1 = Data(x=[1,2,3,4],y=[5,6,7,8],label='d1')
+    d2 = Data(a=['a','b','c'],b=['x','y','z'],label='d2')
+    dc = DataCollection([d1,d2])
+    gapp = GlueApplication(dc)
+    viewer = gapp.new_data_viewer(TableViewer)
+    viewer.add_data(d2)
+
+    #This subset should not be shown in the viewer
```
dhomeier, line 479:
```suggestion
```suggestion
    d1 = Data(x=[1, 2, 3, 4], y=[5, 6, 7, 8], label='d1')
    d2 = Data(a=['a', 'b', 'c'], b=['x', 'y', 'z'], label='d2')
    dc = DataCollection([d1, d2])
    gapp = GlueApplication(dc)
    viewer = gapp.new_data_viewer(TableViewer)
    viewer.add_data(d2)

    # This subset should not be shown in the viewer
```
Here and elsewhere, see coding guidelines in the developer guide, specifically the [style conventions](http://docs.astropy.org/en/stable/development/codeguide.html#coding-style-conventions); those were the E231 and E265 failures in the codestyle test.

dhomeier (changes requested):
I don't have an idea about the failure with `pyqt59-legacy`, but if you cannot find a solution, I'd find it acceptable to mark that as `xfail`.
The `pyqt515-dev` error is unrelated and fixed in #2287.
That leaves mainly the coding style to be addressed, and if you can think of a clearer description that would be a bonus (but I am not very familiar with that part of the code)!

glue/viewers/table/qt/tests/test_data_viewer.py
jfoster17: Yeah, it's a bit more subtle than that. The problem is that `_update_visible()` tries to get a mask for all visible layer_artists if (and only if) the underlying dataset is not visible. I have updated the test description to expand on this point.

glue/viewers/table/qt/tests/test_data_viewer.py
jfoster17: I've resolved the issue with my codestyle setup.

jfoster17:
I do not know the source of the `pyqt59-legacy` error. How do I mark that as xfail for a specific configuration?

dhomeier:
I don't see any existing hooks for PYQT5 versions, so perhaps the best solution would be to define your own `requires_pyqt_gt_59` with `make_skipper` in `helpers.py` analogously to
https://github.com/glue-viz/glue/blob/e8f03f1851ebbb42304b5f96977a81360335d7fa/glue/viewers/histogram/qt/tests/test_data_viewer.py#L641
only this would have to additionally include `not PYSIDE2_INSTALLED`.
In theory the failure might be due to any of the legacy packages in that run, but since this is mostly working on the PyQt interface, that's probably the best bet – and would work either way for skipping the test in the CI.

jfoster17:
Ok -- so I've added a helper to skip this test for PyQt < 5.10, which means that `pyqt59-legacy` runs successfully now and the only failure is due to #2287. Note that I did have to make a more substantial edit to `make_skipper` because PyQt5 does not use `__version__` for a version string.  Happy to wait and see if @astrofrog would prefer another approach.

glue/tests/helpers.py
```diff
@@ -56,6 +61,7 @@ def make_skipper(module, label=None, version=None):
 PYQT5_INSTALLED, requires_pyqt5 = make_skipper('PyQt5')
 PYSIDE2_INSTALLED, requires_pyside2 = make_skipper('PySide2')
 
+
```
dhomeier, line 64:
```suggestion
```
Double spacings only between function/class definitions

glue/tests/helpers.py
```diff
@@ -66,6 +72,12 @@ def make_skipper(module, label=None, version=None):
 requires_qt = pytest.mark.skipif(str(not QT_INSTALLED),
                                  reason='An installation of Qt is required')
 
+PYQT_GT_59, _ = make_skipper('PyQt5', version='5.10')
+
+REQUIRES_PYQT_GT_59 = PYQT_GT_59 and not PYSIDE2_INSTALLED
+
+requires_pyqt_gt_59 = pytest.mark.skipif(str(not REQUIRES_PYQT_GT_59), reason='Requires PyQt > 5.9')
```
dhomeier, line 79:
```suggestion
requires_pyqt_gt_59_or_pyside2 = pytest.mark.skipif(str(not PYQT_GT_59 and not PYSIDE2_INSTALLED),
                                                    reason='Requires PyQt > 5.9 or PySide2')
```
The suggestion probably came around backwards as I was thinking from the `skipif` side... The point anyway was _not_ to skip it for the `pysideXX` jobs either, which hopefully this construct should achieve.

glue/viewers/table/qt/tests/test_data_viewer.py
```diff
@@ -460,6 +461,52 @@ def test_incompatible_subset():
     assert refresh2.call_count == 0
 
 
+@requires_pyqt_gt_59
```
dhomeier, line 464:
```suggestion
@requires_pyqt_gt_59_or_pyside2
```
Getting lengthy, but probably better to stay verbose... ;-)

dhomeier (changes requested):
The `make_skipper` extension should be quite useful in general, thanks. 
For my part this will be good to go once the test is no longer skipped for `*-pyside*`.

glue/tests/helpers.py
jfoster17: Oh, yeah, that makes more sense :)

glue/viewers/table/qt/tests/test_data_viewer.py
```diff
@@ -13,6 +13,7 @@
 from ..data_viewer import DataTableModel, TableViewer
 
 from glue.core.edit_subset_mode import AndNotMode, OrMode, ReplaceMode
+from glue.tests.helpers import requires_pyqt_gt_59
```
dhomeier, line 16:
```suggestion
from glue.tests.helpers import requires_pyqt_gt_59_or_pyside2
```
missed that one

astrofrog (approved):
The test failures are unrelated, so I'll approve and merge now - thanks @jfoster17!

dhomeier (approved):

dhomeier:
Thanks; yes, I haven't seen those pyside14 segfaults before; but everything is passing or skipping the new test now as it should.

Issue #2004, opened 2019-05-13
Faceted subsets don't work with categorical attributes


astrofrog:
```
Traceback (most recent call last):
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/external/echo/qt/connect.py", line 329, in update_prop
    setattr(self._instance, self._prop, value)
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/external/echo/core.py", line 266, in __setattr__
    self._notify_global(**{attribute: value})
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/utils/matplotlib.py", line 165, in wrapper
    return func(*args, **kwargs)
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/viewers/matplotlib/state.py", line 225, in _notify_global
    super(MatplotlibLayerState, self)._notify_global(*args, **kwargs)
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/external/echo/core.py", line 261, in _notify_global
    callback(**kwargs)
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/external/echo/list.py", line 125, in callback
    self.notify(instance, wrapped_list, wrapped_list)
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/external/echo/core.py", line 128, in notify
    cback(new)
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/external/echo/core.py", line 253, in _notify_global_lists
    self._notify_global(**properties)
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/utils/matplotlib.py", line 165, in wrapper
    return func(*args, **kwargs)
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/viewers/matplotlib/state.py", line 182, in _notify_global
    super(MatplotlibDataViewerState, self)._notify_global(*args, **kwargs)
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/external/echo/core.py", line 261, in _notify_global
    callback(**kwargs)
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/utils/matplotlib.py", line 165, in wrapper
    return func(*args, **kwargs)
  File "/Users/tom/Dropbox/Code/Glue/glue/glue/viewers/scatter/layer_artist.py", line 467, in _update_scatter
    self._update_data()
[...]

Issue #1151, opened 2016-10-23
Better faceted subsets for categorical components


astrofrog:
For categorical components, the number of bins should be made to match the number of categories (if that number is reasonable) - similar to histograms.

feldman4:
This would be very useful. Also the subsets can be named after the categories, when possible.

astrofrog:
@feldman4 - indeed. Would you be interested in trying to contribute a fix for this? (no problem if not, but just wanted to check!)

feldman4:
I will take a look and report back asking for guidance. 

There are a few other usability issues and bugs I haven't reported because it's tough for me to understand what's going on without a global picture of the code. Is there up-to-date documentation for this? I found some stuff by googling but I can't tell if it's obsolete.

PR #2288, opened 2022-04-01
Enable categorical subsets

jfoster17:
# Pull Request Template

## Description

This is a work-in-progress towards providing the ability to automatically create subsets for categorical attributes (as suggested in #1151 and #2004). Ultimately, I envision this might expand to allow subsetting on up to two categoricals at once, and for this reason I have added it as a new dialog option instead of modifying the existing facet_subset. Some items that remain:

- [ ] Alert the user if they try to make subsets from a categorical attribute with too many categories
- [ ] Add additional [qualitative](https://matplotlib.org/stable/tutorials/colors/colormaps.html#qualitative) colormaps appropriate for categorical subsets. Currently I just added a single qualitative colormap to the ColormapRegistry, but this is probably not the best solution.
- [ ] Tests
- [ ] (possibly) Expand to allow creating all combinations of up to two categoricals (e.g. for examining combinations experimental factors)

### DIFF
--- a/glue/app/qt/layer_tree_widget.py
+++ b/glue/app/qt/layer_tree_widget.py
@@ -143,15 +143,16 @@ class FacetAction(LayerAction):
         SubsetFacetDialog.facet(self._layer_tree.data_collection,
                                 parent=self._layer_tree, default=default)
 
+
 class SubsetCategoriesAction(LayerAction):
-    
+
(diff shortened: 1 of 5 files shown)


astrofrog:
@jfoster17 - are you still interested in pushing this forward? If so, this will need a rebase - thanks!

PR #2284, opened 2022-03-28
Switch CI to GitHub Actions

dhomeier:
## Description
Transition from Azure pipelines using OpenAstronomy workflows with otherwise mostly unchanged envs; adding the experimental `macos-aarch64` target for `py310`.

PR #2290, opened 2022-04-07
Add publishing workflow to GitHub Actions

dhomeier:
## Description
Follow-up to #2284 adding a `publish_pure_python` job.


dhomeier:
I noticed that one of the last PRs failed with `ValueError` upon teardown with the previous setup; my only explanation is that `excinfo` got lost between the first and second `except`. Strange, but I put it back into a single context here.

astrofrog (approved):
Looks good, thanks!

PR #2292, opened 2022-04-19
Implement as_steps in profile layer state

kecnry:
## Description

This implements a new boolean switch in the profile layer state to toggle displaying the profile as steps vs a line plot.  ~The actual binning logic is adopted from https://github.com/astropy/specutils/blob/main/specutils/spectra/spectral_axis.py#L42.~ (binning logic has now been moved to the bqplot artist in glue-viz/glue-jupyter#309)

### DIFF
--- a/glue/viewers/profile/state.py
+++ b/glue/viewers/profile/state.py
@@ -215,7 +215,7 @@ class ProfileLayerState(MatplotlibLayerState, HubListener):
     percentile = DDSCProperty(docstring='The percentile value used to '
                                         'automatically calculate levels')
 
-    as_steps = DDCProperty(False, docstring='Whether to display the profile as steps')
+    as_steps = DDCProperty(True, docstring='Whether to display the profile as steps')
 
     _viewer_callbacks_set = False
     _layer_subset_updates_subscribed = False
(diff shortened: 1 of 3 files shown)


maartenbreddels:
Nice, I didn't expect it to be as simple, LGTM, also [#309](https://github.com/glue-viz/glue-jupyter/pull/309)

astrofrog:
As mentioned on Slack, I don't think we want to make these changes here - the Matplotlib viewers already draw things as steps by default so it is really the bqplot layer artist in glue-jupyter which should be updated to include this kind of logic.

If we do want to support the option of having both draw styles, then I do support adding a drawstyle or as_steps option to the state, but it should then default to True and simply toggle the drawstyle in the Matplotlib viewer, and again the logic for glue-jupyter should live specifically in the bqplot layer artist (since the Matplotlib layer artist will be taken from glue-core and already draw as steps).

glue/viewers/profile/state.py
```diff
@@ -238,6 +240,7 @@ def __init__(self, layer=None, viewer_state=None, **kwargs):
 
         self.add_callback('layer', self._on_layer_change, priority=1000)
         self.add_callback('visible', self.reset_cache, priority=1000)
+        #self.add_callback('as_steps', self._on_as_steps_change, priority=1000)
```
astrofrog, line 243: Remove if unused?

astrofrog (changes requested):
Looks good! Just a small comment below about removing unused code.

astrofrog:
Can you also fix the codestyle failure (run ``tox -e codestyle`` locally to check) and add a changlog entry?

astrofrog (approved):

PR #2131, opened 2020-05-05
Add a preferred_cmap attribute to visual.py

kakirastern:
## Description

To enable the addition of a `preferred_cmap` attribute so that if this parameter is set a `matplotlib.colors.Colormap` object can be used as the colormap setting of the `2D image`.

PR #2293, opened 2022-04-20
Error when passing in an incorrect string for a colormap for VisualAttributes

nabobalis:
## Description

So I was trying to set a colormap as a string and if the string does not hit either block of logic, `_preferred_cmap` is never set and then it will error during the data loading process.

I was unclear if I should raise an error or just set `_preferred_cmap` as `None` but I decided on `None` and carrying on.

I also added to the gitignore a common(?) vscode plugin directory that is created by that plugin.

### DIFF
--- a/glue/core/tests/test_visual.py
+++ b/glue/core/tests/test_visual.py
@@ -1,20 +1,29 @@
 from glue.core.visual import VisualAttributes
 from matplotlib.cm import get_cmap
-
+import pytest
 
 def test_VA_preferred_cmap():
-    # Not a real CMAP array
-    va = VisualAttributes(preferred_cmap=1)
-    assert va.preferred_cmap == 1
+    # Not a real CMAP array - errors
+    with pytest.raises(ValueError,match="`preferred_cmap` must be a string or an instance of a matplotlib.colors.Colormap"):
+        VisualAttributes(preferred_cmap=1)
 
-    # Not a valid string for a CMAP
-    va = VisualAttributes(preferred_cmap="not_a_cmap")
-    assert va.preferred_cmap is None
+    # Not a valid string for a CMAP - errors
+    with pytest.raises(ValueError,match="not_a_cmap is not a valid colormap name."):
+        VisualAttributes(preferred_cmap="not_a_cmap")
+
+    viridis_cmap = get_cmap("viridis")
 
     # get_cmap cmap name
     va = VisualAttributes(preferred_cmap="viridis")
-    assert va.preferred_cmap == get_cmap("viridis")
-
+    assert va.preferred_cmap == viridis_cmap
     # formal cmap name
     va = VisualAttributes(preferred_cmap="Viridis")
-    assert va.preferred_cmap == get_cmap("viridis")
+    assert va.preferred_cmap == viridis_cmap
+
+    # Valid Colormap
+    va = VisualAttributes(preferred_cmap=viridis_cmap)
+    assert va.preferred_cmap == viridis_cmap
+
+    # None is allowed - it is the default
+    va = VisualAttributes(preferred_cmap=None)
+    assert va.preferred_cmap == None
--- a/glue/core/tests/test_visual.py
+++ b/glue/core/tests/test_visual.py
@@ -2,13 +2,14 @@ from glue.core.visual import VisualAttributes
 from matplotlib.cm import get_cmap
 import pytest
 
+
 def test_VA_preferred_cmap():
     # Not a real CMAP array - errors
-    with pytest.raises(ValueError,match="`preferred_cmap` must be a string or an instance of a matplotlib.colors.Colormap"):
+    with pytest.raises(TypeError, match="`preferred_cmap` must be a string or an instance of a matplotlib.colors.Colormap"):
         VisualAttributes(preferred_cmap=1)
 
     # Not a valid string for a CMAP - errors
-    with pytest.raises(ValueError,match="not_a_cmap is not a valid colormap name."):
+    with pytest.raises(ValueError, match="not_a_cmap is not a valid colormap name."):
         VisualAttributes(preferred_cmap="not_a_cmap")
 
     viridis_cmap = get_cmap("viridis")
@@ -26,4 +27,4 @@ def test_VA_preferred_cmap():
 
     # None is allowed - it is the default
     va = VisualAttributes(preferred_cmap=None)
-    assert va.preferred_cmap == None
+    assert va.preferred_cmap is None
--- a/glue/core/tests/test_visual.py
+++ b/glue/core/tests/test_visual.py
@@ -4,17 +4,17 @@ from matplotlib.cm import get_cmap
 
 def test_VA_preferred_cmap():
     # Not a real CMAP array
-    abc = VisualAttributes(preferred_cmap=1)
-    assert abc.preferred_cmap == 1
+    va = VisualAttributes(preferred_cmap=1)
+    assert va.preferred_cmap == 1
 
     # Not a valid string for a CMAP
-    abc = VisualAttributes(preferred_cmap="not_a_cmap")
-    assert abc.preferred_cmap is None
+    va = VisualAttributes(preferred_cmap="not_a_cmap")
+    assert va.preferred_cmap is None
 
     # get_cmap cmap name
-    abc = VisualAttributes(preferred_cmap="viridis")
-    assert abc.preferred_cmap == get_cmap("viridis")
+    va = VisualAttributes(preferred_cmap="viridis")
+    assert va.preferred_cmap == get_cmap("viridis")
 
     # formal cmap name
-    abc = VisualAttributes(preferred_cmap="Viridis")
-    assert abc.preferred_cmap == get_cmap("viridis")
+    va = VisualAttributes(preferred_cmap="Viridis")
+    assert va.preferred_cmap == get_cmap("viridis")
--- a/glue/core/tests/test_visual.py
+++ b/glue/core/tests/test_visual.py
@@ -1,0 +1,20 @@
+from glue.core.visual import VisualAttributes
+from matplotlib.cm import get_cmap
+
+
+def test_VA_preferred_cmap():
+    # Not a real CMAP array
+    abc = VisualAttributes(preferred_cmap=1)
+    assert abc.preferred_cmap == 1
+
+    # Not a valid string for a CMAP
+    abc = VisualAttributes(preferred_cmap="not_a_cmap")
+    assert abc.preferred_cmap is None
+
+    # get_cmap cmap name
+    abc = VisualAttributes(preferred_cmap="viridis")
+    assert abc.preferred_cmap == get_cmap("viridis")
+
+    # formal cmap name
+    abc = VisualAttributes(preferred_cmap="Viridis")
+    assert abc.preferred_cmap == get_cmap("viridis")
--- a/glue/core/visual.py
+++ b/glue/core/visual.py
@@ -1,4 +1,4 @@
-from matplotlib.colors import ColorConverter
+from matplotlib.colors import ColorConverter, Colormap
 from matplotlib.cm import get_cmap
 
 from glue.config import settings
@@ -128,9 +128,11 @@ class VisualAttributes(HasCallbackProperties):
                         break
                 else:
                     # If the string name fails to be validated
-                    self._preferred_cmap = None
-        else:
+                    raise ValueError(f"{value} is not a valid colormap name.")
+        elif isinstance(value, Colormap) or value is None:
             self._preferred_cmap = value
+        else:
+            raise ValueError("`preferred_cmap` must be a string or an instance of a matplotlib.colors.Colormap")
 
     @callback_property
     def alpha(self):
--- a/glue/core/visual.py
+++ b/glue/core/visual.py
@@ -19,21 +19,21 @@ class VisualAttributes(HasCallbackProperties):
 
     Parameters
     ----------
-    parent : QObject, optional
+    parent : :class:`~PyQt5.QtCore.QObject`, optional
         The object that this visual attributes object is attached to. Default is `None`.
-    color : str, optional
-        A matplotlib color string. Default is `None`.
-    alpha : float, optional
-        Opacity, between 0-1. Default is `None`.
-    preferred_cmap : str, optional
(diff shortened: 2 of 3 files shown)


glue/core/tests/test_visual.py
```diff
@@ -0,0 +1,20 @@
...
+    abc = VisualAttributes(preferred_cmap="not_a_cmap")
+    assert abc.preferred_cmap is None
+
+    # get_cmap cmap name
+    abc = VisualAttributes(preferred_cmap="viridis")
+    assert abc.preferred_cmap == get_cmap("viridis")
+
+    # formal cmap name
+    abc = VisualAttributes(preferred_cmap="Viridis")
+    assert abc.preferred_cmap == get_cmap("viridis")
```
nabobalis, line 20:
```suggestion
    va = VisualAttributes(preferred_cmap=1)
    assert va.preferred_cmap == 1

    # Not a valid string for a CMAP
    va = VisualAttributes(preferred_cmap="not_a_cmap")
    assert va.preferred_cmap is None

    # get_cmap cmap name
    va = VisualAttributes(preferred_cmap="viridis")
    assert va.preferred_cmap == get_cmap("viridis")

    # formal cmap name
    va = VisualAttributes(preferred_cmap="Viridis")
    assert va.preferred_cmap == get_cmap("viridis")
```

astrofrog (approved):
Looks good, thanks!

glue/core/visual.py
```diff
@@ -25,10 +25,11 @@ class VisualAttributes(HasCallbackProperties):
     :param linestyle: The linestyle (``'solid' | 'dashed' | 'dash-dot' | 'dotted' | 'none'``)
     :param marker: The matplotlib marker shape (``'o' | 's' | '^' | etc``)
     :param markersize: The size of the marker (int)
+    :param preferred_cmap: The colormap required (str)
 
     """
 
-    def __init__(self, parent=None, washout=False, color=None, alpha=None, preferred_cmap=None):
+    def __init__(self, parent=None, color=None, alpha=None, preferred_cmap=None):
```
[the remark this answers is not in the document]
dhomeier: Agreeing in principle, but it could break code that was simply passing the args in order – e.g.
`VisualAttributes(parent, False, 'yellow', 0.6)`
will create an instance with invalid `color` and `alpha` attributes after this change.

The `param` listing is slightly misleading in itself since the other parameters are hardcoded in `__init__` and can only be changed by the `setter`s, not on creation. Should probably be clarified as well (or perhaps better, added as `kwargs` accepted by `__init__`?), but that's rather outside the scope of this PR. Also should put this on my list of docstrings yet to be converted to numpydoc format. ;-)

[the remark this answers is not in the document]
nabobalis: > Agreeing in principle, but it could break code that was simply passing the args in order – e.g. `VisualAttributes(parent, False, 'yellow', 0.6)` will create an instance with invalid `color` and `alpha` attributes after this change.

I don't disagree, I don't know a work around to stop this being a breaking change.

> The `param` listing is slightly misleading in itself since the other parameters are hardcoded in `__init__` and can only be changed by the `setter`s, not on creation. Should probably be clarified as well (or perhaps better, added as `kwargs` accepted by `__init__`?), but that's rather outside the scope of this PR. 

I can at least add a ``kwargs`` to the init? 

> Also should put this on my list of docstrings yet to be converted to numpydoc format. ;-)

I can convert it?

glue/core/visual.py
```diff
@@ -25,10 +25,11 @@ class VisualAttributes(HasCallbackProperties):
     :param linestyle: The linestyle (``'solid' | 'dashed' | 'dash-dot' | 'dotted' | 'none'``)
     :param marker: The matplotlib marker shape (``'o' | 's' | '^' | etc``)
     :param markersize: The size of the marker (int)
+    :param preferred_cmap: The colormap required (str)
```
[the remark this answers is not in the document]
nabobalis: I will fix it working.

glue/core/visual.py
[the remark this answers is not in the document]
dhomeier: > I can at least add a `kwargs` to the init?

Adding `linewidth`, `linestyle`, `marker` and `markersize` with the current init values as defaults would make sense IMO.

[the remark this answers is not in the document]
dhomeier: > > Also should put this on my list of docstrings yet to be converted to numpydoc format. ;-)
> 
> I can convert it?

If you want to take this on, you might refer e.g. to https://github.com/glue-viz/glue/pull/2237/files#diff-db5318dc12582d4bf966dfcba587b279c0050d8bdb43a5034719c46e5cffb8b8 for the changes in syntax, thanks!

glue/core/visual.py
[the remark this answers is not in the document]
dhomeier: Looking at its implementation in #2131 and your fixes here, it should work just fine with None passed, so really optional.

glue/core/visual.py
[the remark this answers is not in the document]
nabobalis: This should all be fixed now.

glue/core/visual.py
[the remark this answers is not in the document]
nabobalis: The int here just sets the value, I think it should be a str or a  MPL LinearSegmentedColormap.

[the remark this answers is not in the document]
dhomeier: Yes, misread the code and overlooked the `insinstance(value, str)` conditional – but then what kind of colormap is your `VisualAttributes(preferred_cmap=1)` test example setting?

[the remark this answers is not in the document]
dhomeier: Yes, misread the code and overlooked the `insinstance(value, str)` conditional – but then what kind of colormap is your `VisualAttributes(preferred_cmap=1)` test example setting?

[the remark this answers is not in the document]
nabobalis: The test was just to check it sets the value, as it stands, you could pass it any non-str value and it will just assign it. 

Whether you want that behaviour, I don't know.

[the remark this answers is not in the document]
dhomeier: If it will raise an error later during the data loading as per your original example, I think it should be treated just like an invalid name string. In fact I would rather not silently revert to `None` in such cases, but at least raise a warning – otherwise users may be confused that their inputs are apparently completely ignored. But that may be up for discussion.
In any case this sounds like the only legit inputs here are `mpl.colors.Colormap` instances or valid names thereof.

glue/core/visual.py
```diff
@@ -15,20 +15,29 @@
...
-    :param linewidth: The linewidth (float or int)
-    :param linestyle: The linestyle (``'solid' | 'dashed' | 'dash-dot' | 'dotted' | 'none'``)
-    :param marker: The matplotlib marker shape (``'o' | 's' | '^' | etc``)
-    :param markersize: The size of the marker (int)
-
+    This class is used to define visual attributes for any kind of objects.
+
+    Parameters
+    ----------
+    parent : QObject, optional
```
dhomeier, line 22:
```suggestion
    parent : ~PyQt5.QtCore.QObject, optional
```
I think this should work with intersphinx mapping; worth a try.

glue/core/visual.py
```diff
@@ -15,20 +15,29 @@
...
+        The object that this visual attributes object is attached to. Default is `None`.
+    color : str, optional
+        A matplotlib color string. Default is `None`.
+    alpha : float, optional
+        Opacity, between 0-1. Default is `None`.
```
dhomeier, line 27:
```suggestion
        A matplotlib color string. Default is is set from `~glue.config.SettingsRegistry`.
    alpha : float, optional
        Opacity, between 0-1. Default is is set from `~glue.config.SettingsRegistry`.
```

dhomeier, line 27:
```suggestion
    color : `str`, optional
        A matplotlib color string. Default is is set from `~glue.config.SettingsRegistry`.
    alpha : `float`, optional
        Opacity, between 0-1. Default is is set from `~glue.config.SettingsRegistry`.
```

glue/core/visual.py
```diff
@@ -15,20 +15,29 @@
...
+    preferred_cmap : str, optional
+        The name of the preferred colormap. Default is `None`.
```
dhomeier, line 29:
```suggestion
    preferred_cmap : `str` or :class:`~matplotlib.colors.Colormap`, optional
        A colormap to be used as the preferred colormap, by name or instance. Default is `None`.
```

glue/core/visual.py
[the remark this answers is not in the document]
nabobalis: That makes a lot of sense, would you like me to raise an error if the string is an invalid colormap or not either a string or an instance of `mpl.colors.Colormap`?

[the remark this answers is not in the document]
dhomeier: I tend to raising an exception now, since the intended purpose of setting a different `preferred_cmap` will have failed in that case. Pinging @astrofrog for a second look.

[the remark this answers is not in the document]
nabobalis: I have just changed it for now to raise a ValueError in both cases. If you leave it as None, it will set it as None and carry on.

glue/core/visual.py
```diff
@@ -15,20 +15,29 @@
...
+    This class is used to define visual attributes for any kind of objects.
+
+    Parameters
+    ----------
+    parent : :class:`~PyQt5.QtCore.QObject`, optional
+        The object that this visual attributes object is attached to. Default is `None`.
+    color : `str`, optional
+        A matplotlib color string. Default is is set from `~glue.config.SettingsRegistry`.
+    alpha : `float`, optional
+        Opacity, between 0-1. Default is is set from `~glue.config.SettingsRegistry`.
```
dhomeier, line 27:
```suggestion
        A matplotlib color string. Default is is set from `~glue.config.SettingRegistry`.
    alpha : `float`, optional
        Opacity, between 0-1. Default is is set from `~glue.config.SettingRegistry`.
```
typo

dhomeier:
Thanks! The first two tests will need to be changed to the respective `pytest.raises` now.

nabobalis:
> Thanks! The first two tests will need to be changed to the respective `pytest.raises` now.

I forgot completely! Should be fixed now.

glue/core/visual.py
```diff
@@ -109,13 +118,21 @@ def preferred_cmap(self):
...
-                self._preferred_cmap = get_cmap(value)
-        else:
+                        break
+                else:
+                    # If the string name fails to be validated
+                    raise ValueError(f"{value} is not a valid colormap name.")
+        elif isinstance(value, Colormap) or value is None:
             self._preferred_cmap = value
+        else:
+            raise ValueError("`preferred_cmap` must be a string or an instace of a mpl.colors.Colormap")
```
dhomeier, line 135:
```suggestion
            raise ValueError("`preferred_cmap` must be a string or an instance of a matplotlib.colors.Colormap")
```

nabobalis: Woops!

glue/core/visual.py
```diff
@@ -109,13 +118,21 @@ def preferred_cmap(self):
...
-                self._preferred_cmap = get_cmap(value)
-        else:
+                        break
+                else:
+                    # If the string name fails to be validated
+                    raise ValueError(f"{value} is not a valid colormap name.")
+        elif isinstance(value, Colormap) or value is None:
             self._preferred_cmap = value
+        else:
+            raise ValueError("`preferred_cmap` must be a string or an instance of a matplotlib.colors.Colormap")
```
dhomeier, line 135:
```suggestion
            raise TypeError("`preferred_cmap` must be a string or an instance of a matplotlib.colors.Colormap")
```
this was lost somewhere in the pushes; technical, but quite a clear case of https://docs.python.org/3/library/exceptions.html#TypeError.

nabobalis: Ah I must have not noticed that and forgot to change it.

glue/core/visual.py
dhomeier: has this been reverted again?

nabobalis: Yeah, I am unsure how I did that.

dhomeier: maybe force-pushing from your local branch...

nabobalis: I probably forgot to pull and forced pushed without paying attention.

glue/core/visual.py
```diff
@@ -15,20 +15,29 @@
...
+    This class is used to define visual attributes for any kind of objects.
+
+    Parameters
+    ----------
+    parent : :class:`~PyQt5.api.qtcore.qobject`, optional
+        The object that this visual attributes object is attached to. Default is `None`.
+    color : `str`, optional
+        A matplotlib color string. Default is is set from `~glue.config.SettingRegistry`.
+    alpha : `float`, optional
+        Opacity, between 0-1. Default is is set from `~glue.config.SettingRegistry`.
```
dhomeier, line 27:
```suggestion
        A matplotlib color string. Default is set from :class:`~glue.config.SettingRegistry`.
    alpha : `float`, optional
        Opacity, between 0-1. Default is set from :class:`~glue.config.SettingRegistry`.
```
one last try

dhomeier (approved):
Thanks for the updates, docs are looking great, and with the tests from https://github.com/glue-viz/glue/actions/runs/2260892091 this is good to go for my part!

dhomeier:
@astrofrog just asking for your confirmation because we have now agreed upon raising an exception on invalid inputs rather than just leaving `preferred_cmap` set to `None` as in the version that you approved.

astrofrog:
Looks good to me!

glue/core/visual.py
```diff
@@ -15,20 +15,29 @@
...
-    :param linewidth: The linewidth (float or int)
-    :param linestyle: The linestyle (``'solid' | 'dashed' | 'dash-dot' | 'dotted' | 'none'``)
-    :param marker: The matplotlib marker shape (``'o' | 's' | '^' | etc``)
-    :param markersize: The size of the marker (int)
-
+    This class is used to define visual attributes for any kind of objects.
+
+    Parameters
+    ----------
+    parent : :class:`~PyQt5.api.qtcore.qobject`, optional
```
dhomeier, line 22:
```suggestion
    parent : `QObject`, optional
```
I'll try to get this sorted out in a separate numpydoc PR!

nabobalis:
Thank you @dhomeier and @astrofrog

PR #2294, opened 2022-04-26
Provide interface for doing in-place unit conversions on components

astrofrog:
This is a work in progress to experiment with unit conversions. At the moment this provides an interface to convert the data values of a ``Component`` on-the-fly when accessed. This causes a ``NumericalDataChangedMessage`` message to be broadcast via the hub which then should trigger updates in the viewers (though some viewers I think here and maybe in glue-jupyter don't actually listen for this message - I need to go through and check).

With this, one could do e.g.

```
data.convert_component_units_to('flux', 'mJy')
```

If the original flux of a component was in ``Jy`` and the component will then behave for all intents and purposes as if it was in mJy.

What still needs doing if we decide to go ahead with this:

* [ ] Any kind of error checking to make sure the unit conversion is valid/possible
* [ ] Figuring out how to extend this to derived and coordinate components
* [ ] Dealing with cases where the original units are blank and units are then set (different from conversion)
* [ ] Adding support for equivalencies
* [ ] Thinking about how to deal with the case where an equivalency might need another component, e.g. to convert spectral density one might need the spectral axis depending on the unit conversion
* [ ] Tests and Documentation!

cc @kecnry

### DIFF
--- a/glue/core/component.py
+++ b/glue/core/component.py
@@ -45,7 +45,6 @@ class Component(object):
 
         # The physical units of the data
         self.units = units
-        self._original_units = self._units
 
         # The actual data
         # subclasses may pass non-arrays here as placeholders.
@@ -66,7 +65,13 @@ class Component(object):
         if value is None:
             self._units = ''
         else:
-            self._units = str(value)
+            new_units = str(value)
+        if getattr(self, '_original_units', '') == '':
+            self._original_units = new_units
+        if not u.Unit(self._original_units).is_equivalent(u.Unit(new_units)):
+            raise u.UnitConversionError(f"New unit '{new_units}' must be convertible "
+                                        f"from original one '{self._original_units}'")
+        self._units = new_units
 
     def _convert_to_units(self, units):
         self.units = units
--- a/glue/core/component.py
+++ b/glue/core/component.py
@@ -3,6 +3,8 @@ import logging
 import numpy as np
 import pandas as pd
 
+from astropy import units as u
+
 from glue.core.coordinate_helpers import dependent_axes, pixel2world_single_axis
 from glue.utils import (shape_to_string, coerce_numeric,
                         broadcast_to, categorical_ndarray)
@@ -40,8 +42,10 @@ class Component(object):
     """
 
     def __init__(self, data, units=None):
+
         # The physical units of the data
         self.units = units
+        self._original_units = self._units
 
         # The actual data
         # subclasses may pass non-arrays here as placeholders.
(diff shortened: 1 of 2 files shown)


glue/core/component.py
```diff
@@ -40,8 +42,10 @@ class Component(object):
     """
 
     def __init__(self, data, units=None):
+
         # The physical units of the data
         self.units = units
+        self._original_units = self._units
```
dhomeier, line 48:
```suggestion
```
see comment below about moving this to `setter`

glue/core/component.py
```diff
@@ -64,10 +68,20 @@ def units(self, value):
         else:
             self._units = str(value)
```
dhomeier, line 69:
```suggestion
        else:
            new_units = str(value)
        if getattr(self, '_original_units', '') == '':
            self._original_units = new_units
        if not u.Unit(self._original_units).is_equivalent(u.Unit(new_units)):
            raise u.UnitConversionError(f"New unit '{new_units}' must be convertible "
                                        f"from original one '{self._original_units}'")
        self._units = new_units
```
Doh, that should also include
```python
        if value is None:
            new_units = ''
```
but GH won't let me touch that line...
Anyway, proposing here to allow setting a novel unit type _once_ from a unitless/dimensionless component, and then raising immediately on non-equivalent units so they cannot be set in the first place (otherwise will raise later in `__getitem__` when the damage is already done.

dhomeier (commented):
The test failures highlight the question what to do about incompatible unit assignments; I think handling all of that inside `units.setter` may allow the best control (and along the way resolves the failure about `self._original_units` not being set for `DateTimeComponent`).

Might make it a bit trickier to handle equivalencies, but the best way to solve this is perhaps to always set them from a property as well - makes sense to stick them to the component, I think, as they will normally not change (they still could, making it e.g. as a kwarg to `data.convert_component_units_to()`).

glue/core/component.py
```diff
@@ -62,12 +65,28 @@ def units(self, value):
         if value is None:
             self._units = ''
```
dhomeier, line 66:
```suggestion
            new_units = ''
```

dhomeier:
@astrofrog I have committed my own suggestions as I could not run some of the previously failing `viewer` tests locally (still half-broken PyQt on macOS-arm64...). I suppose you can still force-push from your local branch if you wish to revert them all (and maybe I should open a PR onto this from my own branch?)
The new failure in `test_load_pixel_components_07` looks like it is setting a unit to `'None'` rather than `None`; not sure if that is an idiosyncrasy of that older format, but might make sense to add that check to `units.setter`.

astrofrog:
Perhaps we don't need this, as the current unit conversion stuff we have is maybe sufficient

PR #2296, opened 2022-04-28
Added ability to convert units in profile viewer

astrofrog:
This adds initial support for unit conversion in the profile viewer, as a test case that is also useful for jdaviz.

Remaining TODOs:

* [x] Make sure this works properly with glue-jupyter
* [x] Add documentation about pluggable conversion classes
* [x] Add tests

### DIFF
--- a/doc/customizing_guide/units.rst
+++ b/doc/customizing_guide/units.rst
@@ -1,0 +1,71 @@
+Unit conversion in glue
+=======================
+
+.. note:: Support for automatic unit conversion in glue is experimental - at the moment
+          the ability to select units for the x and y axes is only available in the profile viewer.
+
+Data components can be assigned units as a string (or `None` to indicate no known units).
+By default, glue uses `astropy.units <https://docs.astropy.org/en/stable/units/index.html>`_ package
+to theb carry out unit conversions based on these units. However, it is possible to customize the
+unit conversion machinery, either to use a different unit transformation machinery, or to specify
+e.g. equivalencies in the astropy unit conversion. To customize the unit conversion behavior, you
+will need to define a unit converter as shown below::
+
+    from astropy import units as u
+    from glue.core.units import unit_converter
+
+    @unit_converter('custom-name')
+    class MyCustomUnitConverter:
+
+        def equivalent_units(self, data, cid, units):
+            # Given a glue data object (data), a component ID (cid), and the units
+            # of that component in the data object (units), this method should
+            # return a flat list of units (as strings) that the data could be
+            # converted to. This is used to construct the drop-down menus with the
+            # available units to convert to.
+
+        def to_unit(self, data, cid, values, original_units, target_units):
+            # Given a glue data object (data), a component ID (cid), the values
+            # to convert, and the original and target units of the values, this method
+            # should return the converted values. Note that original_units
+            # gives the units of the values array, which might not be the same
+            # as the original native units of the component in the data.
+
+In both methods, the data and cid are passed in not to get values or units (those should be
+used from the other arguments to the methods) but rather to allow logic for the unit
+conversion that might depend on which component is being converted. An example of
+a simple unit converter based on `astropy.units`_ would be::
+
+    from astropy import units as u
+    from glue.core.units import unit_converter
+
+    @unit_converter('example-1')
+    class ExampleUnitConverter:
+
+        def equivalent_units(self, data, cid, units):
+            return map(u.Unit(units).find_equivalent_units(include_prefix_units=True), str)
+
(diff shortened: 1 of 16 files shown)


pllim (commented):
How do you pass in equivalencies?

How do you ensure flux conversion would be correct, especially in non-linear flux units?

astrofrog:
As mentioned in https://github.com/glue-viz/glue-jupyter/pull/311 I am planning on making the unit conversion pluggable/swappable

pllim:
When you have coded it, I am gonna throw in some magnitudes to make your life interesting, maybe. 😉

doc/customizing_guide/units.rst
```diff
@@ -0,0 +1,71 @@
+Unit conversion in glue
+=======================
+
+.. note:: Support for automatic unit conversion in glue is experimental - at the moment
+          the ability to select units for the x and y axes is only available in the profile viewer.
+
+Data components can be assigned units as a string (or `None` to indicate no known units).
+By default, glue uses `astropy.units <https://docs.astropy.org/en/stable/units/index.html>`_ package
```
pllim, line 8: Do you just wanna use intersphinx here?

doc/customizing_guide/units.rst
```diff
@@ -0,0 +1,71 @@
...
+to theb carry out unit conversions based on these units. However, it is possible to customize the
+unit conversion machinery, either to use a different unit transformation machinery, or to specify
+e.g. equivalencies in the astropy unit conversion. To customize the unit conversion behavior, you
```
pllim, line 11:
```suggestion
unit conversion machinery, either to use a different unit transformation machinery, or to specify,
e.g., equivalencies in the astropy unit conversion. To customize the unit conversion behavior, you
```

doc/customizing_guide/units.rst
```diff
@@ -0,0 +1,71 @@
...
+will need to define a unit converter as shown below::
```
pllim, line 12: So, this is a way to bypass built-in astropy equivalencies? 👀

pllim:
p.s. I think at its current state, this PR broke Specviz parser. I had this installed and was doing something else and it crashed. The error went away when I uninstalled this branch and reinstalled glue-core 1.3.0

doc/customizing_guide/units.rst
astrofrog: Fixed

doc/customizing_guide/units.rst
astrofrog: Yes

doc/customizing_guide/units.rst
```diff
@@ -0,0 +1,71 @@
+Unit conversion in glue
+=======================
+
+.. note:: Support for automatic unit conversion in glue is experimental - at the moment
+          the ability to select units for the x and y axes is only available in the profile viewer.
+
+Data components can be assigned units as a string (or `None` to indicate no known units).
```
pllim, line 7: This breaks backward compatibility -- See failures in https://github.com/spacetelescope/jdaviz/actions/runs/4053913069/jobs/6975103932

astrofrog: Thanks, will have a think if there is a way we can avoid breaking backward compatibility

glue/viewers/profile/state.py
```diff
@@ -338,6 +402,13 @@ def update_profile(self, update_limits=True):
             axis_view[pix_cid.axis] = slice(None)
             axis_values = data[self.viewer_state.x_att, tuple(axis_view)]
 
+            converter = UnitConverter()
+            axis_values = converter.to_unit(self.viewer_state.reference_data,
+                                            self.viewer_state.x_att, axis_values,
+                                            self.viewer_state.x_display_unit)
```
pllim, line 408: Something does not look quite right here. Why does this use reference data but the next line use input data? I tried changing this but I cannot make it work.

I am seeing over at https://github.com/spacetelescope/jdaviz/pull/2048 that when I add a second spectrum with GHz (first spectrum is loaded in micron and plot is showing micron), the second data plots GHz values as if they are micron without any conversion.

The inputs to custom Jdaviz converter registers as such:

```python
# Signature: to_unit(self, data, cid, values, original_units, target_units)

data = Jy_um  # But I am loading mJy_GHz
cid = World 0
values = [142758.31333333 123202.38       108358.71975904  96707.24451613   87318.19165049  79590.9180531   73120.11170732  67622.35894737   62893.52265734  58782.83490196]
original_units = um  # This is wrong, should be GHz
target_units = um
```

It is also very confusing that the custom converter takes one more arg than what is really used over here.

PR #2297, opened 2022-05-12
Allow specifying visual attributes when creating subset group

Carifio24:
This PR allows specifying visual attributes when creating a new subset group, either via the `SubsetGroup` initializer or the `DataCollection::new_subset_group` method.

I don't know that this adds much, if any, utility to the standard glue workflow - the intended use case here is for dashboard-style applications like Cosmic Data Stories. In a setting where the application is doing the styling, the delay between creating the new subset group and modifying the styling is occasionally visible, which this PR is looking to solve.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 Full changelog
 ==============
 
+v1.4.0 (unreleased)
+-------------------
+
+* Add support for specifying visual attributes when creating a subset group. [#2297]
+
 v1.3.0 (2022-04-22)
 -------------------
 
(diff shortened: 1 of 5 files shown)


astrofrog (changes requested):
Looks good! Could you add a test though?

Carifio24:
@astrofrog I've added two tests - one to check that the correct default values are applied in `new_subset_group`, and another with an explicit set of visual attributes. Let me know if there are any other sorts of tests that you think would be good to add.

I've also changed the default value for `markersize` in the `SubsetGroup` constructor to 7, rather than 7.5, since the `VisualAttributes` setter converts it into a `int` anyways. (I had originally put in 7.5 to match the existing value that was assigned - the default value for `markersize` in `VisualAttributes` is 3, and the current code for `new_subset_group` multiplies the `markersize` by 2.5).

PR #2298, opened 2022-05-17
Modify profile viewer so that when in 'Sum' mode, parts of profiles with no valid values are NaN rather than zero

astrofrog:


pllim:
[🐱](https://jira.stsci.edu/browse/JDAT-2464)

PR #2300, opened 2022-05-23
Redraw empty histogram layer

Carifio24:
This PR addresses an issue that was discussed at a glue meeting a little while ago (don't remember the exact date). The issue is that if a subset group is modified to make it empty, the corresponding histogram layer artist is not redrawn. See the video below.

https://user-images.githubusercontent.com/14281631/169853796-d7dd56b3-7d5b-4ed8-90a2-5931bc0a6e2e.mp4

The issue seems to be [here](https://github.com/glue-viz/glue/blob/main/glue/viewers/histogram/layer_artist.py#L82), where `_update_artists` return if the sum of the histogram is empty (that is, the layer has no content). To fix this, I've removed the `mpl_hist.sum() == 0` check. Allowing the possibility for empty layer opens the door for the `largest_y_max` and `smallest_y_min` values computed later in the method to be NaN or infinite (in particular, if all of the layers are empty). I couldn't think of a better default behavior than just punting on updating the bounds in this case, so I've added `numpy.isfinite` checks around both assignments to the relevant viewer state properties.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,12 @@
 Full changelog
 ==============
 
+v1.4.0 (2022-05-31)
+-------------------
+
+* Fix an issue where the histogram layer artist would not redraw if
+* the histogram sum was zero. [#2300]
+
 v1.3.0 (2022-04-22)
 -------------------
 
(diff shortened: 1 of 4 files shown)


astrofrog:
@Carifio24 - this looks good! Could you add a test and changelog entry?

Carifio24:
@astrofrog I've added a changelog entry and a test (`TestHistogramViewer::test_redraw_empty_subset`) that checks whether all of the `Rectangle`s in the layer artist have height zero after the subset state is set to an empty one.

astrofrog (approved):
Looks good, thanks!

PR #2302, opened 2022-05-26
Fix bug when chunking large arrays with a subset defined

rosteen:
## Description

We were running into a bug for large cubes in Jdaviz, where we couldn't retrieve the collapsed spectrum if the cube was above a certain size (above `n_chunk_max` in `compute_statistic` as it turned out). I'm not sure if this is the best fix, but it works to avoid the broadcast error we were encountering on [line 1689](https://github.com/glue-viz/glue/blob/bc9382839d4a28d82c5a24aaedd50708bb198a00/glue/core/data.py#L1689) where there was an array size mismatch with a `RangeSubsetState` defined and number of array elements > `n_chunk_max`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,9 @@ v1.4.0 (unreleased)
 
 * Add support for using degrees in full-sphere projections. [#2279]
 
+* Fixed a bug in when using `compute_statistic` on an array large enough to
+  need chunking when a subset is defined. [#2302]
+
 v1.3.0 (2022-04-22)
 -------------------
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,8 +11,9 @@ v1.4.0 (unreleased)
 
 * Add support for using degrees in full-sphere projections. [#2279]
 
-* Fixed a bug in when using `compute_statistic` on an array large enough to
-  need chunking when a subset is defined. [#2302]
+* Fixed a bug on setting a return view in `compute_statistic` when a subset
+  is defined, resulting in a broadcast error in arrays large enough to need
+  chunking. [#2302]
 
 v1.3.0 (2022-04-22)
 -------------------
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -19,7 +19,7 @@ from glue.core.message import (DataUpdateMessage, DataRemoveComponentMessage,
 from glue.core.decorators import clear_cache
 from glue.core.util import split_component_view
 from glue.core.hub import Hub
-from glue.core.subset import Subset, SubsetState, SliceSubsetState
+from glue.core.subset import Subset, SubsetState, SliceSubsetState, RangeSubsetState
 from glue.core.component_id import ComponentIDList
 from glue.core.component_link import ComponentLink, CoordinateComponentLink
 from glue.core.exceptions import IncompatibleAttribute
@@ -1643,7 +1643,7 @@ class Data(BaseCartesianData):
                 len(axis) > 0 and
                 len(axis) == self.ndim - 1 and
                 self.size > n_chunk_max and
-                not isinstance(subset_state, SliceSubsetState)):
+                not isinstance(subset_state, (SliceSubsetState, RangeSubsetState))):
 
             # We operate in chunks here to avoid memory issues.
 
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -1839,7 +1839,6 @@ class Data(BaseCartesianData):
             else:
                 chunk_shape = subset_state.to_mask(self, chunk_view).shape
                 full_shape = [chunk_shape[idim] for idim in range(self.ndim) if idim not in axis]
-                view_start = [chunk_view[idim].start for idim in range(self.ndim) if idim not in axis][0]
 
             full_result = np.zeros(full_shape) * np.nan
             full_result[result_slices] = result
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -1643,7 +1643,7 @@ class Data(BaseCartesianData):
                 len(axis) > 0 and
                 len(axis) == self.ndim - 1 and
                 self.size > n_chunk_max and
-                not isinstance(subset_state, (SliceSubsetState, RangeSubsetState))):
+                not isinstance(subset_state, SliceSubsetState)):
 
             # We operate in chunks here to avoid memory issues.
 
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -19,7 +19,7 @@ from glue.core.message import (DataUpdateMessage, DataRemoveComponentMessage,
 from glue.core.decorators import clear_cache
 from glue.core.util import split_component_view
 from glue.core.hub import Hub
-from glue.core.subset import Subset, SubsetState, SliceSubsetState, RangeSubsetState
+from glue.core.subset import Subset, SubsetState, SliceSubsetState
 from glue.core.component_id import ComponentIDList
 from glue.core.component_link import ComponentLink, CoordinateComponentLink
 from glue.core.exceptions import IncompatibleAttribute
@@ -1694,6 +1694,7 @@ class Data(BaseCartesianData):
         # later we will need to pad out the result of compute_statistic.
         subarray_slices = None
 
+        chunk_view = None
         if subset_state:
             if isinstance(subset_state, SliceSubsetState) and view is None:
                 mask = None
@@ -1766,6 +1767,8 @@ class Data(BaseCartesianData):
                                 mask_idim += 1
                             else:
                                 new_view.append(view[idim])
+                        # This is the chunk view, which we'll need later
+                        chunk_view = view
                         view = tuple(new_view)
                     else:  # pragma: nocover
                         # This should probably never happen, but just in case!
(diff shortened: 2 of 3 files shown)


pllim:
[🐱](https://jira.stsci.edu/browse/JDAT-2424)

pllim:
Woohoo... looking green!

pllim:
`py310-test-pyqt515-all` failure is not our fault, right?

```
==> macos OS detected
https://uploader.codecov.io/latest/macos/codecov.SHA256SUM
Error: Codecov: Error validating uploader: Misformed armored text
```

pllim:
p.s. I thought this was the astropy conda-forge PR. Sorry for the noise wrt CI status. 😆

dhomeier:
What works for `SliceSubsetState` should work for `RangeSubsetState`, so if there are no other memory issues to be aware of...

dhomeier:
I was more concerned about the memory footprint if chunking is bypassed for huge arrays, but at the location of the broadcast error you pointed to the chunking seems already to be taken care of in the `iterate_chunks` loop.
Leaves the question if `compute_statistic` might anywhere be called directly on a (huge) `RangeSubsetState`.

dhomeier:
> It seemed to work fine dropping into the else case that uses `to_mask` though.

Yes, that would have already worked in the `self.size <= n_chunk_max` case, so should not be affected by larger array sizes.

astrofrog:
I'd like to understand a bit better why this is not working currently - in principe it is possible one would use a RangeSubsetState with a huge array. The reason we special case ``SliceSubsetState`` is because it can be applied efficiently to a huge array using the standard ``to_mask``, but this is not the case with ``RangeSubsetState``.

@dhomeier - would you have time to investigate if we can fix this without special-casing ``RangeSubsetState``?

rosteen:
I'm pretty sure that one of `result[chunk_view[axis_index]]` and `values` has the length of the subset and the other has the length of the full axis, resulting in a mismatch error. Agreed that it's better to resolve that than just bail on chunking for `RangeSubsetState`.

dhomeier:
> @dhomeier - would you have time to investigate if we can fix this without special-casing `RangeSubsetState`?

Sure. Not sure how long it will take; I am not that familiar with that code yet – just with regard to the bugfix release.

pllim:
Is it too controversial to merge this as-is to get the bug fix release out and then investigate a better fix as a follow-up PR? A working but horrible performance solution is better than completely broken?

dhomeier:
> I'm pretty sure that one of `result[chunk_view[axis_index]]` and `values` (on the line linked in the original comment) has the length of the subset and the other has the length of the full axis, resulting in a mismatch error. Agreed that it's better to resolve that than just bail on chunking for `RangeSubsetState`.

Yes, from the exception `values` apparently has length of the original `self.shape[axis_index]`, while `result[chunk_view[axis_index]]` has the same divided by `n_chunks`.
I could also confirm the exception for an `InequalitySubsetState`, so special-casing `RangeSubsetState` will not get us very far. I guess this more likely applies to any `SubsetState` other than `SliceSubsetState` that is already separately treated.

> A working but horrible performance solution is better than completely broken?

It's not just performance per se, since the function potentially doubles-triples the memory footprint.

If that is manageable in your Jdaviz application, maybe a better workaround for now would be to set `n_chunk_max=data.shape` for those cases?

pllim:
> a better workaround for now would be to set `n_chunk_max=data.size` for those cases?

The only call of `compute_statistic` I can find in Jdaviz is here but Ricky would know better if that is the right place to do it or not.

https://github.com/spacetelescope/jdaviz/blob/70eb5f16aa78e8de68faa3c7b8afbb7096f067ad/jdaviz/app.py#L709

rosteen:
@dhomeier @astrofrog I did some more investigating and implemented a fix to actually generate the correct result shape rather than skipping the chunking altogether for `RangeSubsetState`s. Let me know if this looks good to you, and if so @dhomeier if you want to add your additional tests from #2304 to this PR, or make a followup PR with them.

orifox:
Close! This is working much better than before! I can scrub through the cube. But when I draw a spectral subset, it doesn't seem to register in the plugin tool subset dropdown.

Screen Shot 2022-06-23 at 11 51 34 AM

rosteen:
Strange, that works for me. Maybe try updating your Jdaviz install to the latest dev version? I had a PR related to the subset plugin merged recently that might fix this for you.

Screen Shot 2022-06-23 at 12 06 59 PM



I'll investigate the test failure soon.

orifox:
Try a subset with 2 separated spectral regions. This might be an unrelated bug.

rosteen:
@orifox Yes, that's exactly the case my recent PR fixed/improved - again, are you up to date with `main`? If so this is probably a new bug to track down, if not that will probably solve your issue.

Screen Shot 2022-06-23 at 12 41 01 PM

orifox:
I just installed the dev branch an hour ago.

orifox:
2.6.1.dev79+gcf22f84d

rosteen:
Got it, good to know - probably a real bug then.

pllim:
Does fixing failing test also magically fixes Ori's use case?

dhomeier:
It does fix the extended `test_compute_statistic_chunks` in #2304, but not the added `test_compute_statistic_shape_view`, so some progress showing up, thanks!

glue/core/data.py
```diff
@@ -1828,9 +1831,16 @@ def compute_statistic(self, statistic, cid, subset_state=None, axis=None,
             # of the if statement above.
```
dhomeier, line 1831:
```suggestion
            # of the if statement above. Likewise if a view was specified,
            # only the result within the view is returned.
```
the `view` parameter could need some real documentation in general...

glue/core/tests/test_data.py
```diff
@@ -991,6 +1036,43 @@ def test_compute_statistic_shape():
...
+    assert result.shape == (view[1].stop - view[1].start,)
+
+    roi = RangeROI('x', min=1.5, max=2.0)
+    state = roi_to_subset_state(roi, x_att='x')
+
+    result = data.compute_statistic('sum', data.id['x'], subset_state=state, view=view)
+    assert np.isscalar(result)
+
+    result = data.compute_statistic('sum', data.id['x'], subset_state=state, view=view, axis=1)
+    assert result.shape == (view[0].stop - view[0].start, view[1].stop - view[1].start)
```
dhomeier, line 1070:
```suggestion
    assert result.shape == (view[0].stop - view[0].start, view[2].stop - view[2].start)

    result = data.compute_statistic('sum', data.id['x'], subset_state=state, view=view, axis=(0, 2))
    assert result.shape == (view[1].stop - view[1].start,)

    roi = RangeROI('x', min=1.5, max=2.0)
    state = roi_to_subset_state(roi, x_att='x')

    result = data.compute_statistic('sum', data.id['x'], subset_state=state, view=view)
    assert np.isscalar(result)

    result = data.compute_statistic('sum', data.id['x'], subset_state=state, view=view, axis=1)
    assert result.shape == (view[0].stop - view[0].start, view[2].stop - view[2].start)
```
Doh, miscounted the axes!

CHANGES.md
```diff
@@ -11,6 +11,9 @@ v1.4.0 (unreleased)
 
 * Add support for using degrees in full-sphere projections. [#2279]
 
+* Fixed a bug in when using `compute_statistic` on an array large enough to
+  need chunking when a subset is defined. [#2302]
```
dhomeier, line 15:
```suggestion
* Fixed a bug on setting a return view in `compute_statistic` when a subset
  is defined, resulting in a broadcast error in arrays large enough to need
  chunking. [#2302]
```
I think this captures the full scope of the bug better, even if chunking may be the only case where `view` is used in  practice (maybe that's why it is somewhat treated like a private parameter).

glue/core/data.py
```diff
@@ -1825,12 +1828,20 @@ def compute_statistic(self, statistic, cid, subset_state=None, axis=None,
...
-            full_shape = [self.shape[idim] for idim in range(self.ndim) if idim not in axis]
-            full_result = np.zeros(full_shape) * np.nan
             result_slices = tuple([subarray_slices[idim] for idim in range(self.ndim) if idim not in axis])
+
+            if chunk_view is None:
+                full_shape = [self.shape[idim] for idim in range(self.ndim) if idim not in axis]
+            else:
+                chunk_shape = subset_state.to_mask(self, chunk_view).shape
+                full_shape = [chunk_shape[idim] for idim in range(self.ndim) if idim not in axis]
+                view_start = [chunk_view[idim].start for idim in range(self.ndim) if idim not in axis][0]
```
dhomeier, line 1842:
```suggestion
```
This is never used anymore in this branch.

dhomeier (approved):
Turned out I messed up my own tests; apart from the minor suggestions this looks great, thanks! The PySide2 crashes are known issues and unrelated.

glue/core/data.py
rosteen: Whoops, good catch.

rosteen:
> Turned out I messed up my own tests

Ha, it happens! Thanks for the review, I committed your remaining suggestions.

dhomeier:
Thanks again for the fix @rosteen!

pllim:
Yay! @orifox should try again with the latest dev version of glue-core 🤞

orifox:
Works beautifully! I have confirmed I can now do all functionality I originally set out to do under this set of issues/tickets. Screenshot below. I sign off on it. If there are other things that come up in the future, we can open new issues.

Screen Shot 2022-06-27 at 7 22 13 AM

pllim:
@dhomeier , is it possible to do a release with this patch soon? 🙏

dhomeier:
@astrofrog what do you think? I'll merge #2235 today; could make a 1.5.0 then.

astrofrog:
Sounds good, thanks!

dhomeier:
v1.5.0 is on PyPI now; thanks all again for their contributions!

PR #2304, opened 2022-05-29
Fix return shape of compute_statistic with view set or computed in chunks

dhomeier:
## Description
This is a follow-up/replacement for #2302, which identified a broadcast error when calling `compute_statistic` with a `RangeSubsetState` defined on `data.size > n_chunk_max`.
The underlying issue appears to be that a call with `subset_state` and `view` set like
`data.compute_statistic('mean', data.id['x'], axis=(1, 2), subset_state=subset_state, view=(slice(0, 6), slice(None), slice(None)))`
returns an array of the original axis-0 length with values past [:6] filled with NaN, while the same without a `subset_state` returns a length-6 array. This was probably introduced with the shape adjustments added as part of #2147.
Attempts to set the return shape in https://github.com/glue-viz/glue/blob/bc9382839d4a28d82c5a24aaedd50708bb198a00/glue/core/data.py#L1831
from `data.shape` or `mask.shape` instead did fix this issue, but broke the test cases in `test_compute_statistic_shape` again, so instead I am tracking here if a `view` was initially specified to directly return `result`.
I honestly do not completely understand how this works now for the combination of the efficient subset calculation _and_ `view` set, but the added tests are passing...

@rosteen could you test if this also fixes your case?

### DIFF
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -1633,6 +1633,11 @@ class Data(BaseCartesianData):
 
         # TODO: generalize chunking to more types of axis
 
+        if view is None:
+            use_view = False
+        else:
+            use_view = True
+
         # In recent version of Numpy, using lists is not the same as using
         # tuples, so we make sure we always use tuples to avoid confusion.
         if isinstance(view, list):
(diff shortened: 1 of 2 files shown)


dhomeier:
Thanks, this looks similar to what I get in the added tests; really need better coverage of different shapes and views there.

rosteen:
FYI the shape of the data I'm working with is `(93, 75, 9099)`, if you want to test with a cube that size and confirm you see the same shapes.

dhomeier:
Fixed/superseded by #2302

PR #2305, opened 2022-05-31
Apply foreground color to minor ticks

Carifio24:
There is currently a bug where any minor ticks displayed will not match the user's global settings. (Note that this isn't an issue with the default settings, since the labels are black either way.) One prominent example is the scientific notation tick marks that are sometimes used when displaying an axis in log mode. See below for an example:

log-label-colors

This occurs because we only set major tick labels to match the given color in `set_foreground_color` (see the matplotlib documentation [here](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.axes.Axes.tick_params.html#matplotlib.axes.Axes.tick_params); the default value for `which` is `"major"`). This PR fixes the issue by setting `which="both"`).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@ Full changelog
 v1.6.0 (unreleased)
 -------------------
 
+* Fixed a bug where minor tick marks were not respecting the settings colors. [#2305]
+
 * Fix an issue where the histogram layer artist would not redraw if
   the histogram sum was zero. [#2300]
 
@@ -17,7 +19,6 @@ v1.5.0 (2022-06-28)
 * Added rotation angle ``theta`` as a property to regions of interest,
   to be set on instantiation or modified using the ``rotate_to`` and
   ``rotate_by`` methods. [#2235]
-
 
 v1.4.0 (2022-05-31)
 -------------------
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -4,6 +4,8 @@ Full changelog
 v1.4.0 (2022-05-31)
 -------------------
 
+* Fixed a bug where minor tick marks were not respecting the settings colors. [#2305]
```
astrofrog, line 7: Can you move this to 1.6.0? (I am going to work on a way to avoid us having to do this in future)

astrofrog (changes requested):

Carifio24:
@astrofrog I've rebased and moved the changelog entry to the 1.6.0 section.

astrofrog:
Could you check the codestyle failure with ``tox -e codestyle`` - thanks!

Carifio24:
Hmm - it looks like the codestyle failure is actually coming from the main branch. Taking a quick look there, the codestyle check went from succeeding to failing after #2300 was merged - although codestyle passed in that PR and didn't change any of the offending files. I think the culprit is the E275 update described [here](https://pycodestyle.pycqa.org/en/latest/developer.html#changes).

astrofrog (approved):

dhomeier:
> any of the offending files. I think the culprit is the E275 update described [here](https://pycodestyle.pycqa.org/en/latest/developer.html#changes).

Thanks for identifying this @Carifio24 . The codestyle is fixed in #2308 now.

PR #2308, opened 2022-06-30
Add `tests` requirement to publish workflow in CI

dhomeier:
## Description
To let the `publish` job only run after all envs in `tests` have completed successfully; to allow this to actually happen added `allowed_failures` to the (mostly PySide2) envs that are failing or segfaulting more or less erratically.
Also adding a `codecov.yml` to keep it from failing at essentially zero changes (which is probably due rounding errors).

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -57,7 +57,6 @@ jobs:
         - macos: py38-test-pyqt514
         - macos: py310-test-pyqt515-all
           PLAT: arm64
-        - macos: py38-test-pyside514
 
         # Try out documentation build on macOS
         - macos: py37-docs-pyqt513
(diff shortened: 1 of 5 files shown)


dhomeier:
Dunno where codecov is still digging up 4 lines in a completely unmodified file... I'd not stopping `publish` however, so not that critical.
I've also tested a setup running `codestyle` first before anything else; might add that here as well.

astrofrog (approved):
Looks good if CI passes!

dhomeier:
Codestyle is fixed now, but there is a fresh failure on the Linux `pyqt515` runs now; for all recent PRs. No idea why the
> qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.

would  have popped back up – afaics all the installed xcb/xkb packages are identical.
Can you see any change between image versions [20220619.1](https://github.com/actions/runner-images/blob/ubuntu20/20220619.1/images/linux/Ubuntu2004-Readme.md) and 
[20220729.1](https://github.com/actions/runner-images/blob/ubuntu20/20220729.1/images/linux/Ubuntu2004-Readme.md) that might be responsible for this @astrofrog?

Issue #2215, opened 2021-08-03
Expose join_on_key in the UI


jfoster17:
**Is your feature request related to a problem? Please describe it:**
The very useful `join_on_key` functionality is currently only available through the terminal or scripting. Several past responses (#2050 #1917) suggest that there are plans to expose it through the UI, but I'm opening this as a specific issue for this feature.

**Describe the solution you'd like:**
``join_on_key`` should be available through the UI to allow linking different datasets by a common key/ID. I propose that this should be as part of the existing linking UI and that join_on_key links should be visualized with a different connection type (dashed line rather than solid?). 

**Describe alternatives you've considered:**
Alternatively, a separate window could enable the ``join_on_key`` functionality, since this is fundamentally a very different kind of linking operation.

astrofrog:
Short-term, the easiest is probably to list it as another link type in the existing list, so maybe e.g. 'join' or something similar instead of 'identity', as it is similar to a database join.

In the long term I think we want to re-think a bit the UI, and have yet more kinds of linking such as nearest-neighbour linking and so on. I think that we probably want a single visualization for all links with different line types for different conceptual kinds of linking. See also https://github.com/glue-viz/glue/issues/798 which was to keep track of the more general refactor.

jfoster17:
Ok -- I'm working on a PR doing the short-term/easy thing of just adding another link type == "join" in the current UI with a different line type. The basic idea was to use a special dummy superclass of ComponentLink (called KeyLink) to represent the join_on_key links in the UI and enable deleting them. 

I agree that generalizing the linking class hierarchy makes sense if we end up adding more linking types ala nearest-neighbor -- but generalizing for n=2 does not seem worthwhile right now.

PR #2313, opened 2022-08-03
Add the ability to create one-to-one 'join_on_key'-type links to the GUI link editor

jfoster17:
## Description

This is a combination of a plug-in and core changes to the link_editor to enable the creation, display, and removal of join_on_key type links (managed through a `Join_Link` object). This address the immediate need in #2215.


jfoster17:
Hey @astrofrog -- can you take a look at this solution? It's not... ideal, but it basically works. Basically I have a `Join_Link(LinkCollection)` object that represents the join_on_key operation and then the link_manager knows how to use this object to add or remove the appropriate `_key_joins`. Here are some of my concerns.

1. Because the key joins are not the same thing at the Join_Link object, if users (or code) directly calls join_on_key, this won't create a Join_Link and so there will be no representation of the join in the UI.
2. Because Join_Links are funny kinds of things, I end up with `data_collection._link_manager_external_links` holding `Join_Link`s, but `Join_Link`s do NOT show up in `data_collection.links`
3. Currently I have this limited to "Joining on single components". (I don't see a fundamental reason I couldn't extend it to multiple components or one-to-many joins I just haven't done that yet).
4. The display in the link manager is a little funny -- I draw these Join_Links as dashed lines, but a new Join_Link does not show up as dashed right away, only after closes and re-opening the link manager. This can probably be fixed in `link_editor.qt.data_graph` but I haven't dug into it a lot (it's not obvious what to display if two datasets are both joined_on_key and glued through a traditional link for some reason). 
5. I set up Join_Link as a plugin originally, hoping that I might be able to do things all as a plug-in. Now I have core components assume this plugin exists, but is obviously not ideal (a user might disable the plug-in). Presumably I should move it to core.link_helpers?

glue/core/link_manager.py
```diff
@@ -30,6 +30,7 @@
 from glue.core.subset import Subset
 from glue.utils import unbroadcast
 from glue.core.coordinate_helpers import dependent_axes
+from glue.plugins.join_on_key.link_helpers import Join_Link
```
astrofrog, line 33: I think the class should probably just be called JoinLink (no underscore)

glue/dialogs/link_editor/qt/data_graph.py
```diff
@@ -162,8 +166,12 @@ def get_connections(dc_links):
     for link in dc_links:
         data1 = link.data1
         data2 = link.data2
+        if link.join_link:
+            join_link = True
+        else:
+            join_link = False
         if (data1, data2) not in links and (data2, data1) not in links:
-            links.append((data1, data2))
+            links.append((data1, data2, join_link))
```
astrofrog, line 174: I wonder if instead of having join_link be a boolean we should have the third argument be a link_type which could be a string, then it could be currently either 'join' or some name we pick for the current kind of linking (I don't have strong opinions on what that should be currently, could be e.g. 'value' or something like that.

glue/plugins/join_on_key/link_helpers.py
```diff
@@ -0,0 +1,52 @@
+from glue.config import link_helper
+from glue.core.link_helpers import LinkCollection
+
+
+@link_helper(category="Join")
+class Join_Link(LinkCollection):
+    cid_independent = False
+
+    display = "Join on ID"
+    description = "Join two datasets on a common indetifier or index. \
```
astrofrog, line 10: typo in 'indetifier'

glue/plugins/join_on_key/link_helpers.py
```diff
@@ -0,0 +1,52 @@
...
+This is similar to a database join in that subsets defined \
+on one dataset can always propogate through joins."
```
astrofrog, line 12: I would perhaps give a little more detail and an example - to show how it differs to a classic link?

astrofrog (changes requested):
This is looking good so far! A few comments below.

> Because the key joins are not the same thing at the Join_Link object, if users (or code) directly calls join_on_key, this won't create a Join_Link and so there will be no representation of the join in the UI.

To some extent, join_on_key should perhaps become deprecated API once we are confident it can be treated as a regular link - that is, in future maybe creating a ``JoinLink`` should be the preferred way to do it. So I wouldn't worry about this.

> Because Join_Links are funny kinds of things, I end up with data_collection._link_manager_external_links holding Join_Links, but Join_Links do NOT show up in data_collection.links

Ok I don't know immediately why that would be the case, but we should be able to get that to work in future.

> Currently I have this limited to "Joining on single components". (I don't see a fundamental reason I couldn't extend it to multiple components or one-to-many joins I just haven't done that yet).

Sounds good

> The display in the link manager is a little funny -- I draw these Join_Links as dashed lines, but a new Join_Link does not show up as dashed right away, only after closes and re-opening the link manager. This can probably be fixed in link_editor.qt.data_graph but I haven't dug into it a lot (it's not obvious what to display if two datasets are both joined_on_key and glued through a traditional link for some reason).

Ok yes we can look into this later

> I set up Join_Link as a plugin originally, hoping that I might be able to do things all as a plug-in. Now I have core components assume this plugin exists, but is obviously not ideal (a user might disable the plug-in). Presumably I should move it to core.link_helpers?

Yes I think maybe it should just be a core link type.

The bottom line is that I'm happy for this to go into the core package, and I think it was overdue that joins become normal links - thanks @jfoster17!

glue/core/link_manager.py
jfoster17: Good point, I have renamed.

glue/dialogs/link_editor/qt/data_graph.py
jfoster17: That makes sense. For now I have 'join' and 'value' as the two link_types

glue/plugins/join_on_key/link_helpers.py
jfoster17: Fixed

glue/plugins/join_on_key/link_helpers.py
jfoster17: Taking inspiration from [the docs](http://docs.glueviz.org/en/stable/gui_guide/merging.html) I now describe this as:

> Join two datasets on a common ID. Other links 
> in glue connect data columns (two datasets have 'age' columns but 
> the rows are different objects), while Join on ID connects the same 
> rows/items across two datasets.

There isn't a great deal of space in the UI for a longer description, and concise is better anyway. Do you think this is sufficient?

jfoster17:
I assume you would like to be rebase and add a changelog entry when this is good-to-go?

astrofrog:
Yes that would be great, thanks!

glue/plugins/join_on_key/link_helpers.py
jfoster17: @astrofrog -- if you approve of this new description I think this is good-to-go?

Issue #1689, opened 2018-05-02
Have a proper way to colormap categorical components


astrofrog:
For categorical components, we should have a way to use discrete colormaps in e.g. scatter plots. Note that this requires figuring out how to distinguish between a true categorical component and an ordinal string variable.

PR #2317, opened 2022-08-18
Categoricals in scatter

jfoster17:
# Allow Categorical Variables to be Used in Color and Size Scaling in Scatter viewer

## Description

This allows scatter plot cmap and size scaling attributes to be a categorical. This does not include a qualitative color-map for categorical attributes (as proposed in #1689) and obviously the size scaling is, in general, not going to be very meaningful, but for a small number of categorical levels it helps makes the points distinct.

An unresolved problem with this is that it might break glue-jupyter scatter plot because that does not consistently call `ensure_numerical` on the relevant attributes.


astrofrog:
@jfoster17 - would you like me to review this at this point? If so, could you rebase and mark as ready to review? Would you also be able to check that things work properly in glue-jupyter and open a PR there if not?

astrofrog:
@jfoster17 - what is the status of this? Would you like me to review it?

jfoster17:
I'll be re-opening this under the glue-core/glue-qt division.

PR #2319, opened 2022-08-23
DOC: Remove mention of defunct viz packages

pllim:
Stumbled upon this and figured this should be removed.


pllim:
Jdaviz is a standalone package using the Jupyter stuff, so I don't think we should mention it on this particular page.

dhomeier:
So for the standalone app those packages are gone without a replacement (I don't know if the archived versions would still be functional)?

pllim:
Yes, they gone. No more supported.

dhomeier (approved):

PR #2322, opened 2022-09-15
Export color settings in matplotlib viewers

Carifio24:
Currently, the matplotlib viewers do not take into account glue's global color settings (foreground and background colors) when exporting to a Python script. This means that the exported figure will not match if the color settings differ from the default.

This PR updates the export script to take those color settings into account. The code added to the script header is partially taken from `update_foreground_color` and `update_background_color`. I've also added a test that changes the color settings, compares the exports, and resets the color settings to the default. I've placed this test in `BaseTestExportPython` since it's identical for each matplotlib viewer.

Also, in the course of making this update, I fixed a bug [here](https://github.com/glue-viz/glue/blob/main/glue/viewers/matplotlib/viewer.py#L42) where the wrong tick label size is used.


dhomeier:
Thanks for the PR. The only thing I am wondering is, should the colours perhaps only be exported if changed from the defaults? Otherwise a session with no custom colours set will hardcode the defaults, and those will be set even if the script is run in an environment with custom settings. But maybe that is preferable for best reproducibility; also I don't see a general way to find the default values (if they should change from 'black' and 'white' at any point).

dhomeier:
> I had been imagining the script as reproducing the exact view that the user had at the time of export, and so hadn't worried about that.

I see a point in that as well, as it is primarily recreating one plot without much connection to the glue (app) configuration on that end. Not sure what the most typical use case is here.

Carifio24:
I agree that we should match whatever is the most typical use case. @astrofrog do you have any thoughts on which direction is more appropriate here?

astrofrog:
That's a good question - in the past we've tried to make sure the outputs match exactly. However we also want to make it easy for people to make figures look nice for including in a paper, and they might use a 'dark' mode in the UI but want to have the opposite for the exported figure. I think we do want the outputs to match by default but maybe we can reduce the amount of code in the script to be a one-liner that calls a glue function, for example ``set_figure_colors(background=..., foreground=...)`` and users can easily remove or update that line then. Does this seem like a reasonable compromise?

Carifio24:
@astrofrog I think that's a reasonable approach. I've implemented that here - let me know if you see any issues

astrofrog (approved):
Looks good, thanks!

Issue #2193, opened 2021-05-22
log should not appear in axes labels


Morisset:
**Describe the bug**

When a scatter plot is used and the user click on "log" button, the axis scale is changed to log (good), and the axis label changed to have "Log" in front of the normal label (wrong). **The label should not change**, only the axis scale is changed, but what is plotted is still the original data, not the log value of the data, so no "Log" is needed in the label.

<img width="474" alt="Capture d’écran 2021-05-22 à 16 26 28" src="https://user-images.githubusercontent.com/442951/119231783-721a6700-bad7-11eb-98c3-01c99911704b.png">

astrofrog:
@Morisset - indeed! Would you be interested in opening a pull request to fix this?

Morisset:
Well, I do not where in the code this is done ;-) If you just give me the name of the python file, I can look for it, fork glue, correct it and request the pool.

Morisset:
@astrofrog I do not know where is in the package the plotting routine that I have to change.

PR #2323, opened 2022-09-15
Don't change labels when using log scale axes

Carifio24:
This PR addresses #2193. As pointed out there, "Log" shouldn't be prepended to the axis label when using log scale axes, since we aren't actually plotting the log of the component values. This also modifies the `test_axes_labels` test to account for this change.


dhomeier (approved):
Looks all good, thanks! Wondering if that pyside514 test should be moved to `allowed_failures`.

PR #2328, opened 2022-10-10
Implement linking support for all BaseCartesianData subclasses

astrofrog:
This is an attempt to implement linking for all BaseCartesianData subclasses and making sure the fixed resolution buffer calculation works for these.

TODO:

* [ ] Any required updates to documentation
* [ ] Make sure that we re-instate support for pixel<->world linking for datasets that use this
* [ ] Check that things work properly in a real-life test


astrofrog:
Seems to work ok but noticing some slowness in the UI which usually indicates an infinite loop of some kind, so will investigate

PR #2329, opened 2022-10-10
Fix matplotlib and other deprecations

dhomeier:
## Description

Fixing a failing test due to changed `matplotlib.cm` API and various deprecation warnings.
The `external.echo` warning filter is a bit problematic in that it depends on `glue.viewers` being tested after `glue.plugins` if at all, otherwise (if `glue.external.echo` has not been already imported)
`glue/viewers/histogram/qt/tests/test_data_viewer.py::TestHistogramViewer::test_session_back_compat[1]`
will raise a warning as well.
The last commit replaces deprecated matplotlib `canvas.*_event` centrally in `viewers/matplotlib/qt/widget` as a quicker (and dirtier?) fix to changing all calls in the tests individually; may need discussion. The reasons for the upstream change are discussed in matplotlib/matplotlib#16931.


astrofrog:
I think adding the methods to the central Matplotlib widget is fine - thanks!

dhomeier:
Might consider using more exact copies of the versions in `matplotlib.backend_bases`, but as they are only used in the tests, those should suffice.
There's 24 calls in 3 test files in total; could perhaps be changed in place once we drop mpl < 3.6.

dhomeier:
I've added `lts` envs with matplotlib 3.5.* to isolate the errors in the py310 jobs; otherwise no idea what more to make of the 6 recurring `negative xerr` "Exceptions caught in Qt event loop" for each test...
While it's failing here for Linux and macOS, but not Windows, the same tests all pass with mpl 3.6.0 for both my Mac systems, arm64 and x86_64 alike.

PR #2331, opened 2022-10-11
Setup automated changelog generation for release process

dhomeier:
## Description

Adding workflow and configuration for updating changelog automatically from PR titles and labels, using GH actions.

The next release action should thus create entries for these PRs (+ any to come until then), to be verified:
```
### v1.6.0 (unreleased)

### What's Changed

#### Bug Fixes

- Modify scatter viewer to not prepend axis labels with 'Log' when using log scale axes. [#2323]

- Fixed a bug where minor tick marks were not respecting the settings colors. [#2305]

- Fix an issue where the histogram layer artist would not redraw if
  the histogram sum was zero. [#2300]
```

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,21 @@
+changelog:
+  exclude:
+    authors:
+      - pre-commit-ci
+    labels:
+      - no-changelog-entry-needed
+      - skip-changelog
+
+  categories:
+    - title: Bug Fixes
+      labels:
+        - bug
+    - title: New Features
+      labels:
+        - enhancement
+    - title: Documentation
+      labels:
+        - documentation
+    - title: Other Changes
+      labels:
+        - "*"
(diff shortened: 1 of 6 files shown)


dhomeier:
The [release guide](https://glueviz--2331.org.readthedocs.build/en/2331/developer_guide/release.html) is now only linked from the `RELEASE.rst` file, not the general developer docs - but probably should not be that widely needed.
Still could add a line there to write descriptive PR labels.

astrofrog (approved):
Looks good but can you push the last commit again without '[ci skip]' as we need to make sure the docs builds work? Also can you link to the release docs from the developer docs section in ``index.rst``? (I think it would be good to have for completeness)

dhomeier:
In  the top-level index or in `developer_guide.rst`? I agree it should be navigable somewhere from the TOC, but is probably specific enough for the latter.

astrofrog:
Yes sorry that's what I meant :facepalm:

PR #2334, opened 2022-10-13
Require Python>=3.8 and PyQt5/PySide5>=5.14

astrofrog:
As it's becoming increasingly hard to test against all version combinations, I think it's reasonable to now bump the minimum required Python and PyQt/PySide (and looks like PyQt5 5.15 is now the latest version on conda as of a few months ago).

I have tried to simplify the build matrix a bit - and for now I've disabled the allowed failures and will add things back there if they do indeed fail.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{37,38,39,310}-{codestyle,test,docs}-{pyqt59,pyqt510,pyqt511,pyqt512,pyqt513,pyside511,pyside512,pyside513,pyside514,pyside515,pyqt63,pyside63}-all-{dev,legacy}
+    py{38,39,310}-{codestyle,test,docs}-{pyqt514,pyqt515,pyside514,pyside515,pyqt63,pyside63}-all-{dev,legacy}
 requires = pip >= 18.0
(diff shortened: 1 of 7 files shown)


tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{37,38,39,310}-{codestyle,test,docs}-{pyqt59,pyqt510,pyqt511,pyqt512,pyqt513,pyside511,pyside512,pyside513,pyside514,pyside515,pyqt63,pyside63}-all-{dev,legacy}
+    py{38,39,310}-{codestyle,test,docs}-{pyqt514,pyqt515,pyside514,pyside515,pyqt63,pyside63}-all-{dev,legacy}
```
dhomeier, line 3: Where is `envlist` used anyway? `pyqt64` is missing here, but that does not seem to break anything.

astrofrog: Yes it's just for the 'run everything' which I never use anyway. But I can add the missing one.

astrofrog: Actually since not critical I will add it in my PyQt6 PR

dhomeier (approved):

PR #2335, opened 2022-10-14
Fix export to Python script to correctly use `Coordinates` or `WCS` in `reference_data`

dhomeier:
## Description
This fixes an issue I found when trying to use the `Save Python script` function on a viewer using a dataset with a standard (Astropy) WCS, such as the example `w5.fits` data. The script thus created failed with an
> AttributeError: 'astropy.wcs.Wcsprm' object has no attribute 'pixel_n_dim'

on `ax.reset_wcs(slices=['x', 'y', 0], wcs=ref_data.coords.wcs)` as `ref_data.coords` itself is already the WCS.

The fix, to test for that case first in the exporter, seems quite straightforward.
I've parametrised some test to run on a dataset with WCS; could think about running this with all tests, or always adding a WCS in `self.data.coords`; however `test_subset_legend` still fails with this patch, as the version with WCS somehow is not including the subset label:
expected
glue_plot

Basically in the latter case the script is missing these lines:
```python
## Layer 2: mysubset

layer_data = data_collection[0].subsets[0]

# Define a function that will get a fixed resolution buffer of the mask
handle = mpatches.Patch(color='#e31a1c', alpha=0.5)
legend_handles.append(handle)
legend_labels.append(layer_data.label)
```
so this looks like another actual bug in the exporter (those lines were already missing before this change), as if in
https://github.com/glue-viz/glue/blob/aabb5ebb1b5227f973a99f9aa905d6bfe6988b0b/glue/viewers/common/viewer.py#L447

the subset layer is not `visible and enabled` when a WCS ist present (indeed `layer.enabled = False`), but I could not yet figure out why.

Actually comparing to the non-WCS plot it does not seem like the subset is really displayed in the WCS version:
glue_plot

### DIFF
--- a/glue/viewers/image/qt/tests/test_python_export.py
+++ b/glue/viewers/image/qt/tests/test_python_export.py
@@ -41,23 +41,23 @@ class TestExportPython(BaseTestExportPython):
     @pytest.mark.parametrize('wcs', [False, True])
     def test_simple(self, tmpdir, wcs):
         if wcs:
-            self.viewer = self.app.new_data_viewer(ImageViewer)
             self.viewer.add_data(self.data_wcs)
+            self.viewer.remove_data(self.data)
         self.assert_same(tmpdir)
 
     @pytest.mark.parametrize('wcs', [False, True])
     def test_simple_legend(self, tmpdir, wcs):
         if wcs:
-            self.viewer = self.app.new_data_viewer(ImageViewer)
             self.viewer.add_data(self.data_wcs)
+            self.viewer.remove_data(self.data)
         self.viewer.state.show_legend = True
         self.assert_same(tmpdir)
 
     @pytest.mark.parametrize('wcs', [False, True])
     def test_simple_att(self, tmpdir, wcs):
         if wcs:
-            self.viewer = self.app.new_data_viewer(ImageViewer)
             self.viewer.add_data(self.data_wcs)
+            self.viewer.remove_data(self.data)
         self.viewer.state.x_att = self.data.pixel_component_ids[1]
         self.viewer.state.y_att = self.data.pixel_component_ids[0]
         self.assert_same(tmpdir)
@@ -65,8 +65,8 @@ class TestExportPython(BaseTestExportPython):
     @pytest.mark.parametrize('wcs', [False, True])
     def test_simple_visual(self, tmpdir, wcs):
         if wcs:
-            self.viewer = self.app.new_data_viewer(ImageViewer)
             self.viewer.add_data(self.data_wcs)
+            self.viewer.remove_data(self.data)
         self.viewer.state.legend.visible = True
         self.viewer.state.layers[0].cmap = plt.cm.RdBu
         self.viewer.state.layers[0].v_min = 0.2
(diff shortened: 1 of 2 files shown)


dhomeier:
OK, looks like the problem was using the wrong `Data` for the mask, effectively creating an empty subset.

glue/viewers/image/viewer.py
```diff
@@ -214,24 +214,26 @@ def _script_header(self):
...
 
-        script += "ref_data = data_collection[{0}]\n".format(dindex)
+        script += f"ref_data = data_collection[{dindex}]\n"
 
         ref_coords = self.state.reference_data.coords
 
-        if hasattr(ref_coords, 'wcs'):
-            script += "ax.reset_wcs(slices={0}, wcs=ref_data.coords.wcs)\n".format(self.state.wcsaxes_slice)
+        if isinstance(ref_coords, WCS):
+            script += f"ax.reset_wcs(slices={self.state.wcsaxes_slice}, wcs=ref_data.coords)\n"
```
astrofrog, line 229: In principle we should be able to pass any ``Coordinates`` subclass to ``reset_wcs`` - after all this is what we do in the actual image viewer:

```
        ref_coords = getattr(self.state.reference_data, 'coords', None)

        if ref_coords is None or isinstance(ref_coords, LegacyCoordinates):
            self.axes.reset_wcs(slices=self.state.wcsaxes_slice,
                                wcs=get_identity_wcs(self.state.reference_data.ndim))
        else:
            self.axes.reset_wcs(slices=self.state.wcsaxes_slice, wcs=ref_coords)
```

If we don't care about legacy coordinates in the script export, we might be able to get away with just doing:

```
self.axes.reset_wcs(slices=self.state.wcsaxes_slice, wcs=ref_coords)
```

if ``ref_coords`` is defined? (and not have the case below either)

We could test that this works using e.g. ``AffineCoordinates``.

astrofrog (changes requested):
Looks good but I think we could actually be more general as described below.

glue/viewers/image/viewer.py
dhomeier: > If we don't care about legacy coordinates in the script export, we might be able to get away with just doing:

Basically those are the `else` case already covered, but if all the rest can be treated the same way, it becomes much easier of course (I haven't seen an example using `coords.wcs` or `coords.wcs_dict`).
No idea if those `AffineCoordinates` make any sense at all, but they work – apart from the `_slice` and `simple_att` tests, that suffer from some fuzziness in the (lower axis) labels. Can probably live with that?
glue_plot
expected

glue/viewers/image/qt/tests/test_python_export.py
```diff
@@ -57,31 +87,70 @@ def test_simple_visual(self, tmpdir):
...
-        self.viewer.state.x_att = self.data.pixel_component_ids[1]
-        self.viewer.state.y_att = self.data.pixel_component_ids[0]
+    @pytest.mark.parametrize('coords', [None, 'wcs', 'affine'])
+    def test_slice(self, tmpdir, coords):
+        if coords is None:
+            data = self.data
+        else:
+            data = getattr(self, f'data_{coords}')
+            self.viewer.add_data(data)
+            self.viewer.remove_data(self.data)
```
astrofrog, line 97: I wonder if this could perhaps be implemented as a pytest fixture to avoid the repetition in each test?

glue/viewers/image/qt/tests/test_python_export.py
```diff
@@ -57,31 +87,70 @@ def test_simple_visual(self, tmpdir):
...
+
+        self.viewer.state.x_att = data.pixel_component_ids[1]
+        self.viewer.state.y_att = data.pixel_component_ids[0]
         self.viewer.state.slices = (2, 3, 4)
+        if coords == 'affine':
+            pytest.xfail('Known issue with axis label rendering')
```
astrofrog, line 103: Can you open an issue in the repo to keep track of this?

astrofrog (approved):
This looks good to me - see comment below about using a fixture to avoid all the boilerplate code added at the start of each test. If you think this can be done, feel free to do it, otherwise you can just merge.

glue/viewers/image/qt/tests/test_python_export.py
dhomeier: @astrofrog I implemented this as a helper function now, but could not figure out yet how to use it as a fixture in light of https://docs.pytest.org/en/stable/deprecations.html#calling-fixtures-directly (or if that would actually be helpful).

astrofrog:
Thanks for adding the helper method! That'll be fine for now, no need to make it a fixture.

PR #2338, opened 2022-10-18
Add PyQt6 and PySide6 status to `glue-deps` information

dhomeier:
## Description
With support for both backends now functional adding them to `gui_framework`.
Also updating string formatting (probably don't really need most of the `{:s}` formats) and adding info to make `glue-deps info Python` work.

### DIFF
--- a/glue/_deps.py
+++ b/glue/_deps.py
@@ -56,12 +56,12 @@ PIP package name:
 
     def __str__(self):
         if self.installed:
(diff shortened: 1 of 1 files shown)


astrofrog (changes requested):
Looks good, but there are three tests that need updating

dhomeier:
I think the tests were correct. Ah, the subtle details of f-string aligment!

PR #2343, opened 2022-10-27
Performance improvements for CompositeArray

astrofrog:
This attempts to speed up CompositeArray by using several tricks:

* When using colormaps, we don't need to care about layers below the top-most layer that has alpha=1 since it will block everything below - this should speed up viewers with multiple images open.
* When using colormaps and assuming the colormaps don't have any transparency, we can avoid a bunch of temporary array creations.
* We make use of out= in the contrast/bias and stretch to avoid more temporary array creation.

### DIFF
--- a/glue/viewers/image/composite_array.py
+++ b/glue/viewers/image/composite_array.py
@@ -12,6 +12,8 @@ __all__ = ['CompositeArray']
 
 COLOR_CONVERTER = ColorConverter()
 
+CMAP_SAMPLING = np.linspace(0, 1, 256)
+
 STRETCHES = {
     'linear': LinearStretch,
     'sqrt': SqrtStretch,
@@ -80,7 +82,35 @@ class CompositeArray(object):
         img = None
         visible_layers = 0
 
-        for uuid in sorted(self.layers, key=lambda x: self.layers[x]['zorder']):
+        # We first check that layers are either all colormaps or all single
+        # colors, and at the same time, if we use colormaps we check whether
+        # any of the colormaps have non-1 values in the alpha.
+
+        minimum_cmap_alpha = 1
+
+        mode = None
+        for layer in self.layers.values():
+            if mode is None:
+                if isinstance(layer['color'], Colormap):
+                    if mode is None:
+                        mode = 'colormap'
+                    elif mode == 'color':
+                        raise Exception("Mixed color and colormaps in CompositeArray")
+                    else:
+                        minimum_cmap_alpha = min(layer['color'](CMAP_SAMPLING)[:, 3].min(),
+                                                 minimum_cmap_alpha)
+                else:
+                    if mode is None:
+                        mode = 'color'
+                    elif mode == 'colormap':
+                        raise Exception("Mixed color and colormaps in CompositeArray")
+
+        sorted_uuids = sorted(self.layers, key=lambda x: self.layers[x]['zorder'])
+
+        if mode == 'colormap' and self.layers[sorted_uuids[-1]]['alpha'] == 1:
+            sorted_uuids = sorted_uuids[-1:]
+
+        for uuid in sorted_uuids:
 
             layer = self.layers[uuid]
 
@@ -104,10 +134,13 @@ class CompositeArray(object):
             else:
                 scalar = False
 
-            data = STRETCHES[layer['stretch']]()(contrast_bias(interval(array)))
+            stretch = STRETCHES[layer['stretch']]()
+            data = interval(array)
+            data = contrast_bias(data, out=data)
+            data = stretch(data, out=data)
             data[np.isnan(data)] = 0
 
-            if isinstance(layer['color'], Colormap):
+            if mode == 'colormap':
 
                 if img is None:
                     img = np.ones(data.shape + (4,))
@@ -115,16 +148,27 @@ class CompositeArray(object):
                 # Compute colormapped image
                 plane = layer['color'](data)
 
-                alpha_plane = layer['alpha'] * plane[:, :, 3]
+                if plane[:, :, 3].min() == 1.:
 
-                # Use traditional alpha compositing
-                plane[:, :, 0] = plane[:, :, 0] * alpha_plane
-                plane[:, :, 1] = plane[:, :, 1] * alpha_plane
-                plane[:, :, 2] = plane[:, :, 2] * alpha_plane
+                    if layer['alpha'] == 1:
+                        img[...] = 0
+                    else:
+                        plane *= layer['alpha']
+                        img *= (1 - layer['alpha'])
 
-                img[:, :, 0] *= (1 - alpha_plane)
-                img[:, :, 1] *= (1 - alpha_plane)
-                img[:, :, 2] *= (1 - alpha_plane)
+                else:
+
+                    alpha_plane = layer['alpha'] * plane[:, :, 3]
+
+                    # Use traditional alpha compositing
+                    plane[:, :, 0] = plane[:, :, 0] * alpha_plane
+                    plane[:, :, 1] = plane[:, :, 1] * alpha_plane
+                    plane[:, :, 2] = plane[:, :, 2] * alpha_plane
+
+                    img[:, :, 0] *= (1 - alpha_plane)
+                    img[:, :, 1] *= (1 - alpha_plane)
+                    img[:, :, 2] *= (1 - alpha_plane)
+
                 img[:, :, 3] = 1
 
             else:
--- a/glue/viewers/image/composite_array.py
+++ b/glue/viewers/image/composite_array.py
@@ -83,7 +83,7 @@ class CompositeArray(object):
         visible_layers = 0
 
         # We first check that layers are either all colormaps or all single
-        # colors, and at the same time, if we use colormaps we ccheck what
+        # colors, and at the same time, if we use colormaps we check what
         # the smallest alpha value is - if it is 1 then colormaps do not change
         # transparency and this allows us to speed things up a little.
 
--- a/glue/viewers/image/composite_array.py
+++ b/glue/viewers/image/composite_array.py
@@ -110,13 +110,12 @@ class CompositeArray(object):
         # affect the output, assuming also that the colormaps do not change the
         # alpha
         if self.mode == 'colormap':
-            start = 0
-            for i in range(len(sorted_uuids)):
+            for i in range(len(sorted_uuids) - 1, -1, -1):
                 layer = self.layers[sorted_uuids[i]]
                 if layer['visible']:
-                    if layer['alpha'] * layer['cmap'](CMAP_SAMPLING)[:, 3].min() == 1:
-                        start = i
-            sorted_uuids = sorted_uuids[start:]
+                    if layer['alpha'] == 1 and layer['cmap'](CMAP_SAMPLING)[:, 3].min() == 1:
+                        sorted_uuids = sorted_uuids[i:]
+                        break
 
         for uuid in sorted_uuids:
 
--- a/glue/viewers/image/composite_array.py
+++ b/glue/viewers/image/composite_array.py
@@ -101,20 +101,10 @@ class CompositeArray(object):
         img = None
         visible_layers = 0
 
-        # We first check that layers are either all colormaps or all single
-        # colors, and at the same time, if we use colormaps we check what
-        # the smallest alpha value is - if it is 1 then colormaps do not change
-        # transparency and this allows us to speed things up a little.
-
-        minimum_cmap_alpha = 1
-        if self.mode == 'colormap':
-            for layer in self.layers.values():
-                minimum_cmap_alpha = min(layer['cmap'](CMAP_SAMPLING)[:, 3].min(),
-                                            minimum_cmap_alpha)
-
         # Get a sorted list of UUIDs with the top layers last
         sorted_uuids = sorted(self.layers, key=lambda x: self.layers[x]['zorder'])
 
+        # We first check that layers are either all colormaps or all single colors.
         # In the case where we are dealing with colormaps, we can start from
         # the last layer that has an opacity of 1 because layers below will not
         # affect the output, assuming also that the colormaps do not change the
@@ -166,7 +156,11 @@ class CompositeArray(object):
                 # Compute colormapped image
                 plane = layer['cmap'](data)
 
-                if minimum_cmap_alpha == 1.:
+                # Check what the smallest colormap alpha value for this layer is
+                # - if it is 1 then this colormap does not change transparency,
+                # and this allows us to speed things up a little.
+
+                if layer['cmap'](CMAP_SAMPLING)[:, 3].min() == 1:
 
                     if layer['alpha'] == 1:
                         img[...] = 0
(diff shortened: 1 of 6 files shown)


glue/viewers/image/composite_array.py
```diff
@@ -80,7 +82,44 @@ def __call__(self, bounds=None):
         img = None
         visible_layers = 0
 
-        for uuid in sorted(self.layers, key=lambda x: self.layers[x]['zorder']):
+        # We first check that layers are either all colormaps or all single
+        # colors, and at the same time, if we use colormaps we ccheck what
```
pllim, line 86:
```suggestion
        # colors, and at the same time, if we use colormaps we check what
```

glue/viewers/image/composite_array.py
```diff
@@ -80,7 +82,44 @@ def __call__(self, bounds=None):
...
+
+        minimum_cmap_alpha = 1
+        mode = None
+
+        for layer in self.layers.values():
+            if isinstance(layer['color'], Colormap):
+                if mode is None:
+                    mode = 'colormap'
+                elif mode == 'color':
+                    raise TypeError("Mixed color and colormaps in CompositeArray")
```
pllim, line 98: How often would someone encounter this now that it throws exception? I don't think there is a mechanism to prevent user from setting the first viewer layer to colormap but the second layer to monochromatic?

Also, are things like subsets and linked tables already excluded when this logic is invoked? I am not familiar with the internals here.

duytnguyendtn (commented):
Setting the contrast/bias via API in `jdaviz` is much improved with this PR. Tested with one image, the original performance for 200 random adjustments for contrast/bias was ~40 seconds, down to ~19 seconds with https://github.com/glue-viz/glue-jupyter/pull/325, now to ~9 seconds with this PR!

I did run into an issue with two images: if I test the contrast/bias sliders and then try to blink, the first image seems to be blank for me. Steps to reproduce:

1. Load BOTH datasets in https://github.com/spacetelescope/jdaviz/blob/main/notebooks/ImvizDitheredExample.ipynb
2. Change Plot Options dataset to the second (since there are two datasets, the second layer (B) is visible on top. By default, Plot options iterates on the first dataset (A).)
3. Fiddle with contrast/bias values via API. I used the following loop:

```
def test():
    for i in range(0,100):
        val = np.random.randint(0,100)
        imviz.plugins['Plot Options']._obj.image_bias_value = val/100
        imviz.plugins['Plot Options']._obj.image_contrast_value = (1.0 - val/100)*4
%prun test() 
```

4. Blink image using "B" key to show the first dataset (A) on top. The image is blank for me, even after trying to interact with the image layer via pan/zoom. Doesn't seem to redraw

astrofrog:
Ah yes that should be an easy fix

astrofrog:
@duytnguyendtn - could you try the blinking again?

duytnguyendtn:
Yup, that fixed it! Thanks @astrofrog!

glue/viewers/image/composite_array.py
astrofrog: Glue doens't actually allow image viewers to have mixed layers - the mode is controlled by:

https://github.com/glue-viz/glue/blob/e4e875a7a54c4925db244f62f7f16644578617dd/glue/viewers/image/state.py#L89-L92

In fact I should make CompositeArray have such a settable mode rather than checking it like this at drawing time - but that can be cleaned up/improved in future. For now, these exceptions should never be tripped by users and if they are it indicates a bug in the image viewers.

kecnry: The speed up is looking very promising!  But I am seeing this error raised when adding data to the viewer if the viewer itself is visible.  Let me know if you need the full traceback or more details to reproduce.

astrofrog: Could you provide the full traceback?

kecnry:
```
File ~/repos/jdaviz/jdaviz/app.py:1380, in Application.set_data_visibility(self, viewer_reference, data_label, visible, replace)
   1378 [data] = [x for x in self.data_collection if x.label == data_label]
   1379 color_cycler = ColorCycler()
-> 1380 viewer.add_data(data, percentile=95)
   1382 add_data_message = AddDataMessage(data, viewer,
   1383                                   viewer_id=viewer_id,
   1384                                   sender=self)
   1385 self.hub.broadcast(add_data_message)

File /opt/miniconda3/envs/jdaviz-dev/lib/python3.8/site-packages/glue_jupyter/view.py:117, in IPyWidgetView.add_data(self, data, color, alpha, **layer_state)
    113 def add_data(self, data, color=None, alpha=None, **layer_state):
    115     data = validate_data_argument(self._data, data)
--> 117     result = super().add_data(data)
    119     if not result:
    120         return

File ~/repos/glue/glue/viewers/common/viewer.py:218, in Viewer.add_data(self, data)
    216 with ignore_callback(layer.state, 'zorder'):
    217     self._layer_artist_container.append(layer)
--> 218 layer.update()
    219 self.draw_legend()  # need to be called here because callbacks are ignored in previous step
    221 # Add existing subsets to viewer

File ~/repos/glue/glue/utils/matplotlib.py:181, in defer_draw.<locals>.wrapper(*args, **kwargs)
    177 @wraps(func)
    178 def wrapper(*args, **kwargs):
    180     if len(DEFER_DRAW_BACKENDS) == 0:
--> 181         return func(*args, **kwargs)
    183     # Don't recursively defer draws. We just check the first draw_idle
    184     # method since all should be modified in sync.
    185     if isinstance(DEFER_DRAW_BACKENDS[0].draw_idle, DeferredMethod):

File ~/repos/glue/glue/viewers/image/layer_artist.py:202, in ImageLayerArtist.update(self, *event)
    200 ARRAY_CACHE.pop(self.state.uuid, None)
    201 PIXEL_CACHE.pop(self.state.uuid, None)
--> 202 self._update_image(force=True)
    203 self.redraw()

File /opt/miniconda3/envs/jdaviz-dev/lib/python3.8/site-packages/glue_jupyter/bqplot/image/layer_artist.py:127, in BqplotImageLayerArtist._update_image(self, force, **kwargs)
    123 changed = self.pop_changed_properties()
    125 if force or any(prop in changed for prop in ('layer', 'attribute',
    126                                              'slices', 'x_att', 'y_att')):
--> 127     self._update_image_data()
    128     force = True  # make sure scaling and visual attributes are updated
    130 if force or any(prop in changed for prop in ('v_min', 'v_max', 'contrast',
    131                                              'bias', 'alpha', 'color_mode',
    132                                              'cmap', 'color', 'zorder',
    133                                              'visible', 'stretch',
    134                                              'bitmap_visible', 'contour_visible')):

File /opt/miniconda3/envs/jdaviz-dev/lib/python3.8/site-packages/glue_jupyter/bqplot/image/layer_artist.py:140, in BqplotImageLayerArtist._update_image_data(self, *args, **kwargs)
    139 def _update_image_data(self, *args, **kwargs):
--> 140     self.composite_image.invalidate_cache()
    141     # if the image data change, the contour lines are invalid
    142     self._contour_line_cache.clear()

File /opt/miniconda3/envs/jdaviz-dev/lib/python3.8/site-packages/glue_jupyter/bqplot/image/frb_mark.py:107, in FRBImage.invalidate_cache(self)
    106 def invalidate_cache(self):
--> 107     self.update()

File /opt/miniconda3/envs/jdaviz-dev/lib/python3.8/site-packages/glue_jupyter/bqplot/image/frb_mark.py:95, in FRBImage.update(self, *args, **kwargs)
     92 bounds = [(ymin, ymax, ny), (xmin, xmax, nx)]
     94 # Get the array and assign it to the artist
---> 95 image = self.array_maker(bounds=bounds)
     96 if image is not None:
     97     if image.dtype == np.dtype("float64"):

File ~/repos/glue/glue/viewers/image/composite_array.py:106, in CompositeArray.__call__(self, bounds)
    104             mode = 'color'
    105         elif mode == 'colormap':
--> 106             raise TypeError("Mixed color and colormaps in CompositeArray")
    108 # Get a sorted list of UUIDs with the top layers last
    109 sorted_uuids = sorted(self.layers, key=lambda x: self.layers[x]['zorder'])

TypeError: Mixed color and colormaps in CompositeArray
```

pllim: Aww, man.

`viewer.add_data(data, percentile=95)` -- I added the `percentile` in https://github.com/spacetelescope/jdaviz/pull/1386 but the `add_data` call has always been there.

kecnry: we now pass a default `color` as well, but I temporarily removed it locally in case that was triggering the error, but that doesn't seem to be the case.

astrofrog:
@kecnry - could you try the latest version of this PR along with https://github.com/glue-viz/glue-jupyter/pull/331 to see if the error you were seeing goes away?

@duytnguyendtn - could you test this again as there are some non-trivial changes that I have made?

kecnry:
The error does seem to have gone away, but am now seeing the deprecation warning about color vs cmap and the behavior seems to have changed with opaque layers that are set to use colormaps (it seems the layers are now "adding" instead of only showing the top layer).  Is there something else on our end we need to change because of this deprecation?

astrofrog:
@kecnry - just to check, are you using the PR for glue-jupyter mentioned above? If so I will double check if I missed any other calls that need to be updated.

astrofrog:
@pllim - no it is from this PR.

kecnry:
Ah, something happened with the dev install and it wasn't updating.  I forced to re-install and now that went away - sorry about the false alarm!  The errors do go away now and everything I've quickly tested so far seems to work as expected, but I'll keep testing.

astrofrog:
@kecnry - ok thanks! I will leave this open a bit and merge later today if no further issues.

dhomeier:
Will changing `test_pv_slicer.py` to use `self.w._composite.layers['image']['cmap']` do for fixing the tests?

astrofrog:
Yes that will probably do 1 would you've happy to push the change to this PR?

astrofrog:
@dhomeier did you mean to push all these commits? (Np if so, just checking!)

dhomeier:
Sorry for messing up the commit history; perhaps should have rebased on this branch rather than main!
I've also tested against latest glue-jupyter, maybe would be good to add a `dev` test suite there as well to test after merging this PR.

dhomeier:
Oh, maybe I should have rebased on `main` (again) after pulling this branch.

astrofrog:
Could you still do the rebase? Thanks!

dhomeier:
OK, a number of warnings about skipping previously applied commits locally, but that seems to have done it.

glue/viewers/image/composite_array.py
```diff
@@ -80,7 +101,34 @@ def __call__(self, bounds=None):
...
+                                            minimum_cmap_alpha)
+
+        # Get a sorted list of UUIDs with the top layers last
+        sorted_uuids = sorted(self.layers, key=lambda x: self.layers[x]['zorder'])
+
+        # In the case where we are dealing with colormaps, we can start from
+        # the last layer that has an opacity of 1 because layers below will not
+        # affect the output, assuming also that the colormaps do not change the
+        # alpha
+        if self.mode == 'colormap' and minimum_cmap_alpha == 1.:
```
dhomeier, line 122:
```suggestion
        if self.mode == 'colormap' and minimum_cmap_alpha == 1.:
            sorted_uuids = sorted_uuids[-1:]
        elif self.mode == 'colormap':
```
Not sure if I understand the combination of layers right, but if _all_ of them have alpha=1, wouldn't you automatically see only the last, topmost one? Whereas for the case that _some_ are transparent, still everything below the last opaque one could be ignored?
The first case could probably be caught even earlier by just picking the layer with highest `self.layers[x]['zorder']`, avoiding further creating of temporary arrays.

dhomeier: @astrofrog I've added a test with a faded colormap and the use of layer-minimum alpha above – not sure if it will make a relevant difference, but I can push the commit if you want to review.

astrofrog: > Still wondering if it would not suffice for the last layer of opacity 1, if that layer has minimum cmap_alpha of 1, to ignore all layers below.

Is that not effectively what we are doing now?

dhomeier: If any layer has a cmap_alpha < 1 (`minimum_cmap_alpha ` < 1), we're alway keeping the full `sorted_uuids` list. Instead testing for `if layer['alpha'] * layer['cmap'](CMAP_SAMPLING)[:, 3].min() == 1` we may still truncate it at that layer (which works at least for my two-layer test).

dhomeier: It _might_ actually work with computing `plane` further down as well (does at least with that same test), but that may need some more testing.

astrofrog: Feel free to push the commit so I can make sure I understand :)

dhomeier: I've pushed the first change to just set `sorted_uuids[start:]` separately so they are easier to revert.
This is still not really testing cmap alpha values _between_ 0 and 1.

glue/viewers/image/composite_array.py
```diff
@@ -104,27 +143,45 @@ def __call__(self, bounds=None):
...
+                else:
+
+                    # Use traditional alpha compositing
+
+                    alpha_plane = layer['alpha'] * plane[:, :, 3]
 
-                alpha_plane = layer['alpha'] * plane[:, :, 3]
+                    plane[:, :, 0] = plane[:, :, 0] * alpha_plane
+                    plane[:, :, 1] = plane[:, :, 1] * alpha_plane
+                    plane[:, :, 2] = plane[:, :, 2] * alpha_plane
```
dhomeier, line 179:
```suggestion
                    plane[:, :, :3] *= np.atleast3d(alpha_plane)
```
Probably not necessarily faster with the bit of extra vectorisation – doing the channels separately could in fact be more memory-efficient.

dhomeier (approved):
With or without either of my last two commits, should be ready to merge.

glue/viewers/image/composite_array.py
```diff
@@ -80,7 +101,24 @@ def __call__(self, bounds=None):
...
+        # affect the output, assuming also that the colormaps do not change the
+        # alpha
+        if self.mode == 'colormap':
+            start = 0
+            for i in range(len(sorted_uuids)):
+                layer = self.layers[sorted_uuids[i]]
+                if layer['visible']:
+                    if layer['alpha'] * layer['cmap'](CMAP_SAMPLING)[:, 3].min() == 1:
+                        start = i
+            sorted_uuids = sorted_uuids[start:]
```
astrofrog, line 119: Would it maybe be more efficient to loop over the layers backwards by inverse zorder?

astrofrog (commented):
Looks good to me! See one comment below in case it would be easy for you to implement but otherwise feel free to merge.

glue/viewers/image/composite_array.py
dhomeier: Good point, yes we can just stop at the highest opaque layer.

dhomeier:
I've tried @duytnguyendtn's test above, and everything looks ok there. Might be a tiny speedup (from 14 to 15 ms per `CompositeArray` call) from the combined last 4 or 5 commits.

PR #2345, opened 2022-12-13
Fix world coordinates for 1D WCS

astrofrog:
Found this while working on the tick labels in profile viewer.

### DIFF
--- a/glue/core/coordinate_helpers.py
+++ b/glue/core/coordinate_helpers.py
@@ -63,9 +63,8 @@ def pixel2world_single_axis(wcs, *pixel, world_axis=None):
         result = result.reshape(pixel_shape)
     else:
         result = wcs.pixel_to_world_values(*pixel)
-
-    if len(pixel) > 1:
-        result = result[world_axis]
+        if len(pixel) > 1:
+            result = result[world_axis]
 
     return np.broadcast_to(result, original_shape)
 
@@ -121,9 +120,8 @@ def world2pixel_single_axis(wcs, *world, pixel_axis=None):
         result = result.reshape(world_shape)
     else:
         result = wcs.world_to_pixel_values(*world)
-
-    if len(world) > 1:
-        result = result[pixel_axis]
+        if len(world) > 1:
+            result = result[pixel_axis]
 
     return np.broadcast_to(result, original_shape)
 
--- a/glue/core/coordinate_helpers.py
+++ b/glue/core/coordinate_helpers.py
@@ -55,7 +55,7 @@ def pixel2world_single_axis(wcs, *pixel, world_axis=None):
 
     result = wcs.pixel_to_world_values(*pixel)
 
-    if len(pixel) > 1:
+    if len(pixel) > 1 or np.ndim(result) > 1:
         result = result[world_axis]
 
     return broadcast_to(result, original_shape)
--- a/glue/core/coordinate_helpers.py
+++ b/glue/core/coordinate_helpers.py
@@ -59,8 +59,7 @@ def pixel2world_single_axis(wcs, *pixel, world_axis=None):
 
     if len(pixel) == 1 and pixel[0].ndim > 1:
         pixel_shape = pixel[0].shape
-        pixel = [pixel[0].ravel()]
-        result = wcs.pixel_to_world_values(*pixel)
+        result = wcs.pixel_to_world_values(pixel[0].ravel())
         result = result.reshape(pixel_shape)
     else:
         result = wcs.pixel_to_world_values(*pixel)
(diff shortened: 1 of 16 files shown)


dhomeier:
@astrofrog some cases with `result` shapes like `(1, N)` needed an extra check to pass.
`broadcast_to` should have outlived its existence for some time now, although since it's not strictly private, there's perhaps some small risk to break something downstream (I did test glue-jupyter and glue-astronomy against this branch).

astrofrog:
@dhomeier - I had to revert https://github.com/glue-viz/glue/pull/2345/commits/a1d114c3ec11f99478189abaf162b96684cef7c7 in my latest commit - what use case required your fix?

glue/core/coordinate_helpers.py
```diff
@@ -53,9 +53,22 @@ def pixel2world_single_axis(wcs, *pixel, world_axis=None):
...
-    return broadcast_to(result[world_axis], original_shape)
+    if len(pixel) == 1 and pixel[0].ndim > 1:
+        pixel_shape = pixel[0].shape
+        pixel = [pixel[0].ravel()]
+        result = wcs.pixel_to_world_values(*pixel)
+        result = result.reshape(pixel_shape)
+    else:
+        result = wcs.pixel_to_world_values(*pixel)
+
+    if len(pixel) > 1:
```
dhomeier, line 68:
```suggestion
    if len(pixel) > 1 or np.shape(result)[0] == 1:
```
This fixes all tests for me. Of course this implies `world_axis == 0`, not sure if that should be explicitly tested as well or will always be satisfied...

astrofrog: I think this is not the correct fix - I will push another commit up shortly (short version: some of our coordinate classes were not following APE 14 strictly)

astrofrog:
Still a WIP, but getting closer

astrofrog:
@dhomeier the remaining issue was that IdentityCoordinates, for single coordinates, should return a scalar or array, not a tuple with one element (this is prescribed by the APE 14 API)

dhomeier:
Yes, that seemed weird, but it wasn't clear to me that came from inside glue.

dhomeier:
But are you sure it is always a tuple? The `self.pixel_n_dim == 1` case still does not seem to catch it.

astrofrog:
> But are you sure it is always a tuple?

In principle APE 14 mandates that a tuple is returned if there are >coordinates. But there may still be cases where this doesn't happen?

> The self.pixel_n_dim == 1 case still does not seem to catch it.

Is there a test failure still where this is not caught? Or can you come up with a failing example?

dhomeier:
> Is there a test failure still where this is not caught? Or can you come up with a failing example?

Strangely it seems to pass in most of the py310 envs, but none of the others. But on my Mac it's passing Pythons 3.9-3.11... Very weird.

astrofrog:
I think this should be fixed - it was an issue with LegacyCoordinates. Not quite sure why it would fail only on some Python versions though... Let's see if the CI is happy.

pllim:
I think this broke Jdaviz, see log at https://github.com/spacetelescope/jdaviz/actions/runs/3954350017/jobs/6771598108

`ValueError: input operand has more dimensions than allowed by the axis remapping`

astrofrog:
Investigating!

dhomeier:
Seems they are all failing on `glue_astronomy.spectral_coordinates.SpectralCoordinates` instances of shape `(1, N)` which are using their own `pixel_to_world_values` method (even if they were subclassed on `LegacyCoordinates`).

dhomeier:
> In principle APE 14 mandates that a tuple is returned if there are >coordinates. But there may still be cases where this doesn't happen?
> 
Does this mean it must not return a tuple for the single coordinate case? Specifically should `SpectralCoordinates` with `n_dim=1` always return `np.interp(...)[0]` or `tuple(np.interp(...)[0])`?
Either change does fix the `jdaviz` tests.

pllim:
https://github.com/glue-viz/glue-astronomy/pull/82

pllim:
glue-core needs a release. We see failure in Jdaviz now because glue-astronomy is released but not this PR.

dhomeier:
These are the updates currently in
### Bug Fixes
* More fixes for compatibility with Qt6 by @astrofrog in https://github.com/glue-viz/glue/pull/2349
* Fix world coordinates for 1D WCS by @astrofrog in https://github.com/glue-viz/glue/pull/2345
### Other Changes
* API update to numpy 1.24 by @dhomeier in https://github.com/glue-viz/glue/pull/2346

#### Still open
+ #2348 that might also go in.

But in addition there is a new regression in 
+ #2353. As that only concerns numpy-dev, it might not be as urgent yet.

PR #2346, opened 2022-12-20
API update to numpy 1.24

dhomeier:
## Description
Fixing a number of failures that popped up after 1.24.0 release, mostly due to use of the deprecated `np.bool`, `np.int` etc. dtype attributes.
Also bumping `echo` to >= 0.6 to resolve `_widget.setValue()` errors.


dhomeier:
Checking for numpy `AttributeError` primarily for `ASTRODENDRO_INSTALLED`; this means no released astrodendro is available now anywhere with Python >= 3.9 or numpy >= 1.24. Should we add a test against the GitHub dev, or are there any plans for bringing out a new astrodendro release, @astrofrog?

PR #2348, opened 2023-01-09
Relax longitude range in fullsphere projections

Carifio24:
As noticed by Alyssa and Theo, longitude angles in the full-sphere projections of the scatter viewer are currently restricted to [-π, π]. This PR relaxes this restriction by mapping the longitude component into this range as appropriate (for example, 3π/2 is mapped to -π/2).

Additionally, as longitude is defined differently for Earth and sky, this PR adds the ability to reverse the full-sphere axes. Matplotlib doesn't let us change bounds in these projections, so this PR mimics this adding a "flip" option in the UI which negates the relevant values. The intended use case here is for longitude, but I don't see a reason not to offer the same option for latitude as well. The buttons in the UI that allow this flipping are only visible when in a full-sphere projection.

### DIFF
--- a/glue/viewers/scatter/qt/tests/test_python_export.py
+++ b/glue/viewers/scatter/qt/tests/test_python_export.py
@@ -1,3 +1,5 @@
+from itertools import product
+
 import numpy as np
 import matplotlib.pyplot as plt
 from astropy.utils import NumpyRNGContext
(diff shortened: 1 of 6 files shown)


Carifio24:
@astrofrog @dhomeier I saw there were some recent CI updates so I've rebased this (and 2350) onto the main branch.

glue/viewers/scatter/qt/tests/test_python_export.py
```diff
@@ -260,25 +261,17 @@ def test_full_sphere_degrees(self, tmpdir):
...
-        self.assert_same(tmpdir)
-        self.viewer.state.plot_mode = 'lambert'
-        self.viewer.state.x_att = self.data.id['g']
-        self.viewer.state.y_att = self.data.id['h']
-        self.assert_same(tmpdir)
-        self.viewer.state.plot_mode = 'mollweide'
-        self.viewer.state.x_att = self.data.id['a']
-        self.viewer.state.y_att = self.data.id['b']
-        self.assert_same(tmpdir)
+        self._fullsphere_common(tmpdir)
```
dhomeier, line 274: Why are the other projections not tested anymore? Maybe you could make `plot_mode` an input parameter to `_fullsphere_common`?

Carifio24: Oops, removing the other projections was not intentional. Will add a parameter as you suggest and check each projection

Carifio24: Looks like I didn't forget the other projections after all - `_fullsphere_common` checks each plot type

Carifio24:
I'll admit I don't know of an intended use case for flipping the latitude - was purely thinking of flexibility

glue/viewers/scatter/qt/tests/test_python_export.py
dhomeier: Ah sorry, I only looked at the code as far as rendered in the diff!

dhomeier (approved):
Thanks, looks all good – I hope the descriptions are self-explanatory enough.

glue/viewers/scatter/qt/tests/test_python_export.py
Carifio24: So did I...I forgot I'd covered all the projections until I opened up my editor 😆

glue/viewers/scatter/state.py
```diff
@@ -29,6 +29,8 @@ class ScatterViewerState(MatplotlibDataViewerState):
     dpi = DDCProperty(72, docstring='The resolution (in dots per inch) of density maps, if present')
     plot_mode = DDSCProperty(docstring="Whether to plot the data in cartesian, polar or another projection")
     angle_unit = DDSCProperty(docstring="Whether to use radians or degrees for any angular coordinates")
+    flip_lonaxis = DDCProperty(False, docstring="Whether or not to flip x coordinates; used for full-sphere projections")
+    flip_lataxis = DDCProperty(False, docstring="Whether or not to flip y coordinates; used for full-sphere projections")
```
astrofrog, line 33: Just to make sure i understand, how is using these options different to setting an ``x_min`` value larger than ``x_max``? If it's different, is there any scope to give these more general names? (e.g. ``flip_x`` and ``flip_y``)

Carifio24: Using these has the same effect as `x_min` > `x_max`, in that the axis will run in the opposite direction to usual. I added these in because matplotlib doesn't allow us to change bounds for the full-sphere projections. For the other projections one can just use the bounds instead. But maybe `flip` names are more future-proof, in case we someday have other bound-less projections?

astrofrog: I wonder if we could do this without adding new callback properties - what if we just use x_min and x_max but at the point of setting the limits, if one can't actually set the full limits we would still use the order of x_min/x_max to determine whether to flip the limits in the plot. The benefit of this is that if a user uses the flip buttons in the existing axes limits in the UI it should still work?

Carifio24:
@astrofrog I've changed the axis-flipping to work essentially how you had suggested. The flip button is now active in full-sphere mode. Using this changes `x_min` and `x_max`, but just doesn't pass this along to matplotlib.

Since doing this "flip" involves redrawing the data in this case, I've added some logic to the layer artist to redraw when bounds are changed and we're in a full-sphere projection.

astrofrog (approved):
Looks great, thanks!

PR #2350, opened 2023-01-19
Allow specifying visual attributes and colormap when faceting subsets

Carifio24:
This PR adds arguments to `facet_subsets` to allow directly specifying subset styling and a colormap from that function call (as opposed to calling `colorize_subsets` after, as the dialog currently does. This PR also updates the subset faceting dialog to use the updated function signature, though it doesn't expose the styling options to the UI (though there's no reason that we couldn't).

I think there are a few use cases here. Subset faceting via the dialog should see a slight performance increase with a large number of subsets, since we no longer need another pass through to update colors.  Dashboard-style applications could similarly benefit from this to avoid the quick re-rendering that inspired augmenting `new_subset` and `new_subset_group`. I could also see a tool or plugin that does automatic faceting benefiting from this as well.

### DIFF
--- a/glue/core/tests/test_util.py
+++ b/glue/core/tests/test_util.py
@@ -3,8 +3,8 @@
 import numpy as np
 from matplotlib.cm import gray
 
-from .. import Data, DataCollection
-from ..util import facet_subsets, colorize_subsets
+from .. import Data, DataCollection, VisualAttributes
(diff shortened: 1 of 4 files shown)


astrofrog (approved):
Looks great, thanks!

PR #2354, opened 2023-01-24
Add traverse method

javerbukh:
# Pull Request Template

## Description
Example of what a traverse() method could look like in the subset file so that we could get a tree with readable values for subset bounds. This does not include the work it would take to resolve the operations throughout the tree.

### DIFF
--- a/glue/core/subset.py
+++ b/glue/core/subset.py
@@ -797,6 +797,9 @@ class RangeSubsetState(SubsetState):
     def copy(self):
         return RangeSubsetState(self.lo, self.hi, self.att)
 
+    def traverse(self):
+        return self.lo, self.hi
+
 
 class MultiRangeSubsetState(SubsetState):
     """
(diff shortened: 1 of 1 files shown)


glue/core/subset.py
```diff
@@ -797,6 +797,9 @@ def to_mask(self, data, view=None):
     def copy(self):
         return RangeSubsetState(self.lo, self.hi, self.att)
 
+    def traverse(self):
```
pllim, line 800: Hmm, I am a little surprised this is needed. The `__str__` on L1098 also traverse in a similar way but I don't see any custom `__str__` defined here.

dhomeier:
Before the next commit, could you rebase? Need a pin to the maximum numpy revision for the `dev` jobs, otherwise they will hang for hours in the CI.

javerbukh:
Looks like this implementation needs to happen in jdaviz, closing this ticket.

PR #2355, opened 2023-01-30
Move PySide builds to allowed failures

astrofrog:
Not quite sure why this is happening but it seems to be pretty systematic

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -41,7 +41,6 @@ jobs:
         # Linux builds - test on all supported PyQt5/6 and PySide2/6 versions,
         # and include all dependencies in some builds
         - linux: py38-test-pyqt514-all
-        - linux: py38-test-pyside514
         - linux: py39-test-pyqt515
         - linux: py39-test-pyside515
         - linux: py310-test-pyside63
@@ -86,6 +85,10 @@ jobs:
           - enchant
           - hdf5
       envs: |
+
+        # Segmentation fault on test suite exit
+        - linux: py38-test-pyside514
+
         # PySide6 6.4 failures due to https://github.com/spyder-ide/qtpy/issues/373
         # and https://github.com/matplotlib/matplotlib/issues/24155
         # Python 3.11.0 failing on Windows in test_image.py on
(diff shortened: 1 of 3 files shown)


astrofrog:
Hmm some other PySide builds are running into issues now

astrofrog:
@dhomeier - I'm moving all the PySide builds to allowed failures because they are failing non-deterministically with a segmentation fault. If you have any ideas what this could be due to, let me know! But for now I need to move these otherwise the failures interfere with the auto release workflow.

.github/workflows/ci_workflows.yml
```diff
@@ -86,6 +83,12 @@ jobs:
           - enchant
           - hdf5
       envs: |
+
+        # Non-deterministic segmentation faults
```
dhomeier, line 87:
```suggestion
        # Non-deterministic QThread exceptions
```
Those, and the macos-pyside63 one, are actually all failing on the
> QThread: Destroyed while thread is still running
test return, but that seems to be non-deterministic nonetheless (timing issue?)

dhomeier (approved):
Just trying to keep in mind one has to look for `QThread` calls, but the solution for the moment would be just the same. On macOS Ventura PySide 6.3 seems barely supported at all, segfaulting as soon as I try to start the tests. Would perhaps be more productive to look into the `layerTree` failure with PySide 6.4.

doc/conf.py
```diff
@@ -139,7 +139,8 @@
                   ('py:class', 'glue.viewers.image.layer_artist.RGBImageLayerBase'),
                   ('py:class', 'glue.viewers.image.state.BaseImageLayerState'),
                   ('py:class', 'sip.voidptr'),
-                  ('py:class', 'PyQt5.sip.voidptr')]
+                  ('py:class', 'PyQt5.sip.voidptr'),
+                  ('py:class', 'PYQT_SLOT')]
```
dhomeier, line 143:
```suggestion
                  ('py:class', 'QKeySequence.StandardKey')],
                  ('py:class', 'PYQT_SLOT')]
```
Another one, in `BasicToolbar.grabShortcut`?

PR #2356, opened 2023-01-31
Represent `None` `component.units` as empty strings

dhomeier:
## Description

To fix issues as in spacetelescope/jdaviz#1909 where `f'{unit}'` returns unexpected results, this returns `self._units` of `None` as `''` for all `Component` subclasses. This is analogous to how these are already handled in `Coordinates`.
In case the internal value of `self._units == None` is still required somewhere, the second commit reverses the transformation on setting units, but it does not seem to be required for passing tests.

### DIFF
--- a/glue/core/component.py
+++ b/glue/core/component.py
@@ -54,7 +54,7 @@ class Component(object):
 
     @property
     def units(self):
-        return self._units
+        return self._units or ''
 
     @units.setter
     def units(self, value):
@@ -226,7 +226,7 @@ class CoordinateComponent(Component):
     @property
     def units(self):
         if self.world:
-            return self._data.coords.world_axis_units[self._data.ndim - 1 - self.axis]
+            return self._data.coords.world_axis_units[self._data.ndim - 1 - self.axis] or ''
         else:
             return ''
 
@@ -514,7 +514,7 @@ class DaskComponent(Component):
 
     @property
     def units(self):
-        return self._units
+        return self._units or ''
 
     @units.setter
     def units(self, value):


astrofrog:
I think I'm happy for the public API to not change, I just wanted to be be able to distinguish internally between units being set vs default.

pllim:
Is this related?

DeprecationWarning: Passing `np.nan` to mean no clipping in np.clip has always been unreliable, and is now deprecated. In future, this will always return nan, like it already does when min or max are arrays that contain nan. To skip a bound, pass either None or an np.inf of an appropriate sign.

dhomeier:
> I think I'm happy for the public API to not change, I just wanted to be be able to distinguish internally between units being set vs default.

Then perhaps it's better to cherry-pick the first commit, so passing `units=''` would not be translated to `None`.

dhomeier:
> Is this related?
> 
> DeprecationWarning: Passing `np.nan` to mean no clipping in np.clip has always been unreliable, and is now deprecated. I

To the units treatment? No, showed already up before the units PR was merged; I suspect it is related to the
> glue/glue/viewers/scatter/layer_artist.py:60: RuntimeWarning: All-NaN slice encountered
    return 10. ** (np.log10(np.nanmax(array)) * self.contrast)

from the same `test_density_map_incompatible_subset`, but can't tell where exactly this is raised.

astrofrog:
> Then perhaps it's better to cherry-pick the first commit, so passing `units=''` would not be translated to `None`.

@dhomeier - sounds good, I'll remove the second commit

astrofrog (approved):
The other jdaviz failures are unrelated to this PR, so merging this and will open a separate PR to fix the other issue.

PR #2360, opened 2023-02-06
Add pretransform for full-sphere selection

Carifio24:
I realized that I missed one thing in #2348 with regards to selection. Any points whose latitudes were wrapped back into [-π,π] won't be selected when making a ROI selection because the scatter viewer's `apply_roi` doesn't know about that transformation. This PR adds a pretransform that lets it know about what happens with longitudes.

### DIFF
--- a/glue/viewers/scatter/qt/tests/test_data_viewer.py
+++ b/glue/viewers/scatter/qt/tests/test_data_viewer.py
@@ -43,6 +43,8 @@ class TestScatterViewer(object):
                          y=[3.2, 3.3, 3.4, 3.5], z=['a', 'b', 'c', 'a'])
         self.data_2d = Data(label='d2', a=[[1, 2], [3, 4]], b=[[5, 6], [7, 8]],
                             x=[[3, 5], [5.4, 1]], y=[[1.2, 4], [7, 8]])
+        self.data_fullsphere = Data(label='d3', x=[6.9, -1.1, 1.2, -3.7],
+                                    y=[-0.2, 1.0, 0.5, -1.1])
 
         self.app = GlueApplication()
         self.session = self.app.session
@@ -51,6 +53,7 @@ class TestScatterViewer(object):
         self.data_collection = self.session.data_collection
         self.data_collection.append(self.data)
         self.data_collection.append(self.data_2d)
+        self.data_collection.append(self.data_fullsphere)
 
         self.viewer = self.app.new_data_viewer(ScatterViewer)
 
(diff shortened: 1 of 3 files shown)


glue/viewers/scatter/qt/tests/test_data_viewer.py
```diff
@@ -914,61 +914,90 @@ def test_limits_log_widget_fullsphere(self):
...
+        }
 
-        viewer_state.y_log = True
-        self.viewer.apply_roi(roi)
-        state = self.data.subsets[0].subset_state
-        assert state.pretransform
-        pretrans = state.pretransform
-        assert isinstance(pretrans, ProjectionMplTransform)
-        assert pretrans._state['y_scale'] == 'log'
+        for angle_unit in ['radians', 'degrees']:
```
dhomeier, line 927:
```suggestion
        for angle_unit in ['radians', 'degrees']:
```
I think testing structure would benefit from instead setting this and the other test up with
```python
@pytest.mark.parametrize('angle_unit', ['radians', 'degrees'])
```

dhomeier (changes requested):
Thanks for the fix and improving the tests!
I could not read out of the test setup if there is actually an example with, say RA > 2*pi input, tested. Otherwise only the small restructuring request.

glue/viewers/scatter/qt/tests/test_data_viewer.py
Carifio24: I agree, using `parametrize` would be better. Maybe should include the expected mask as a parameter as well?

dhomeier: As in `parametrize('angle_unit,expected_mask' (...)`? Had not even thought of that, but yes, that might make it even clearer.

Carifio24:
@dhomeier I've updated the tests to use `parametrize` like we discussed, and have the full-sphere test also check the expected masks. I also made sure that some of the data points require longitude wrapping as you had mentioned, at least in radians mode.

Some of the export tests are now failing. I suspect these failures are unrelated because they're using the `rectilinear` projection, but my local tests are passing and I haven't had time yet to investigate further.

Carifio24:
The same errors with the image export are popping up in the CI of other PRs (e.g. #2365), so I think it's safe to say that they're unrelated.

Carifio24:
Just rebased to pick up the changes from #2370 for CI purposes

dhomeier (approved):
Looks all good now; I trust that the tests have the correct expected outcomes, thanks!

PR #2365, opened 2023-02-19
MNT: drop runtime dependency on pkg_resources (setuptools)

### LINKED ISSUE #2364: BUG: importing glue with setuptools>=67.3 raises a DeprecationWarning
**Describe the bug**
`pkg_resources.get_distribution` was changed in `setuptools` 67.3, released last week, and now raises a deprecation warning on import.

`pkg_resources` as a whole is legacy and should be avoided, in particular because it creates a de facto runtime dependency on setuptools, which is currently undeclared.

**To Reproduce**
```shell
python -m venv venv
source venv/bin/activate
python -m pip install --upgrade setuptools
python -m pip install glue
python -Werror -c "import glue"
```

output
```
...
DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('mpl_toolkits')`.
Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
```

**Expected behavior**
should return without error

**Details:**

- Operating System: all
- Python version (``python --version``): all
- Glue version (``glue --version``): 1.7.0
- How you installed glue: pip

I should be able to patch `pkg_resources` out completely.

neutrinoceros:
## Description

resolves #2364

I left one call to `pkg_resources.get_distribution` in `doc/conf` because I don't know how to replace it without requiring `glue` be installed to build docs, and anyway it's not an issue for downstream code.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,7 +8,6 @@ classifiers =
     Operating System :: OS Independent
     Programming Language :: Python
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
@@ -44,6 +43,7 @@ install_requires =
     openpyxl>=3.0
     mpl-scatter-density>=0.7
     pvextractor>=0.2
+    importlib_resources>=1.3; python_version<'3.9'
 
 [options.entry_points]
 glue.plugins =
--- a/glue/__init__.py
+++ b/glue/__init__.py
@@ -6,12 +6,9 @@ import os
 
 import sys
 
-from pkg_resources import get_distribution, DistributionNotFound
+import importlib.metadata
 
-try:
-    __version__ = get_distribution('glue-core').version
-except DistributionNotFound:
-    __version__ = 'undefined'
+__version__ = importlib.metadata.version('glue-core')
 
 from ._mpl_backend import MatplotlibBackendSetter
 sys.meta_path.append(MatplotlibBackendSetter())
--- a/glue/_deps.py
+++ b/glue/_deps.py
@@ -213,8 +213,8 @@ def plugins():
     modules = []
     dependencies = []
     for entry_point in iter_plugin_entry_points():
-        module_name = entry_point.module_name.split('.')[0]
-        package = entry_point.dist.project_name
+        module_name, _, _ = entry_point.module.partition('.')
+        package = entry_point.dist.name
         modules.append((module_name, package))
     for module, package in sorted(set(modules)):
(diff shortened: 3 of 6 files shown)


setup.cfg
```diff
@@ -8,7 +8,6 @@ classifiers =
     Operating System :: OS Independent
     Programming Language :: Python
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.7
```
neutrinoceros, line 11: I saw `python_requires = >=3.8` a few lines down, which is why I'm cleaning this marker while I'm in there

neutrinoceros:
I think all failures are unrelated and were already seen in previously merged PRs, so I'll open this for review

neutrinoceros:
@astrofrog, anything else I can do here ?

astrofrog (approved):
Looks good to me, thanks!

dhomeier:
@neutrinoceros this PR has introduced a failure on `test_optional_dependency_not_imported` in all py38 and py39 envs, e.g.
https://github.com/glue-viz/glue/actions/runs/4217654549/jobs/7321596947#step:9:485

Unfortunately this was missed among all the very verbose python_exporter failures that are still under review.
I cannot reproduce this with Python 3.9.15 on macOS with the same optional packages installed, but I don't see the ImportError from the test raised on any of the packages in question, but instead an
> AttributeError: 'EntryPoint' object has no attribute 'dist'

Yet the test is passing locally with any of setuptools 66.0.0, 67.4.0 or 67.6.0. Any ideas?

neutrinoceros:
I am able to reproduce the error locally (on Python 3.9.15), however I don't yet understand what is happening. Here are my notes at this point

The `dist` attribute is still defined at the class level as of `importlib_metadata==6.1.0`, which is the version you currently get when installing `glue` in a fresh env with Python 3.9.
https://github.com/python/importlib_metadata/blob/700f2c7d74543e3695163d5487155b92e6f04d65/importlib_metadata/__init__.py#L198

Python 3.11 ships with `importlib_metadata==4.13`.

neutrinoceros:
I had it backwards ! So I _assumed_ that `importlib.metadata` was compatible with the code I wrote no matter what version I used, however this is not quite true. [The `EntryPoint.dist` attribute is only available in `importlib_metadata>=3.3`](https://github.com/python/importlib_metadata/pull/266), and [Python 3.8 ships `importlib_metadata==1.3`](https://pypi.org/project/importlib-metadata/) (I have no information on what version Python 3.9 ships but it seems to be `<3.3`). Now that I got it, I can patch this.

PR #2366, opened 2023-02-21
Update polar log transform test values to correct matplotlib 3.7 representation

Carifio24:
One of the CI tests that has been failing lately is the test of `ProjectionMplTransform` in the case of a polar projection with a radial log axis. I believe the reason for the error is the fix in matplotlib PR 24825, which was added in 3.7.0. This PR aims to fix this test.

Plotting out the test points by hand, the new values in this PR are exactly what one would expect to see - at least to me, the old values don't make sense (hence the need for an upstream fix). Since this is a bug in old versions of matplotlib, I decided to mark the test as an xfail on an older MPL version. Along the way, I created a `make_xfailer` similar to `make_skipper`, and did a bit of refactoring since they have the same basic idea.


dhomeier:
Thanks for the updates; I did not have a good idea how the polar log example was supposed to look like, but from the discussion around matplotlib/matplotlib#24825 get that it really was broken before 3.7.

> Since this is a bug in old versions of matplotlib, I decided to mark the test as an xfail on an older MPL version. Along the way, I created a `make_xfailer` similar to `make_skipper`, and did a bit of refactoring since they have the same basic idea.

Wondering if this could be further generalised to also trigger on `_ge_NN` conditions (e.g. with a kwarg to `make_marker` etc.) – asking because for the moment I'd prefer to mark the remaining 4 tests xfail until #2368 is resolved, so the `publish` CI can at least run.

Carifio24:
Do you mean a kwarg that reverses the direction of the version check? If so then I can add that in.

dhomeier:
Yes, so one could also define a ` xfail_matplotlib_ge_37` etc.
Maybe patching `MplPolygonalROI` will do for a quick fix of the legend tests now, but this could still be useful.

Carifio24:
@dhomeier I've just pushed a commit to allow `make_marker` to be more flexible - one can choose which operation (lt, gt, le, ge) to check for.

So for example, to make `xfail_matplotlib_ge_37`, one would do

`MATPLOTLIB_LT_37, xfail_matplotlib_ge_37 = make_xfailer('matplotlib', version='3.7', xfail_if='ge')`

dhomeier (changes requested):
Awesome, thanks! Just a few style suggestions, should be ready to go then when CI passes (except for the 4 known ones).

Carifio24:
Requested changes have been made

dhomeier (approved):

dhomeier:
Great, thanks again!

PR #2369, opened 2023-02-28
DOC: Fix equivalent_units in examples

pllim:
I think the example is wrong?

https://github.com/glue-viz/glue/blob/11a5dded32a5d87409377a45729a671e83bc218a/glue/core/units.py#L52-L53


dhomeier (approved):
Looks correct, thanks!

astrofrog (approved):
Thanks for the fix!

Issue #2368, opened 2023-02-24
Drawn ROIs affect legend placement in matplotlib 3.7


Carifio24:
**Describe the bug**
With matplotlib 3.7 (in particular the updates from matplotlib/matplotlib#9598), ROIs drawn with selection tools now affect the placement of the legend on the graph when using the "draggable" (default) or "best" legend location options. The reason for this is that the path of non-rectangular patches is now used when determining the best option. This means that the `Patch` objects drawn during ROI selection affect the legend placement (even though they aren't visible). Note that some of the patches have an initial size (e.g. the patch for `MplPolygonalROI` has initial vertices [0, 0] and [1, 1])  which can cause this issue to manifest before any selections are made, depending on the viewer bounds.

**To Reproduce**
Steps to reproduce the behavior such as:
1. Open up glue
2. Add some data to a scatter plot and enable the legend
3. Wherever the legend shows up, make an ROI selection such that the ROI patch overlaps the legend (which points are selected doesn't matter).
4. The legend will likely move

**Expected behavior**
Selection ROIs shouldn't affect legend placement.

**Screenshots**
Here's an example screenshot. Matplotlib prefers the center-right placement of the legend, which overlaps a visible data point, over the open top-right area, because I had made a selection with a rectangular ROI there.
![legend](https://user-images.githubusercontent.com/14281631/221237540-1e0a4afe-f532-47e7-ade5-546bce2b5308.png)

**Details:**

- Operating System: Ubuntu 20.04
- Python version (``python --version``): 3.10.9
- Glue version (``glue --version``): 1.7.0
- How you installed glue: pip

**Additional context**
Note that this is the cause of four of the five current CI failures (see #2366 for a resolution to the other one). The discrepancy in the tests occurs because the patches aren't included in the export script.

Carifio24:
One solution to this problem could be to only add the patch to the axes during the ROI's `start_selection` and remove it during `finalize_selection` rather than changing the visibility. I'm not sure how much of an effect, if at all, this would have on performance.

dhomeier:
> Selection ROIs shouldn't affect legend placement.

Not quite sure if I actually agree with this. If an ROI exists when plotting, one may reasonably expect that anything inside it is of some interest (that's in the name at least ;-), so it makes sense to avoid covering that area, if possible.
[...]

PR #2370, opened 2023-03-01
WIP: avoid interference of ROI patches with legend placement

dhomeier:
## Description
Reported in #2368; as of matplotlib 3.7 legend auto-placement is affected by the position of ROIs, even if empty. Notably 4 tests in viewers/scatter failed as the default placement of a `MplPolygonalROI` pushed the legends out of their usual position.
As a workaround the default extent and position of the ROI is moved from `[0, 1], [0, 1]` to `[0.2, 0.3], [0.2, 0.3]` to keep out of the way of any of the 9 standard positions for the legend. Better fix would be to hold of syncing the patch until the ROI is made visible. Alternatively xfail for matplotlib >= 3.7.


glue/core/roi.py
```diff
@@ -1229,7 +1228,10 @@ def _sync_patch(self):
             self._patch.set_height(height)
             self._patch.set(**self.plot_opts)
             self._patch.set_visible(True)
+            self._axes.add_patch(self._patch)
         else:
+            if self._patch in self._axes.patches:
+                self._axes._remove_method(self._patch)
```
Carifio24, line 1234: I'm getting an error here the second time I try to use any particular ROI tool. I think we want to call `self._patch.remove()` (or I guess maybe `self._patch.remove_method()`, rather than the axes' remove method?

Carifio24 (changes requested):

glue/core/roi.py
dhomeier: Thanks, I have chosen the second option (first one seemed to let the tests pass as well, but I checked the KeyError on selecting an ROI a second time in the app was resolved by this one.
Still be good if you can confirm in your installation; I am not sure about the performance when using the app.

Carifio24: This change has fixed the issue on my end

Carifio24 (approved):

dhomeier:
Thanks for your assistance @Carifio24!

PR #2372, opened 2023-03-06
Only recreate table model when Data layer changes

Carifio24:
This PR implements a change(/bugfix?) requested by @aagoodman. Currently, after a user makes a selection to create a new subset after having sorted the table, the sorting is undone once the subset is created. This leads to a not-great UX - if one is trying to create multiple selections based on a column's values, having to re-sort each time is annoying.

The reason for this is that the table model is recreated any time the layers change, which resets any sorting. This PR modifies the layer changed callback to only recreate the table model if the actual `Data` layer has changed - otherwise, we just call the model's `_update_visible`. This also necessitates making the masks in `_update_visible` aware of the current order.

### DIFF
--- a/glue/viewers/table/qt/data_viewer.py
+++ b/glue/viewers/table/qt/data_viewer.py
@@ -166,7 +166,7 @@ class DataTableModel(QtCore.QAbstractTableModel):
         visible = np.zeros(self.order.shape, dtype=bool)
         for layer_artist in self._table_viewer.layers:
             if layer_artist.visible:
-                mask = layer_artist.layer.to_mask()
+                mask = layer_artist.layer.to_mask()[self.order]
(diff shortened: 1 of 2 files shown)


Carifio24:
Rebased to get CI fixes

astrofrog:
@Carifio24 - do you think it would be easy to add a test for this, or would it be non-trivial?

Carifio24:
@astrofrog I think it should actually be pretty straightforward. I'll try adding one and see if that is indeed the case.

Carifio24:
@astrofrog I just added a test - it creates a new subset and checks that the table model and the row ordering are preserved afterwards.

astrofrog (approved):
Looks good to me provided the CI passes!

Issue #2351, opened 2023-01-20
Add an ExtendedComponent to glue


jfoster17:
**Is your feature request related to a problem? Please describe it:**
Some datasets such as geographic entities, features on the genome, and regions-of-interest in an image (identified algorithmically) inherently contain information about their extent in a way that glue should be able to visualize and reason about. For anything except a tiny dataset it is not feasible or desirable to translate all these extents into glue subsets.

**Describe the solution you'd like:**
Add an `ExtendedComponent` type to glue that would allow for a component to specify a shape (or set of shapes) for each element of a dataset. `ExtendedComponents` would describe an extant on one or more other `Components` in the dataset (`parent_components`— which may be `CoordinateComponents` or regular `Components`), and should contain references to those parent_components. Data Loaders would be responsible for parsing input data files into the appropriate components and individual Viewers would be updated to support plotting `ExtendedComponents`. `ExtendedComponents` should perhaps be hidden from the user and only used by Viewers when plotting the `parent_components` and transforming ROIs to `SubsetStates`.

**Describe alternatives you've considered:**
The following packages provide two partial and alternative approaches to this problem:

- Geographic regions (e.g. https://github.com/jfoster17/glue-map) -- uses a custom `Data` class which includes geometry and centroid attributes to represent the extent of geographic regions.
- Regions-of-interest in astronomy images (e.g. https://github.com/glue-viz/glue-regions) -- casts regions as subsets

**Additional context:**

The biggest decisions to be made are:

- How to handle ROI selections that partially intersect with an `ExtendedComponent`
- How to deal with the conversion that a user will probably want to make from a subset over data points (i.e. states) to a subset over the extent of that data points (i.e. the geographic region contained within a set of states)

![CaseA](https://user-images.githubusercontent.com/3639698/213817207-d671a14e-3dd3-4866-b8a1-36b09db06a4f.png)

![CaseB](https://user-images.githubusercontent.com/3639698/213817250-2e16ec22-8d36-45b4-84b0-f0799ad52eba.png)

See the attached overly long PDF for my full thinking through of what this might look like.

[ExtendedComponent.pdf](https://github.com/glue-viz/glue/files/10470684/ExtendedComponent.pdf)

**Related Issues**

- https://github.com/glue-viz/glue/issues/320
- https://github.com/glue-viz/glue/issues/2086
- https://github.com/glue-viz/glue-astronomy/issues/55
- https://github.com/glue-viz/glue/issues/1686

[...]

Issue #2330, opened 2022-10-11
Pandas DataFrames with type == 'object' cannot be save/restored 


jfoster17:
**Describe the bug**
Pandas DataFrames created within glue and added to the data_collection manager may have columns of type 'object', which mean they cannot be save/restored by glue (`glue.core.state._load_numpy` calls `np.load() `without `allow_pickle=True`). This is generally not a problem when reading files using the Pandas data_factory (which converts columns), but does, for instance cause problems for datasets retrieved from external sources within a glue session. 

**To Reproduce**
Steps to reproduce the behavior such as:
1. Create a Pandas DataFrame within glue and add it to the data_collection. For instance, one might use the [process described in the documentation](http://docs.glueviz.org/en/stable/python_guide/data_translation.html?highlight=pandas#working-with-non-glue-data-objects)
```from pandas import DataFrame
df1 = DataFrame()
df1['a'] = [1.2, 3.4, 2.9]
df1['g'] = ['r', 'q', 's']
dc['dataframe'] = df1
````
3. Save Session (this new Data object will be stored as a numpy array within the session file since it did not come from an external file)
4. Restore Session
5. Get  the following error:
> value error: 'Object arrays cannot be loaded when allow_pickle=False'

**Expected behavior**
Pandas objects created within glue should not break session files. 

We could simply add `allow_pickle` to `np.load()`, but perhaps this has undesired side effects?

**Details:**

- Operating System: MacOS 12.6
- Python version Python 3.9
- Glue version 1.6
- How you installed glue: conda

**Additional context**
Sample session file attached:
[pandas_dataframe_session.glu.gz](https://github.com/glue-viz/glue/files/9759648/pandas_dataframe_session.glu.gz)

PR #2373, opened 2023-03-06
WIP: Add extended component

jfoster17:
# Pull Request Template

## Description

Here is one approach to displaying many regions (e.g. regions-as-data) in glue (ala #2351). I've implemented this for the first place I wanted it, which is as a different kind of ScatterLayer in the ImageViewer. 

The pieces are basically:

- An `ExtendedComponent(Component)` which is a list of Shapely geometries
- A `RegionData(Data)` which is probably unnecessary (more `Data` types are hard to support) but for now is used to determine if a `Data` object should be plotted as regions or as points.
- `ScatterRegionLayerState` and `ScatterRegionLayerArtist` classes to deal with actually configuring and plotting regions
- Some code from GeoPandas (with LICENSE) for the actual plotting/updating of arbitrary polygons. 

Major unfinished things
- [ ] Projections/transformations don't work on regions (would need to apply transformation to all points in the region definition)
- [ ] Regions can be displayed and selected, but cannot be used for logical operations
- [ ] Regions are selected into subsets based on their representative point (a cheap centroid that must be inside the geometry). This might not always be the desired behavior.
- [ ] It's much less clear what the best behavior is when adding RegionData to a scatterplot. The assumption is that if you have regions, you want to plot regions on an image (perhaps color-coded by another component). If you have a table of data about regions you *might* want to make a scatterplot that just shows those regions (i.e. you don't have an image and just want to see the geometry) or you *might* want to make a scatterplot 2 components of those regions.

### DIFF
--- a/glue/core/component.py
+++ b/glue/core/component.py
@@ -475,6 +475,86 @@ class CategoricalComponent(Component):
         return pd.Series(self.labels, dtype=object, **kwargs)
 
 
+class ExtendedComponent(Component):
+    """
+    Container for data that describes an extent
+
+    This component can be used when a dataset describes regions
+    and it does not make sense to convert all those regions
+    to glue subsets.
+
+    Data loaders are required to know how to
+    provide these extended regions as lists of shapely objects and
+    explicitly create these components. If a tabular dataset provides
+    e.g. a range over another component or a circular region then
+    an ExtendedComponent needs to be manually created from these definitions.
+
+    Data Viewers can choose to implement custom Layer Artists for extended
+    components, typically in the form of looking if there is an ExtendedComponent
+    with parent components that are being plotted.
+
+    The challenge with this approach is that we tend to add a LayerArtist when
+    we create a layer... in the mapping context we have distinct LayerArtists
+    and swapping them out when we change state.lon_att and state.lat_att might be annoying?
+
+    NOTE that this implementation precludes support for regions in more than
+    two dimensions
+
+
+    Geopandas does this in a somewhat optimized way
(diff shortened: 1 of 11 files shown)


jfoster17:
Here are what the basic interactions look like:
https://user-images.githubusercontent.com/3639698/223265425-c45a4d0e-74bf-4fba-a94e-a2fc914c3f87.mov

astrofrog:
@jfoster17 this is very nice! It looks like this will need rebasing due to conflicts. Would you like me to review this yet at this point or are there things you still need to work on before a first review?

jfoster17:
There are still a few more things I need to fix/add before this is ready for review. I will try to keep it to the smallest amount required to be useful though, so that it doesn't get too large.

jfoster17:
I wanted to highlight 79e4fa57ec44c50977c4056d9ba3fbc2c9a3322b, because it opens (or increases?) the potential for malicious session files and deserves some consideration. If I'm going to represent regions as Shapely objects we need something like this to enable save/restore of sessions that bundle the data. This also solves #2330.

astrofrog:
I'd rather not use allow_pickle if at all possible - is there any way we can serialize/deserialize the Shapely objects to strings?

astrofrog:
As noted in https://github.com/glue-viz/glue/issues/2430, I will be splitting out Qt-related code into a separate repository - however that repository will have the same history as the present one here so it should be easy to split the Qt-specific parts of this PR into a separate PR to glue-qt. Sorry for the disruption!

jfoster17:
Closing in favor of #2442

PR #2375, opened 2023-03-09
Fix broken link editor under Qt5

### LINKED ISSUE #2374: Selecting nodes in the link editor is broken with PyQt5
**Describe the bug**
The link editor find_object function is broken with 3+ datasets and at least some versions of PyQt5 and the latest dev version of glue. The problem is that 71ca079 changed to use event.position() [which seems correct for Qt6](https://stackoverflow.com/questions/67496362/qmouseevent-object-has-no-attribute-pos), but is not (generally?/always?) available for earlier versions.

**To Reproduce**
Add 3+ datasets and open the link editor with Qt5 installed. Attempting to click on a node in the link editor fails with 

> Traceback (most recent call last):
>   File "/Users/jfoster/python-repos/glue/glue/dialogs/link_editor/qt/data_graph.py", line 374, in mouseMoveEvent
>     selected = self.find_object(event)
>   File "/Users/jfoster/python-repos/glue/glue/dialogs/link_editor/qt/data_graph.py", line 366, in find_object
>     if obj.contains(event.position()):
> AttributeError: 'QMouseEvent' object has no attribute 'position'
> 

This error actually happens with 2 datasets too, but since the two datasets are already selected by default it does not block user action.

**Details:**

Glue: 1.7.1.dev21+geeb78251.d20230307
Python: 3.10.9
PyQt5: PyQt: 5.15.7 - Qt: 5.15.6
PyQt6: Not installed
PySide2: Not installed
PySide6: Not installed
qtpy: 2.3.0
setuptools: 67.5.1
echo: 0.8.0
numpy: 1.24.2
matplotlib: 3.7.1
pandas: 1.5.3
astropy: 5.2.1
dill: 0.3.6
h5py: 3.8.0
xlrd: 2.0.1
openpyxl: 3.1.1
mpl-scatter-density: 0.7
casa-formats-io: 0.2.1
glue-core: 1.7.1.dev21+geeb78251.d20230307
glue-vispy-viewers: 1.0.7
IPython: 8.11.0
qtconsole: 5.4.0
ipykernel: 6.21.2
traitlets: 5.9.0
pygments: 2.14.0
pyzmq: 25.0.0
scipy: 1.10.1
scikit-image: 0.19.3
PyAVM: Not installed
spectral-cube: 0.6.0
astrodendro: Not installed
mock: Not installed
pytest: 7.2.2
plotly: Not installed

**Additional context**
We don't test this interaction (using a mock instead), so I'm not sure the best way to add a regression test.

jfoster17:
# Fixes a bug where the link editor did not allow you to select nodes under Qt5

## Description

Close #2374 by simply using try/except to do both the Qt6 and Qt5 methods.

### DIFF
--- a/glue/dialogs/link_editor/qt/data_graph.py
+++ b/glue/dialogs/link_editor/qt/data_graph.py
@@ -363,8 +363,12 @@ class DataGraphWidget(QGraphicsView):
 
     def find_object(self, event):
         for obj in list(self.nodes) + self.edges:
-            if obj.contains(event.position()):
-                return obj
+            try:  # event.position() is Qt6 
+                if obj.contains(event.position()):
+                    return obj
+            except AttributeError:  # event.pos() is Qt5
+                if obj.contains(event.pos()):
+                    return obj
 
     def mouseMoveEvent(self, event):
 
--- a/glue/dialogs/link_editor/qt/data_graph.py
+++ b/glue/dialogs/link_editor/qt/data_graph.py
@@ -363,7 +363,7 @@ class DataGraphWidget(QGraphicsView):
 
     def find_object(self, event):
         for obj in list(self.nodes) + self.edges:
-            try:  # event.position() is Qt6 
+            try:  # event.position() is Qt6
                 if obj.contains(event.position()):
                     return obj
             except AttributeError:  # event.pos() is Qt5


jfoster17:
It looks like all these test failures are just related to #2368. Can I get this merged?

And, ideally, a bugfix release before bio-glue-con (3/30 - 3/31) so that new glue users at that event won't be tripped up by this bug?

PR #2377, opened 2023-03-21
Fixed a bug in the plugin iteration on Python 3.8 and 3.9

neutrinoceros:
fix a regression introduced and discussed in #2365
ping @dhomeier

### DIFF
--- a/glue/_plugin_helpers.py
+++ b/glue/_plugin_helpers.py
@@ -9,16 +9,15 @@
 import os
 import sys
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


dhomeier (approved):
Ah, I did not get the difference between `importlib.metadata` and `importlib_metadata`. 
Actually `'importlib-metadata>=3.6; python_version < "3.10"'` had already been installed, apparently pulled in via something else, but better to make this explicit.
Thanks for the quick fix!

PR #2380, opened 2023-03-27
Save legend state

Carifio24:
While we were dealing with the issues with legends and matplotlib 3.7, I noticed that we don't save the legend state of the matplotlib viewers. This PR adds functionality to do that, which gives a small but nice UX improvement. This shouldn't have any effect on loading existing sessions, since then the relevant record won't have a `"legend"` key and the new code in `__setgluestate__` won't run.


astrofrog (approved):
Thanks!

PR #2385, opened 2023-04-11
Only force reference data change if message indicates that numerical values of data have changed

astrofrog:
The original intent of the ``_reference_data_changed(force=True)`` call was for cases where the coordinate system is changed, but we don't need to always force an update in other situations that might cause ``_update_data`` to be called.


pllim:
@kecnry , should probably test this against the unit conversion branch over at Jdaviz since I don't know if changing viewer unit should trigger this or not.

Sounds like we can ignore https://github.com/glue-viz/glue-jupyter/pull/350 ?

PR #2387, opened 2023-04-13
Add functionality for moving viewers between tabs

Carifio24:
This PR adds the ability to move a viewer between tabs. This is achieved via the new `GlueApplication` method `move_viewer_to_tab`, and exposed in the UI via a subtool (as discussed in the glue meeting on 4/13/2023). As discussed there, this is implemented as a subtool with the idea that we could put tools to do other "window-y" operations (like changing the viewer title) in this menu as well. (I should also note that we can't easily add something like this to the title bar's context menu, as the title bar comes from the system window manager.)

In the UI, this looks as below

https://user-images.githubusercontent.com/14281631/232147567-252a1f8e-f524-46a9-8f2a-e4c93c21534f.mp4

move_tabs

As for the actual implementation, it's relatively straightforward. As explained in the comment in `move_viewer_to_tab`, I chose to destroy the current `GlueMdiSubWindow`  and then call `add_widget` for the new tab, rather than simply re-parent the existing one, because this allows us to get the nice behavior (that we also get in e.g. `new_data_viewer`) where the window will be placed in a nice location in the new tab. If we just re-parent then the window is given the same position in the new tab it had in the old one, which can lead to one viewer covering another (for example, if the user has two viewers that they haven't moved/resized in different tabs).

If anyone has any thoughts (particularly on the UI/icons), please let me know!


glue/app/qt/tests/test_application.py
```diff
@@ -235,6 +235,31 @@ def test_subset_facet(self):
...
+        self.app.move_viewer_to_tab(viewer, 0)
+        assert viewer.parent().mdiArea() is self.app.tab(0)
+        assert set(self.app.tab(0).subWindowList()) == {self.app._terminal, viewer.parent()}
+        assert len(self.app.tab(1).subWindowList()) == 0
+
+        # Check that we do nothing if the given tab is the same as the
+        # viewer's current tab
+        parent = viewer.parent()
+        self.app.move_viewer_to_tab(viewer, 0)
+        assert parent is viewer.parent()
```
astrofrog, line 261: Could you test that viewer state is preserved when moving it between tabs?

astrofrog (changes requested):
Looks good! Just one small request regarding checking that state is preserved

glue/app/qt/tests/test_application.py
Carifio24: I've added in that check

astrofrog (changes requested):
Thanks! This works nicely, except for a small issue I've noticed - if I keep moving an image viewer between two tabs it keeps shrinking vertically each time it is moved (it might happen with other viewers too but haven't checked yet).

Another things that would be nice is if the default selected tab in the dropdown in the dialog is not set to the current tab, if there is more than one tab, but don't worry if you don't have time to implement this.

Carifio24:
@astrofrog Good catch on the resizing! I hadn't noticed that, but yeah it's very apparent if you move the same viewer back and forth a few times. A viewer actually will shrink horizontally as well, though it's much less.

It looks like what's happening is that when we call a Qt viewer's `mdi_wrap` method, we resize the MDI window to match the viewer size, which then causes the viewer to shrink a small amount to fit inside. Then when we move, the new MDI wrapper is again sized based on the viewer, which is now slightly smaller...rinse and repeat. To fix this, I just resized the viewer to the MDI window's current size (after it's removed from the MDI area, so this should never be visible to the user).

As for the default selected tab, I just made it 1 if the current tab is 0 and there are multiple tabs, and 0 otherwise. This way the default is never the current tab if that can be avoided, and since tab order doesn't have any particular meaning I don't think we can be any more clever.

astrofrog (approved):
Looks great, thanks!

Issue #2208, opened 2021-07-15
ROI implementation of setting the center position is inconsistent


pllim:
`RectangularROI.set_center()` does not exist even though the base class implies it should.

pllim:
`EllipticalROI` does not have it either but `CircularROI` does. 🤷

pllim:
I guess `move_to` is supposed to serve the same purpose?

PR #2391, opened 2023-04-26
Make ROI.move_to() behave consistently and expose it at subset_state level

pllim:
Make `ROI.move_to()` behave consistently such that all of them take absolute coords now, not some absolute and some delta. Also expose `move_to` machinery at `subset_state` level so we can move ROIs in composite subset all at once.

The main motivation is so I can have user click to recenter a subset that can be a simple shape with single ROI or an annulus that is actually a subset state with 2 attached ROIs (https://github.com/glue-viz/glue-astronomy/pull/90).

I don't know why some `move_to` expects only delta but that kind of inconsistent behavior is more of a bug than feature from my perspective.

### DIFF
--- a/glue/core/roi.py
+++ b/glue/core/roi.py
@@ -1706,7 +1706,7 @@ class MplPolygonalROI(AbstractMplRoi):
             xval, yval = axes_trans.transform([event.x, event.y])
 
         if self._scrubbing:
-            old_x, old_y = self._roi.centroid()
+            old_x, old_y = self._roi.center()
             new_x = old_x + xval - self._cx
             new_y = old_y + yval - self._cy
             self._roi.move_to(new_x, new_y)
--- a/glue/core/roi.py
+++ b/glue/core/roi.py
@@ -1,4 +1,5 @@
 import copy
+import warnings
 
 import numpy as np
 from matplotlib.patches import Ellipse, Polygon, Rectangle, Path as MplPath, PathPatch
@@ -487,17 +488,20 @@ class CircularROI(Roi):
             y = np.asarray(y)
         return (x - self.xc) ** 2 + (y - self.yc) ** 2 < self.radius ** 2
 
-    def set_center(self, x, y):
+    def set_center(self, x, y):  # pragma: no cover
         """Set the center of the circular region"""
-        self.xc = x
-        self.yc = y
+        warnings.warn("set_center is deprecated and may be removed "
+                      "in a future release, use move_to", DeprecationWarning)
+        self.move_to(x, y)
 
     def set_radius(self, radius):
         """Set the radius of the circular region"""
         self.radius = radius
 
-    def get_center(self):
-        return self.xc, self.yc
+    def get_center(self):  # pragma: no cover
+        warnings.warn("get_center is deprecated and may be removed "
+                      "in a future release, use center", DeprecationWarning)
+        return self.center()
 
     def get_radius(self):
         return self.radius
@@ -509,8 +513,8 @@ class CircularROI(Roi):
         self.radius = 0.
 
     def defined(self):
-        return self.xc is not None and \
-            self.yc is not None and self.radius is not None
+        return (self.xc is not None and
+                self.yc is not None and self.radius is not None)
 
     def to_polygon(self):
         if not self.defined():
@@ -523,9 +527,12 @@ class CircularROI(Roi):
     def transformed(self, xfunc=None, yfunc=None):
         return PolygonalROI(*self.to_polygon()).transformed(xfunc=xfunc, yfunc=yfunc)
 
-    def move_to(self, xdelta, ydelta):
-        self.xc += xdelta
-        self.yc += ydelta
+    def center(self):
+        return self.xc, self.yc
+
+    def move_to(self, x, y):
+        self.xc = x
+        self.yc = y
 
     def __gluestate__(self, context):
         return dict(xc=context.do(self.xc),
@@ -556,8 +563,8 @@ class EllipticalROI(Roi):
 
     Notes
     -----
-        The `radius_x`, `radius_y` properties refer to the semiaxes along the `x` and `y`
-        axes *before* any rotation is applied.
+    The `radius_x`, `radius_y` properties refer to the semiaxes along the `x` and `y`
+    axes *before* any rotation is applied.
     """
 
     def __init__(self, xc=None, yc=None, radius_x=None, radius_y=None, theta=None):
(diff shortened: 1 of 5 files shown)


pllim:
I cannot tell if `Fatal Python error: Segmentation fault` is related or not.

Since this fundamentally changed `move_to` for a couple of ROI classes, maybe you want to tell me what else I have to fix to make this work.

pllim:
This error does look related. Any advice on how to fix the test? I am not familiar with that this scrubbing is supposed to do. Thanks!

```
____________________________ TestPolyMpl.test_scrub ____________________________

self = <glue.core.tests.test_roi.TestPolyMpl object at 0x7f091b8d1c90>
    def test_scrub(self):
        self.send_events()
        roi = self.scrub(roi=self.roi)
>       assert roi._roi.vx[0] == 6
E       assert -1.5 == 6
glue/core/tests/test_roi.py:1181: AssertionError
```

dhomeier:
> I don't know why some `move_to` expects only delta but that kind of inconsistent behavior is more of a bug than feature from my perspective.

I agree that's been bugging me before (may have been discussed in #2235, but I think there it was more about `rotate_to`/`rotate_by`). Behind this may be that at least `CircularROI` also has a `set_center` method, but that's been very inconsistently implemented (cf. #2208!). So I'd rather change `set_center` to `self.move_to(x, y)` now (and perhaps deprecate it at some point), and should it really be required, add a distinct `move_by` method.
The changed test results probably indicate that the `Mpl*ROI`s calling `move_to` within the scrubbing mechanism adapted to the old versions. I think this will need a bit more careful checking downstream to make sure this does not break anything.

For the `test_scrub` failure I think the correct fix would be to change `MplPolygonalROI.update_selection` to
```python
        if self._scrubbing:
            self._roi.move_to(xval, yval)
```

pllim:
Thanks for the pointers! I updated the code so that test result will not change.

glue/core/subset.py
```diff
@@ -1079,6 +1086,12 @@ def __init__(self, state1, state2=None):
     def copy(self):
         return type(self)(self.state1, self.state2)
 
+    def move_to(self, *args):
+        """Move any underlying ROI to the new given center."""
+        self.state1.move_to(*args)
+        if self.state2:
+            self.state2.move_to(*args)
```
dhomeier, line 1093: Thinking of your previous issue and comment on offset (non-concentric) AnnulusROIs, I am wondering if such `CompositeSubsetState`s could be constructed here. In that case the inner ROI would probably have to be moved as
```suggestion
           self.state2.move_to(self.state2.center()[0] - self.state1.center()[0] + x, self.state2.center()[1] - self.state1.center()[1] + y, *args)
```
...

pllim: I don't understand your suggestion. I had tested this downstream and the annulus moved fine with my patch as-is. I am relying on the built-in state recursion.

pllim: Oh... I just saw "non-concentric". Is that a thing? I have not seen such things implemented in both `regions` and `photutils`.

dhomeier:
Thanks, I did not trace the Polygon movement exactly, but [6, 7] looks rather like the correct position.

glue/core/subset.py
dhomeier: Yes, not saying that this is presently implemented anywhere, but if it could be done in principle, better to allow for such a case.

dhomeier:
I have not found any direct uses of `move_to()` or `set_center()` downstream in any of the glueviz packages, so with MplROI taken care of, this might be safe.
I did notice that `CircularROI` of all things does not implement `self.center`, so would add that as a wrapper/replacement for `get_center`, and then also have `set_center` just wrap `move_to`. @astrofrog: do you think this would be a good moment for getting rid of `set_center` internally (again, it is only used in `MplCircularROI` and the tests, and nowhere downstream in glueviz afaics, but might be used in userland code, so I'd probably not remove it completely yet)?

glue/core/subset.py
pllim: But the subset state currently does not have a `.center()` and I am not 100% all ROIs have consistent `.center()` methods. This is not something I want to deal with in this patch. I'd rather we open up a follow up issue for non-concentric support.

dhomeier: Yes, `CircularROI` is lacking it, not sure of any others; but if it is not available for `self.state2` generally, this can not be added here (I guess that means `CompositeSubsetState`s with different centres cannot be constructed ATM).

pllim: I mean, you could do composite subset state with whatever you want, and you correctly pointed out that `move_to` here won't work correctly for most of those except when they share the same center. I was hoping this won't be a problem for Jdaviz but now that I think about it... sigh, someone is going to draw something weird over there and encounter exactly this. 😭 

So, why don't you and Tom hash out what is the plan with https://github.com/glue-viz/glue/pull/2391#issuecomment-1524084015 and then we can try to address that here? Thanks!

astrofrog:
Yes I think that would be fine - I doubt many people use this API so it is ok to update.

pllim:
@dhomeier and @astrofrog , I think I addressed your comments. I tested the `move_to` with some weird composite subset and seems to work. It is reflected in the new tests I added.

glue/core/roi.py
```diff
@@ -865,7 +877,13 @@ def centroid(self):
 
         return np.dot(xs, dxy) * scl + x0, np.dot(ys, dxy) * scl + y0
 
-    def move_to(self, xdelta, ydelta):
+    def center(self):
+        return self.centroid()  # centroid is more robust than mean
```
dhomeier, line 881:
```suggestion
        # centroid is more robust than mean, but undefined for degenerate (1D) polygons
        if self.area() == 0:
            return self.mean()
        else:
            return self.centroid()
```
I think that special case needs to be taken care of (although it seems not to be tested :( ), just as done in https://github.com/glue-viz/glue/blob/d2adb6895514044d4b40fdf1f2c082ca92df6a4b/glue/core/roi.py#L887-L889 – `rotate_to` can then be simplified to just use
```python
        center = self.center() if center is None else center
```

pllim: Good call. Done. Thanks!

dhomeier (approved):
Huge thank you for taking care of all the refactoring and test updates!

pllim:
Thanks for the reviews and merge!

astrofrog:
I'm going to remove the Bug label because it's not really a bug, more of a clean-up, so I think I will bump the minor version as there is effectively an API change.

dhomeier:
Agreed; should go under enhancement for the changelog then.

PR #2392, opened 2023-04-26
Add filter/search to table viewer

### LINKED ISSUE #2386: Search/Filter in Table Viewer
**Is your feature request related to a problem? Please describe it:**
In large tables with text columns, it would be nice to be able to filter the Table Viewer to find rows where text matches a pattern. My specific use case is for tables of information about genes, where the user wants to filter on gene names and/or gene name patterns (e.g. all mitochondrial genes are named MT-*), so that the user can create subsets from matching gene names.

**Describe the solution you'd like:**
The Table Viewer would need a search field and a drop-down to select the column within which to search. Another option would be to allow the user to click on a column name and enter a filter string there, but I think this is less discoverable.

**Describe alternatives you've considered:**
It is possible to do this search/filter from the terminal, but being able to search/filter a table is a very natural GUI interaction.

I *think* this might not be too hard. If we follow this approach (https://stackoverflow.com/a/6791177) I just need to figure out how to make the QSortFilterProxyModel combined with our DataTableModel. If anyone else has thoughts on what this should look like or suggestions on how to implement this, that would be very welcome!

jfoster17:
# Add filter/search to the Table Viewer

table_filter

## Description

This PR adds a search/filter field and component selection widget to the Table Viewer. The displayed elements in the table are then filtered based on applying `re.search()` to the selected component. The primary motivation is to allow the user to use the GUI to find matching rows in a dataset and create a subset from those rows. I limit the user to searching on categorical components.

This necessitated adding a TableViewerState in order to keep track of the things going into the filter. 

I experimented with using `QSortFilterProxyModel` just for the filtering and although I was able to get a working implementation, I was then getting frequent segfaults that I was unable to diagnose. Since we are already doing sort within our custom `DataTableModel` it was easy enough to apply the filtering here as well. (I also that using `QSortFilterProxyModel` was rejected in b7a0f67500 for performance concerns, but performance seemed okay for the filtering component).

Still to do:

- [x] Save/restore tests are failing
- [x] Add tests specific to the filtering

Closes #2386

### DIFF
--- a/glue/viewers/table/compat.py
+++ b/glue/viewers/table/compat.py
@@ -22,7 +22,7 @@ def update_table_viewer_state(rec, context):
         rec['state'] = {}
         rec['state']['values'] = {}
 
-        properties = rec.pop('properties')
+        _ = rec.pop('properties')
 
         layer_states = []
 
--- a/glue/viewers/table/compat.py
+++ b/glue/viewers/table/compat.py
@@ -22,7 +22,10 @@ def update_table_viewer_state(rec, context):
         rec['state'] = {}
         rec['state']['values'] = {}
 
-        rec.pop('properties')
+        properties = rec.pop('properties')
+        viewer_state = rec['state']['values']
+        viewer_state['filter_att'] = properties['filter_att']
+        viewer_state['filter'] = properties['filter']
 
         layer_states = []
 
--- a/glue/viewers/table/compat.py
+++ b/glue/viewers/table/compat.py
@@ -23,9 +23,6 @@ def update_table_viewer_state(rec, context):
         rec['state']['values'] = {}
 
         properties = rec.pop('properties')
-        viewer_state = rec['state']['values']
-        viewer_state['filter_att'] = properties['filter_att']
-        viewer_state['filter'] = properties['filter']
 
         layer_states = []
 
--- a/glue/viewers/table/qt/data_viewer.py
+++ b/glue/viewers/table/qt/data_viewer.py
@@ -293,15 +293,6 @@ class TableViewer(DataViewer):
         self.state.add_callback('layers', self._on_layers_changed)
         self._on_layers_changed()
 
-    def get_col(self, cid):
-        """
-        Get the column id for a ComponentID, necessary for QtFilterProxy
-        """
-        for i, comp in enumerate(self.model.columns):
-            if comp == cid:
-                return i
-        return -1  # This defaults to searching all columns
-
     def selection_changed(self):
         app = get_qapp()
         if app.queryKeyboardModifiers() == Qt.AltModifier:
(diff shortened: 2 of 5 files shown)


astrofrog:
Very nice! Let me know once you would like me to review this :) One request is whether we could hide the search box by default and have a viewer tool with a 'funnel' icon for filtering that when clicked would toggle the search box?

jfoster17:
> Very nice! Let me know once you would like me to review this :) One request is whether we could hide the search box by default and have a viewer tool with a 'funnel' icon for filtering that when clicked would toggle the search box?

I considered that. I think that it is better to have the search box by default because 

1. It does not take up that much space
2. The "default" reason to open up a Table Viewer is to identify specific rows (perhaps the most common reason is actually to get a look at your dataset, but finding specific items is probably the second most common reason)

Thus I think the friction of requiring a user to correctly identify and toggle the 'funnel' item for the most common interaction with a table is not worth the marginal cost of having a little bit less space in the viewer for showing the data.

jfoster17:
@astrofrog -- this is ready for your review.

I presented at our glue weekly meeting and they were fine with having the search bar be shown all the time. The main request was to be able to toggle the regex search so that users won't be tripped up if they want to search for unescaped special characters; I have now added this button.

glue/viewers/table/qt/data_viewer.py
```diff
@@ -260,8 +290,15 @@ def __init__(self, session, state=None, parent=None, widget=None):
         self.data = None
         self.model = None
 
-        self.ui.table.selection_changed.connect(self.selection_changed)
+        self._connection1 = connect_combo_selection(self.state, 'filter_att', self.ui.combosel_filter_att)
+        self._connection2 = connect_text(self.state, 'filter', self.ui.valuetext_filter)
+        self._connection3 = connect_checkable_button(self.state, 'regex', self.ui.bool_regex)
```
astrofrog, line 295: You could perhaps use autoconnect here like we do in a lot of the other widgets?

glue/viewers/table/compat.py
```diff
@@ -22,7 +22,7 @@ def update_table_viewer_state(rec, context):
         rec['state'] = {}
         rec['state']['values'] = {}
 
-        rec.pop('properties')
+        properties = rec.pop('properties')
```
astrofrog, line 25: If we don't use this, maybe best to write as ``_ = rec.pop(...)``

glue/viewers/table/qt/data_viewer.py
astrofrog: See https://echo.readthedocs.io/en/latest/api/echo.qt.autoconnect_callbacks_to_qt.html#echo.qt.autoconnect_callbacks_to_qt for more info

astrofrog (changes requested):
Looks great! Just a few small things

astrofrog (approved):
Looks great, thanks!

PR #2394, opened 2023-04-27
Avoid resetting limits when changing display units in profile viewer

astrofrog:
Should fix https://github.com/spacetelescope/jdaviz/issues/2126


pllim:
I think this broke Jdaviz. Maybe @kecnry unit conversion branch is good with this, but that is not yet in Jdaviz main. FYI.

PR #2399, opened 2023-05-08
Make sure __getitem__ always returns a numpy array

### LINKED ISSUE #2397: compute_fixed_resolution_buffer fails with some components powered by dask arrays
**Describe the bug**
In a real-world example (a large .zarr file) that I am reading in as a dask array, compute_fixed_resolution_buffer fails when called by the image viewer.

The culprit is this line:
https://github.com/glue-viz/glue/blob/84ed937df55bd52966f9e094b1a8849c6c9bd8a1/glue/core/fixed_resolution_buffer.py#L256
which is returning a dask array rather than a numpy array so that `array = array[translated_coords]` a few lines down fails with a warning that dask arrays do not support fancy indexing. 

`NotImplementedError: Don't yet support nd fancy indexing
`

One simple solution is simply to explicitly cast `array` as a `np.array`

I don't yet have a minimal reproducible version of this bug (and the example .zarr file is 35 GB), so right now this issue is mainly to remind me of this problem :) I can read in other .zarr files without a problem, so this is something about the dimensions, chunking, or something else as-of-yet-undetermined about this particular file and how it translates to dask arrays.

jfoster17:
# Pull Request Template

## Description

In some circumstances, dask.compute() can still return a dask array (rather than the desired numpy array). I was able to trigger this with an external data loader library that presumably falls foul of [this bug](https://stackoverflow.com/questions/56944723/why-sometimes-do-i-have-to-call-compute-twice-on-dask-delayed-functions). Explicit to np.array fixes this for my use case and seems otherwise totally innocuous. 

I am not providing a new unit test because I'm not sure how to reliably reproduce the extra-lazy dask array outside of the specific data loading library I was using.

Fixes #2397

### DIFF
--- a/glue/core/component.py
+++ b/glue/core/component.py
@@ -536,7 +536,7 @@ class DaskComponent(Component):
         return len(self._data.shape)
 
     def __getitem__(self, key):
-        return self._data[key].compute()
+        return np.array(self._data[key].compute())
 
     @property
     def numeric(self):


glue/core/component.py
```diff
@@ -536,7 +536,7 @@ def ndim(self):
         return len(self._data.shape)
 
     def __getitem__(self, key):
-        return self._data[key].compute()
+        return np.array(self._data[key].compute())
```
astrofrog, line 539: I think it would be better to use ``np.asarray`` in case the result of the computation is already a Numpy array, otherwise it will be unnecessarily copied in this case.

astrofrog (changes requested):
Thanks for tracking this down! Just a small comment below.

PR #2401, opened 2023-05-11
MNT: Remove unused .travis.yml

pllim:
I don't think you use Travis CI anymore? I don't see the badge in README. If you don't use it, might as well remove its YAML file.


astrofrog (approved):
Thanks!

PR #2403, opened 2023-05-11
FEAT: New CircularAnnulusROI class to represent circular annulus ROI

pllim:
This PR adds a new class to represent circular annulus as ROI.

Motivation: Even though we can use annulus now after https://github.com/glue-viz/glue-astronomy/pull/90 , it is not very user-friendly, as we have to expose parameters from the outer and inner circles separately, which would cause user to easily make the subset state no longer a circular annulus (see https://github.com/spacetelescope/jdaviz/pull/2182#issuecomment-1542882590 for exciting screenshots). Therefore, there is a need to represent this shape as ROI so that we only expose the relevant parameters (one center only, inner radius, outer radius) without a lot of hacky workaround downstream.

With this patch, I could do the following in Jdaviz after unpinning maxversion of glue-core (xref https://github.com/spacetelescope/jdaviz/issues/2183).

```diff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ dependencies = [
     "traitlets>=5.0.5",
     "bqplot>=0.12.37",
     "bqplot-image-gl>=1.4.11",
-    "glue-core>=1.6.0,!=1.9.0,<1.10",
+    "glue-core>=1.6.0,!=1.9.0",
     "glue-jupyter>=0.15.0",
     "echo>=0.5.0",
     "ipykernel>=6.19.4",
```

```python
import numpy as np
from glue.core.roi import CircularAnnulusROI
from glue.core.subset import RoiSubsetState

from jdaviz import Imviz

a = np.arange(100).reshape((10, 10))

imviz = Imviz()
imviz.load_data(a, data_label='a')
imviz.show()

data = imviz.app.data_collection[0]
ann = CircularAnnulusROI(xc=4.5, yc=4.5, inner_radius=2, outer_radius=4)
sbst = RoiSubsetState(xatt=data.pixel_component_ids[1], yatt=data.pixel_component_ids[0], roi=ann)
imviz.app.data_collection.new_subset_group(subset_state=sbst)
```

Screenshot 2023-05-11 162530

The next step following this PR is to support this new ROI both in glue-astronomy (https://github.com/glue-viz/glue-astronomy/pull/92) and Jdaviz.

[🐱](https://jira.stsci.edu/browse/JDAT-3380)

### DIFF
--- a/glue/core/roi.py
+++ b/glue/core/roi.py
@@ -13,7 +13,7 @@ from glue.utils import points_inside_poly, iterate_chunks, rotation_matrix_2d
 np.seterr(all='ignore')
 
 
-__all__ = ['Roi', 'RectangularROI', 'CircularROI', 'PolygonalROI',
+__all__ = ['Roi', 'RectangularROI', 'CircularROI', 'CircularAnnulusROI', 'PolygonalROI',
            'AbstractMplRoi', 'MplRectangularROI', 'MplCircularROI',
            'MplPolygonalROI', 'MplXRangeROI', 'MplYRangeROI',
            'XRangeROI', 'RangeROI', 'YRangeROI', 'VertexROIBase',
@@ -54,7 +54,7 @@ def pixel_to_axes(axes, x, y):
     return axes.transAxes.inverted().transform(xy)
 
 
-class Roi(object):  # pragma: no cover
+class Roi:  # pragma: no cover
 
     """
     A geometrical 2D region of interest.
@@ -215,7 +215,7 @@ class RectangularROI(Roi):
     """
 
     def __init__(self, xmin=None, xmax=None, ymin=None, ymax=None, theta=None):
-        super(RectangularROI, self).__init__()
+        super().__init__()
         self.xmin = xmin
         self.xmax = xmax
         self.ymin = ymin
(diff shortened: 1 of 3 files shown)


glue/core/roi.py
```diff
@@ -544,6 +543,90 @@ def __setgluestate__(cls, rec, context):
...
+            status = True
+        else:
+            status = False
+        return status
+
+    # FIXME: We need 2 polygons, with the inner one inverted, just like subset state.
+    def to_polygon(self):
+        if not self.defined():
+            return [], []
+        raise NotImplementedError("Cannot transform annulus to simple polygon")
```
astrofrog, line 604: Maybe better than nothing would be to transform it to a polygon like this:

Screenshot 2023-05-12 at 10 05 18

I've exaggerated the separation of the lines at the top, these could be actually over each other if you see what I mean. Does this make sense?

astrofrog (changes requested):
Looks good to me! I think it would be better to implement something for the polygon though, have left a comment below.

pllim:
Using the example posted above, this is what the polygon looks like:

```python
from matplotlib import pyplot as plt

x, y = ann.to_polygon()

plt.gca().set_aspect('equal')
plt.plot(x, y)
```

annulus_poly

astrofrog:
Great! Maybe we should have 100 points instead of 20?

pllim:
Done!

ann_poly_100

pllim:
Not sure why CI didn't kick off.

astrofrog (approved):
Looks great, thanks!

PR #2404, opened 2023-05-13
Add center and move_to methods to RangeSubsetState

pllim:
Motivation: Jdaviz currently has to have special logic to calculate center for `RangeSubsetState`. But after #2391 , there is no reason for it not to be part of the unifed `obj.center()` API to get center.


astrofrog (approved):
Looks good, thanks!

PR #2408, opened 2023-05-30
Fix filter logic when underlying data set is not visible

jfoster17:
# Fix Filtering and Underlying Data Not Visible

## Description

PR #2392 introduced a bug wherein making the underlying table dataset not visible did not remove it from the table model and display. This PR adds tests for this behavior and fixes the problem. 

A second PR to be posted shortly will fix the separate problem that under Qt6 you cannot re-enable a not-visible layer artist using the UI checkbox.


jfoster17:
Sorry, I made a garbage merge here. I'll just close this PR and open a new one!

PR #2410, opened 2023-05-30
Fix Qt6 bug that prevented layers from being re-enabled

jfoster17:
## Description

In Qt6, check boxes emit integers instead of the proper enum (https://bugreports.qt.io/browse/QTBUG-104688). Current logic for enabling/disabling layer artists checks the value emitted against the Qt.CheckState enum, which means that the user cannot make a layer artist they have turned not-visible visible again.

This PR fixes this behavior in the layer artist view by checking for the type of the value emitted by the checkbox and adds a test for it. I have not exhaustively checked that this problem does not occur with other checkboxes elsewhere in the GUI with Qt 6.


astrofrog (approved):
Oops, thanks for the fix!

PR #2411, opened 2023-05-30
Fix filter logic so that table viewer can hide main data layer

jfoster17:
## Description

PR https://github.com/glue-viz/glue/pull/2392 introduced a bug wherein making the underlying table dataset not visible did not remove it from the table model and display. This PR adds tests for this behavior and fixes the problem.


astrofrog (approved):
Thanks!

PR #2412, opened 2023-05-31
Ensure lazy dask arrays become numpy arrays in calculate_histogram

jfoster17:
# Pull Request Template

## Description

Similar to #2399, dask arrays read from disk with particularly lazy loaders may return dask arrays even after calling compute. This applies the same fix in calculate_histogram, which I missed in #2399 of calling np.asarray() to make sure we always get a numpy array when we want one.


astrofrog (approved):
Thanks!

PR #2415, opened 2023-06-01
Optionally 'freeze' subsets when removing a dataset from the DataCollection

astrofrog:
WIP


pllim (changes requested):
As discussed, I have concerns on memory usage and roundtripping with this approach.

astrofrog:
Yes it sounds like we aren't going down this road, so I will close

PR #2422, opened 2023-06-27
Set better starting xlim and ylim for custom viewers

### LINKED ISSUE #2420: Custom data viewer is not working
**Describe the bug**
Title

**To Reproduce**

Followed the instructions given here https://docs.glueviz.org/en/stable/customizing_guide/custom_viewer.html, ie added the first block of code to my `config.py`.

**Expected behavior**
Be able to load the `Shot Plot` new custom viewer

**Screenshots**
Captura de pantalla de 2023-06-13 11 32 30

**Details:**
Nothing happens as shown in the screenshot above, no errors in the console.

- Operating System: elementary 7 (based on Ubuntu 22.04.2 LTS)
- Python version (``python --version``): 3.11.3
- Glue version (``glue --version``): 1.11.0
- How you installed glue: pip

**Additional context**

jfoster17:
# Pull Request Template

## Description

As pointed out in #2420 the documentation example for a Custom Viewer (the Shot Chart) does not work well with current versions of glue because the Shot Chart x and y limits are set to 0 - 1, which contains no data. For new users this looks like it's just not working. This fix does not require any changes to the example scripts; we add a custom `add_data` function that uses the Matplotlib bounding box to set some sensible initial limits if we haven't already set limits (which the example later does in a `setup` function).

Note that we can't use `axes.relim()` for this because hexbin returns a Collection [which is not supported](https://github.com/matplotlib/matplotlib/blob/v3.7.1/lib/matplotlib/axes/_base.py#L2459-L2483).

Closes #2420

### DIFF
--- a/glue/viewers/custom/qt/custom_viewer.py
+++ b/glue/viewers/custom/qt/custom_viewer.py
@@ -74,6 +74,7 @@ import numpy as np
 from qtpy.QtWidgets import QWidget, QGridLayout, QLabel
 
 from echo.qt import autoconnect_callbacks_to_qt
+from echo import ignore_callback
 
 from glue.config import qt_client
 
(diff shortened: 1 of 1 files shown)


glue/viewers/custom/qt/custom_viewer.py
```diff
@@ -752,6 +753,25 @@ def apply_roi(self, roi):
...
+        most custom viewers will show data as expected.
+        """
+        super(CustomMatplotlibDataViewer, self).add_data(data)
+        if (self.state.x_min == 0 and self.state.x_max == 1 and self.state.y_min == 0 and self.state.y_max == 1):
+            with ignore_callback(self.state, "x_min", "x_max", "y_min", "y_max"):
+                self.state.x_min = self.axes.dataLim.xmin
+                self.state.x_max = self.axes.dataLim.xmax
+                self.state.y_min = self.axes.dataLim.ymin
+                self.state.y_max = self.axes.dataLim.ymax
+        self.limits_to_mpl()
```
astrofrog, line 772: Just curious, why is this needed?

astrofrog (approved):
Looks good to me! Just a quick question below.

glue/viewers/custom/qt/custom_viewer.py
jfoster17: Sorry, which line(s) are you asking about? The general idea is that without something like this the limits in the state object stay at the default values of (0,1) for custom viewers rather than getting updated when data (which is normally handled by limit helpers).

astrofrog: I meant the last line - since we are updating the state from the Matplotlib viewer, why do we then need to convert the state limits back to Matplotlib? Is it in case we make any changes in the state because of aspect ratio/etc?

astrofrog: In any case I'll go ahead and merge, we can always tweak later if you think this isn't needed but I don't think there's any harm in doing it.

PR #2425, opened 2023-07-14
Fix bug that caused incompatible subsets to be rendered in some contexts rather than be hidden

### LINKED ISSUE #2405: Incompatible Subsets use too much memory
**Describe the bug**
When an image viewer is open, creating an Incompatible Subset (that is, a subset defined over attributes not present in the reference_data of the image viewer) will create an extra array that is the full size of reference_data.

**To Reproduce**
From the glue terminal:
```
import numpy as np
from glue.core import Data
im = Data(label='data1',x=np.arange(100_000_000).reshape((10000, 10000)))
catalog = Data(label='catalog', c=[1, 3, 2], d=[4, 3, 3])
dc.append(im)
dc.append(catalog)
from glue.viewers.image.qt.data_viewer import ImageViewer
viewer = session.application.new_data_viewer(ImageViewer)
viewer.add_data(im)
```
glue should be using about 1 GB of memory at this point. Creating a subset that can be shown:
```
dc.new_subset_group(subset_state=im.id['x'] > 50_000_000, label='A')
```
does not appreciably increase the memory usage of glue. However, creating a subset that cannot be shown:
```
dc.new_subset_group(subset_state=catalog.id['c'] > 2, label='B')
```
Causes memory usage to ~double to 2 GB. New Incompatible Subsets continue to grow the memory used.

**Expected behavior**
Incompatible subsets should not take up more memory that normal subsets.

**Additional context**
`ImageSubsetArray` is called when a new subset tries to show itself on an Image Viewer. The `__call__` method creates a broadcasted array of `np.nan` that is the full size of the Image Viewer reference_data. This array is then used in the `make_image` function of `mpl_scatter_density.base_image_artist` line 190 (`self.set_data(array)`) which causes the broadcasted array to materialize fully in memory. 

I think the problem would essentially be solved by returning just the portion of the large nan array defined by bounds (if we don't trigger the IncompatibleAttribute we get a mask with the shape of bounds), but I confess I don't really understand why we're making a potentially giant nan array here in the first place, so I wanted to open a discussion before trying to fix this.

jfoster17:
# Pull Request Template

## Description

This is my attempt that fixes https://github.com/glue-viz/glue-jupyter/issues/367 and is at least a reasonable solution to #2405. I'm still not 100% sure that this is correct approach, but it seems to work and passes all current tests. 

- [x] Understand the workflow that requires us to compute the mask here for IncompatibleAttributes and make sure this solution still works for that.

Fixes #2405 
Fixes https://github.com/glue-viz/glue-jupyter/issues/367


astrofrog:
@jfoster17 regarding your question in #2405 about why we create a large NaN array, I have to admit I'm not sure anymore why we had to do this - but it might very much be the case that we don't need to anymore so if everything works fine with the approach in this PR (both in Qt and Jupyter glue) then I'm ok with the solution!

jfoster17:
After poking at this a bit more, we don't actually have to return anything here for incompatible layers, which is cleaner.

It's not easy to write a test for this in glue-core, since this is sort of inherently a front-end thing, but one that hits both glue-core and glue-qt. @astrofrog : Thoughts on how to proceed with testing this?

(The current test failures are unrelated and due to the new version of Matplotlib 3.8.0)

astrofrog:
This works, thanks! I tried setting up a regression test in glue-jupyter: https://github.com/glue-viz/glue-jupyter/pull/404 - however for some reason while the test seems to reproduce the failure locally, it does not when run on CircleCI. In any case we should merge and release this.

PR #2427, opened 2023-08-07
Render nans as 'bad' in mpl cmaps when mode='colormap'

bmorris3:
## Description

By default, image viewers cast NaNs to zero before applying a colormap. This prevents the use of the matplotlib "bad" feature. For each colormap, values above or below the colormap vmin/vmax are rendered with an optionally specified value, and "bad" values (usually nans) are rendered by a third value:
```python
import matplotlib.pyplot as plt

plt.cm.viridis
```
Screen Shot 2023-08-07 at 12 39 22

By default, `bad` is `color='k', alpha=0`, but this can be changed with:
```python
plt.cm.viridis.set_bad(color='r')
```
which allows users to clearly visualize nans efficiently and without special handling.

Glue currently forces nans to zero before they are converted to colors by the colormap:
https://github.com/glue-viz/glue/blob/0f83a03c997b78476043485f4145f6e69cb8a8cf/glue/viewers/image/composite_array.py#L140
which prevents the user from ever making use of the `bad` functionality.

This PR allows the user to set a private attribute on the `CompositeArray` class called `_set_nans_to_zero`. When True, the behavior is unchanged. When set to False by a user, nans are not forced to zero, and the `bad` values are rendered as described by the mpl colormap. An example is shown below using Imviz from jdaviz:

Screen Shot 2023-08-07 at 12 43 43

### DIFF
--- a/glue/viewers/image/composite_array.py
+++ b/glue/viewers/image/composite_array.py
@@ -29,7 +29,6 @@ class CompositeArray(object):
 
         self._first = True
         self._mode = 'color'
-        self._set_nans_to_zero = True
 
     @property
     def mode(self):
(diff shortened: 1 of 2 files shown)


glue/viewers/image/composite_array.py
```diff
@@ -137,7 +138,8 @@ def __call__(self, bounds=None):
             data = interval(array)
             data = contrast_bias(data, out=data)
             data = stretch(data, out=data)
-            data[np.isnan(data)] = 0
+            if self._set_nans_to_zero:
+                data[np.isnan(data)] = 0
```
dhomeier, line 142: This was only introduced in fa60bf1 for compatibility with Astropy; if we can reproduce what the issue with Astropy was, preserving NaN may perhaps be made the default again.

bmorris3: @astrofrog Any memory/tips of what we should check? Thanks!

astrofrog: I *think* it was actually that astropy did not preserve NaNs and then started to so it's not an issue with astropy as such - but more that when you do color compositing you don't really have the option to give NaN values their own color since each channel has only one color assigned to it. So I think it's as I mentioned on the call - for multi-color compositing we should always set NaNs to 0 and for the colormap we could never set it to zero.

bmorris3:
As of [e01db76](https://github.com/glue-viz/glue/pull/2427/commits/e01db76863357b3d3e4b7b2b0f3110cfd6d58315), nans are forced to zero when `mode='color'`, and no zeros are forced when `mode='colormap'`.

astrofrog:
It looks like some tests will need to be updated

bmorris3:
@astrofrog All fixed!

astrofrog (approved):
Looks good, thanks!

Issue #2428, opened 2023-08-08
Tests hanging with serialisation failure in `test_save_aggregate_slice`


dhomeier:
**Describe the bug**
Almost all tests on Python > 3.8 (except the `dev` ones) have been hanging and timed out after 6 hours for several weeks (basically since 1.12.0, it seems) in test_data_viewer, e.g.
https://github.com/glue-viz/glue/actions/runs/5393566950/job/14600937184

Running locally I found a failure in
```python
__________________________________________________________ TestImageViewer.test_save_aggregate_slice ___________________________________________________________

self = <glue.viewers.image.qt.tests.test_data_viewer.TestImageViewer object at 0x15f5238d0>
tmpdir = local('/private/var/folders/bb/f_z2vsjd21d_p0jpqxzhhwhh00027w/T/pytest-of-derek/pytest-5/test_save_aggregate_slice0')

    def test_save_aggregate_slice(self, tmpdir):
    
        # Regression test to make sure that image viewers that include
        # aggregate slice objects in the slices can be saved/restored
    
        self.viewer.add_data(self.hypercube)
        self.viewer.state.slices = AggregateSlice(slice(1, 3), 10, np.sum), 3, 0, 0
    
        filename = tmpdir.join('session.glu').strpath
    
        self.application.save_session(filename)
        self.application.close()
    
>       app2 = GlueApplication.restore_session(filename)

glue/viewers/image/qt/tests/test_data_viewer.py:679: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
glue/app/qt/application.py:1135: in restore_session
    ga = Application.restore_session(path)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[...]

Issue #2353, opened 2023-01-24
`GlueSerializer` fails on re-wrapped `__array_function__` dispatcher in numpy-dev


dhomeier:
**Describe the bug**
A new wrapper for dispatching `__array_function__` introduced in numpy/numpy#23020 fails to be serialised with
```python
glue.core.state.GlueSerializeError: Don't know how to serialize <function sum at 0x104aa4ed0> of type <class 'numpy._ArrayFunctionDispatcher'>
```
**To Reproduce**
Further details in #2352

dhomeier:
Followed up in #2428

PR #2429, opened 2023-08-08
Skip `test_save_aggregate_slice` for Numpy 1.25.x and unpin numpy-dev

dhomeier:
## Description
The `GlueSerializer` failure/hang first reported in #2353 has now made it into Numpy 1.25, so to keep jobs from just timing out (and burning CI hours), this pins all tests to 1.24.x as a temporary workaround.
Alternatively one could skip `test_save_aggregate_slice` conditionally on Numpy >= 1.25, but that would leave that test to be run in only one or two tests – but perhaps better, than lagging by an entire Numpy release with everything else?
Might reactivate an `lts` job with `numpy==1.24.*`...

Workaround for #2428

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -45,6 +45,7 @@ jobs:
         - linux: py310-test-pyqt63-all
         - linux: py310-test-pyqt64-all
         - linux: py311-test-pyqt514
+        - linux: py311-test-pyqt515-lts-all
 
         # Documentation build
         - linux: py38-docs-pyqt514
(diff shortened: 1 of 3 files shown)


astrofrog:
I think I would rather skip one test if possible?

dhomeier:
Done. It's still tested in the py38 and py311-lts envs now.

Any thoughts how to address the actual `GlueSerializer` failure?
Something like https://github.com/numpy/numpy/pull/23034 (and if so, would this have to be done upstream)?

astrofrog (approved):

Issue #2430, opened 2023-08-10
Split glue-core into glue-core and glue-qt


astrofrog:
Now that glue-jupyter is picking up steam, we should press ahead with the plan to split glue-core out into glue-core and glue-qt, so that packages relying on glue-core for Jupyter or other front-ends don't need to be importing anything needed for the Qt Application.

After much thought, I think that the present repository should become the one for glue-core, the frontend-independent package, because it is the one that will likely be around for the duration of the project, and glue-qt should be a separate repository. **For now** I think we should also not rename this present repo to glue-core (but we could in future).

Given this, my proposed plan is as follows:

* [x] Split out all the Qt-specific code into a separate glue-qt repository which will provide a glue-qt package with a glue_qt module. That repository should also contain most of the present docs, since they are very much centered around the Qt application currently. The glue-qt repository should share the same history as the present repository as - well - they do! (and will make it easier to transition PRs from one to the other) (https://github.com/glue-viz/glue-qt/pull/1)
* [x] Remove all the Qt-specific code from the present repository (https://github.com/glue-viz/glue/pull/2431)
* [x] Add files to the glue repository which are compatibility imports from glue-qt to glue-core, with deprecation warnings, for now.
* [x] Update the glueviz package to depend on glue-core and glue-qt so that users using the glueviz package should see no change
* [ ] Once we have checked that everything is working fine and that none of the glue ecosystem packages use any of the deprecated imports, do releases of all the packages that have been updated as well as glue-core and glue-qt
* [x] Update docs.glueviz.org to point to the glue-qt docs
* [ ] Over time we can start moving glue-qt-specific issues over to glue-qt, and moving some of the core (non-Qt) docs over from glue-qt to glue.

At this point, everything should continue to work seamlessly for all users although deprecation warnings will be emitted. Advertise to users that this change is happening.

Once we are confident that we have given enough time for people to adapt (could be a few months), we can drop the glue-qt depdendency from glue-core and also remove all the qt/ sub-packages. This will warrant bumping the version of glue-core to 2.0.0 since it is a pretty major API change.

astrofrog:
On second thoughts, I think perhaps the glue repo for now should continue to be the Qt specific one and have a new glue-core repository. This would minimize any disruption related to the docs and issue tracker etc. The actual package name in the repo would still have to change to glue-qt instead of glue-core (which it is right now) but I think that should be ok.

We can then move over any issues and docs that are not specific to the glue-core repository over time.

[...]

PR #2431, opened 2023-08-10
Remove Qt-specific code

astrofrog:
This is part of https://github.com/glue-viz/glue/issues/2430 - no plan to merge anytime soon as need to then try out setting up glue-qt and also make sure that all downstream packages that don't need Qt can work with this update. But just opening it so I can keep track of it easily.


glue/core/cli_parsers/__init__.py
```diff
@@ -0,0 +1 @@
+from .parsers import parse_data
```
dhomeier, line 1:
```suggestion
from .parsers import parse_data  # noqa: F401
```
need a codestyle protection here?

dhomeier (changes requested):
> doc/conf.py:6:1: F401 'sys' imported but unused

also keeping the tests from running

glue/core/cli_parsers/__init__.py
astrofrog: Thanks! I just pushed a commit I had locally that already fixed this.

Carifio24:
I think `glue-wwt` should be in the "needs updating" list as well, as it has some Qt-based glue imports.

dhomeier (approved):

dhomeier:
Tests for glue-jupyter are passing with this version and no Qt installed (except for those requiring glue-vispy-viewers, which cannot be installed without PyQt).

astrofrog:
@dhomeier - hmm good point about glue-vispy-viewers - I think we should probably move the 3D state objects out of glue-vispy-viewers and into glue-core since they should be reasonably generic, and then we could depend on that instead of glue-vispy-viewers in glue-jupyter.

dhomeier:
> except for those requiring glue-vispy-viewers, which cannot be installed without PyQt).

The depency is only on conda-forge though; with pip-installed glue-vispy-viewers the other tests are passing as well...

astrofrog:
I think the thing is that glue-jupyter does depend on glue-vispy-viewers for the 3D volume rendering viewer but we just haven't declared the dependency. However we should be able to refactor things to not have that dependency at all. I think I will delay this until after glue-core and glue-qt are split.

astrofrog:
I'm going to go ahead and merge this as I think it is ready. I will also merge https://github.com/glue-viz/glue-qt/pull/1 shortly. I will then do a release of glue-qt and add it as a dependency in glue-core so that glue-core should then continue to function as before just with deprecation warnings being raised now when Qt-related functionality is accessed.

PR #2444, opened 2023-09-19
Do not use fancy index for subsets over Dask data

jfoster17:
# Pull Request Template

## Description

Fix FRB so that we can reliably compute subsets over data with Dask components.


astrofrog (approved):
Rebased to get rid of the merge commit, but looks good to me otherwise! Let's merge if CI passes.

PR #2446, opened 2023-09-19
Support dynamic components

jfoster17:
# Pull Request Template

## Description

This is basically behind-the-scenes stuff to enable some functionality in glue-genes whereby some components update their values automatically in response to subset definitions changing and we want viewers to be able to react to these updated values. This should have no impact on existing glue viewers/etc.

### DIFF
--- a/glue/core/layer_artist.py
+++ b/glue/core/layer_artist.py
@@ -228,9 +228,9 @@ class LayerArtistBase(PropertySetMixin, metaclass=ABCMeta):
             self._changed = True
             self._state = state
 
-    def update_component_limits(self, components_changed):
+    def _on_components_changed(self, components_changed):
         """
-        Update component limits for this layer
+        React to a change to one or more of the components in this layer.
         """
         pass
 
--- a/glue/core/layer_artist.py
+++ b/glue/core/layer_artist.py
@@ -228,6 +228,12 @@ class LayerArtistBase(PropertySetMixin, metaclass=ABCMeta):
             self._changed = True
             self._state = state
 
+    def update_component_limits(self, components_changed):
+        """
+        Update component limits for this layer
+        """
+        pass
+
     def __str__(self):
         return "%s for %s" % (self.__class__.__name__, self.layer.label)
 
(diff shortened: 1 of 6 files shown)


glue/core/layer_artist.py
```diff
@@ -228,6 +228,12 @@ def _check_subset_state_changed(self):
             self._changed = True
             self._state = state
 
+    def update_component_limits(self, components_changed):
```
astrofrog, line 231: Can you give a real life example of what this method would do? I wonder if this should be a more general method such as ``on_component_change`` or something similar that could be more versatile?

jfoster17: Sure -- in glue genes we have a SyncComponent which is a component that is added to a dataset when a subset is created and is updated as that subset is updated. For instance, we can calculate gene expression levels in a spatial dataset for a given subset of genes; as we change the subset of genes, we update the values in the SyncComponent on the spatial dataset, but we need to prompt the cmap_lim_helpers so that they display the full range of values in the new component. 

Thus, a Viewer can add something like this:

```
@defer_draw
    def update_component_limits(self, components_changed):
        for limit_helper in [self.state.cmap_lim_helper]:
            if limit_helper.attribute in components_changed:
                limit_helper.update_values(force=True)
        self.redraw()
```

And the user experience is nice because changing the subset automatically scales the color to show the range of values. 
Screen Recording 2023-09-21 at 5 10 13 PM

This could definitely be called something more general; I just had the one example and I thought a more specific name would be more readable in that case.

astrofrog: Perhaps the main 'hook' that gets called could be called with a more general name, and in your case it could then call from there ``update_component_limits``?

jfoster17: Yeah, sounds good. I've updated to this approach.

jfoster17:
@astrofrog -- are we good to merge this now?

astrofrog (approved):
Yes sorry for dropping the ball on this - thanks!

PR #2451, opened 2023-10-26
Fix bug when changing the number of bins on a histogram with modified/deleted data

astrofrog:
This fixes an issue that was seen in https://github.com/spacetelescope/jdaviz/pull/2529 which is that if we change the dataset used for the x attribute for the histogram, this failed if the original dataset was removed from the data collection or if the new dataset was not linked to the original one.


astrofrog:
The docs failure is unrelated

PR #2453, opened 2023-10-26
Make stretches be customizable on a layer by layer basis

astrofrog:
For the image and scatter viewers, this adds a new dictionary property ``stretch_parameters`` which can be used to set stretch parameters specific to the layer. These parameters are then set as attribute on state instances. So the requirement for new stretch classes is that parameters can be set as attributes on existing instances (e.g. ``LogStretch.exp``). Example:

Screenshot from 2023-10-26 15-25-44

Glue deals with keeping ``stretch_parameters`` in sync with the stretch instance/object, and will deal with refreshing the plot.

### DIFF
--- a/glue/viewers/common/stretch_state_mixin.py
+++ b/glue/viewers/common/stretch_state_mixin.py
@@ -1,0 +1,47 @@
+from glue.config import stretches
+from glue.viewers.matplotlib.state import (
+    DeferredDrawDictCallbackProperty as DDDCProperty,
+    DeferredDrawSelectionCallbackProperty as DDSCProperty,
+)
+
+__all__ = ["StretchStateMixin"]
+
+
+class StretchStateMixin:
+    stretch = DDSCProperty(
+        docstring="The stretch used to render the layer, "
+        "which should be one of ``linear``, "
+        "``sqrt``, ``log``, or ``arcsinh``"
+    )
+    stretch_parameters = DDDCProperty(
+        docstring="Keyword arguments to pass to the stretch"
+    )
+
+    _stretch_set_up = False
+
+    def setup_stretch_callback(self):
+        type(self).stretch.set_choices(self, list(stretches.members))
+        type(self).stretch.set_display_func(self, stretches.display_func)
+        self._reset_stretch()
+        self.add_callback("stretch", self._reset_stretch)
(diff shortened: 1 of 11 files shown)


glue/viewers/common/stretch_state_mixin.py
```diff
@@ -0,0 +1,47 @@
...
+        return self._stretch_object
+
+    def _sync_stretch_parameters(self, *args):
+        for key, value in self.stretch_parameters.items():
+            if hasattr(self._stretch_object, key):
+                setattr(self._stretch_object, key, value)
+            else:
+                raise ValueError(
+                    f"Stretch object {self._stretch_object.__class__.__name__} has no attribute {key}"
+                )
```
kecnry, line 43: If a stretch accepts multiple attributes but needs to validate them wrt each other (for example: x/y coordinates of knot locations for a spline), what would be the best way to handle that?

As this stands, those will need to be passed as a single entry (`stretch_parameters['knots'] = ([x1, x2], [y1, y2])`) so that validation can occur in a _single_ setter on the stretch object.  

Alternative ideas:
* support some sort of context manager (similar to `hold_sync`) that validates the attributes after setting all of them at the end of this for-loop?  This of course needs to be compatible with the astropy stretch objects that are used by default as we probably can't modify those upstream just to account for this.
* instead of looping and checking for `hasattr` on the individual keys, could we instead require a `update_parameters` method (if `stretch_parameters` is not empty) on the stretch object which handles this logic internally by getting the _entire_ dictionary in one pass?

cc @haticekaratay

astrofrog: Just to put another idea out there, could you perhaps not validate when the attributes are set but when the stretch is called?

kecnry: That's an option, but probably causes unnecessary cost when calling the stretch repeatedly.  In our case, we can probably just validate externally since we manage the stretch object, but then if a user modifies the values, they will need to make sure they make sense.

pllim:
Is this patch not backward compatible?

astrofrog:
I did try and make it backward compatible but may have missed something. I will investigate the jdaviz failures tomorrow.

PR #2455, opened 2023-10-31
Customizable limits percentile and behavior on updating bins when resetting limits

astrofrog:
This makes it possible to specify the percentile to use when auto-determining limits, and makes it also possible to avoid resetting bins when calling ``reset_limits()``.


dhomeier (approved):
Code and tests look ok, taking https://github.com/spacetelescope/jdaviz/pull/2529#issuecomment-1787978441 as confirmation that this works.

PR #2456, opened 2023-10-31
Support regions in imageviewer

jfoster17:
# Add Region Layer Artist to ImageViewer

## Description

This adds a `ScatterRegionLayerArtist` to the ImageViewer so that `RegionData` datasets are plotted as regions rather than points. A parallel PR for glue-qt is required to actually make this work.


jfoster17:
The codecov report here is misleading because we actually test out this code in the associated [glue-qt PR](https://github.com/glue-viz/glue-qt/pull/10). We don't generally have standalone non-qt state or layer_artist tests because they work mostly in the context of an actual viewer. 

This PR does not enable exporting a RegionScatterLayer as a python script. Currently RegionScatterLayers are only implemented as a layer on top of an ImageViewer where export-to-python [does not work with non-image-layers](https://github.com/glue-viz/glue/issues/2457) anyway.

astrofrog (approved):
Looks good to me! I do want to start being able to add tests in glue-core for this kind of functionality, and I'll open a PR soon to set up some simple bare-bones Matplotlib-only 'viewers' that can be used to write tests against.

PR #2465, opened 2023-12-05
Provide stronger checking for region display. Fix x/y flip.

### LINKED ISSUE #2464: X and Y problems with ScatterRegions
**Describe the bug**
There are two related problems with the current implementation of ScatterRegions on the image viewer. In the following, the `RegionData` region_data has regions describe by (x', y') coordinate pairs and thus centroid attributes x' and y' while the image_data has x and y pixel coordinates (and optionally wx and wy world coordinates).

1. If the user links x' <-> x and y' <-> y, everything is good. If the user links x' <-> y and y' <-> x (perhaps because the coordinates in region_data come from a data source with ambiguous names (`spatial_0` and `spatial_1` are common in biology) then the regions will *still* display correctly on the image, because we only check if the viewer x_att and y_att are in the list of special region centroid coordinates [x',y'] and we only display the regions naively assuming that x_att is the same as x', but subsets defined _on the image_ will select the wrong points on region_data. 
2. If the user switches the x_att and y_att in the ImageViewer options so that x_att = y and y_att = x, the regions will not display correctly. 

Probably the same improved logic for checking viewer attributes will work for both these cases, but case 2 will require a transformation of the region geometries (as we currently do for images with world coordinates).

jfoster17:
# Pull Request Template

## Description

Fixes #2464 

We now directly check that the x/y-axis in the region specifications is the same as a the x/y-axis displayed in the image viewer. This allows us to correctly display the regions when the image axes are flipped and prevents erroneous display of the regions if the linking was incorrectly set up.

### DIFF
--- a/glue/tests/visual/py311-test-visual.json
+++ b/glue/tests/visual/py311-test-visual.json
@@ -1,7 +1,8 @@
 {
   "glue.viewers.histogram.tests.test_viewer.test_simple_viewer": "cb08123fbad135ab614bb7ec13475fcc83321057d884fe80c3a32970b2d14762",
   "glue.viewers.image.tests.test_viewer.test_simple_viewer": "72abd60b484d14f721254f027bb0ab9b36245d5db77eb87693f4dd9998fd28be",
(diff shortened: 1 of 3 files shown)


astrofrog:
I've rebased this and will try and merge/release today.

Issue #2470, opened 2023-12-14
TST: Python 3.12


pllim:
There is no py312 in your test suite even though Python 3.12 is out and widely available now.

PR #2471, opened 2023-12-15
Compatibility and tests against Numpy 2.0.0rc1

dhomeier:
## Description
Installing numpy dev in CI is currently blocked by various upstream packages pinned to `numpy < 2.0`; adding those to the developer versions.
This would update both `dev` jobs to `numpy>=0.0.dev0`; restrict to a new env `npdev` instead?

- [x] ~~pandas has no compatible wheel and a hard dependency on `numpy < 2.0`~~.
- [x] ~~contourpy has no compatible wheel and a hard dependency on `numpy < 2.0`~~.
- [x] ~~shapely has no compatible wheel for `numpy < 2.0`~~ (shapely/shapely#1949), build depends on libgeos.
- [x] ~~fast-histogram has no compatible wheel~~.
- [x] casa-formats-io has no compatible wheel ~~and a pinned dependency on `numpy < 2.0`~~.

Only workaround for the pins seems to be to install with `--no-deps` (pandas-dev/pandas#55488) which in turn requires every single test dependency to be pulled in explicitly; if anyone has an idea how to run a first pass without `--no-deps` an then a second `pip install` to override the numpy pin, that could be simplified a bit.

Building shapely seems to require compiling its own `libgeos` installation per its
[install script](https://github.com/shapely/shapely/blob/main/ci/install_geos.sh); unless I find a way to get an apt install for that, this is a showstopper for now.
(It seems to be available, but would require pointing the CI to fetch from an [alternative repo](https://libgeos.org/usage/install/#ubuntu)).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -21,11 +21,9 @@ changedir =
     test: .tmp/{envname}
     docs: doc
 deps =
-    dev: numpy==2.0.0rc1
-    dev: scipy==1.13.0
-    dev: astropy==6.1.0rc1
-    dev: shapely>=0.0.dev0
-    dev: git+https://github.com/radio-astro-tools/casa-formats-io.git
+    dev: numpy>=0.0.dev0
+    dev: scipy>=0.0.dev0
+    dev: astropy>=0.0.dev0
     # LTS
     lts: astropy==5.0.*
     lts: matplotlib==3.5.*
@@ -55,7 +53,7 @@ extras =
 # for as long all test deps need to be manually pulled in above as well!
 install_command =
     !dev: python -I -m pip install
-    dev: python -I -m pip install -v --pre
+    dev: python -I -m pip install -v
 commands =
     test: pip freeze
     test: pytest --pyargs glue --cov glue --cov-config={toxinidir}/setup.cfg {env:MPLFLAGS} {posargs}
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{38,39,310,311}-{codestyle,test,docs}-all-{dev,legacy}{,-visual}
+    py{38,39,310,311,312}-{codestyle,test,docs}-all-{dev,legacy}{,-visual}
 requires = pip >= 18.0
            setuptools >= 30.3.0
 
@@ -10,7 +10,7 @@ passenv =
     DISPLAY
     HOME
 setenv =
-    dev: PIP_EXTRA_INDEX_URL = https://pypi.anaconda.org/astropy/simple https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
+    dev: PIP_EXTRA_INDEX_URL = https://pypi.anaconda.org/astropy/simple https://pypi.anaconda.org/liberfa/simple https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
     visual: MPLFLAGS = -m "mpl_image_compare" --mpl --mpl-generate-summary=html --mpl-results-path={toxinidir}/results --mpl-hash-library={toxinidir}/glue/tests/visual/{envname}.json --mpl-baseline-path=https://raw.githubusercontent.com/glue-viz/glue-core-visual-tests/images/{envname}/
 whitelist_externals =
     find
@@ -22,7 +22,12 @@ changedir =
     docs: doc
 deps =
     dev: numpy>=0.0.dev0
+    dev: scipy>=0.0.dev0
     dev: astropy>=0.0.dev0
+    dev: pyerfa>=0.0.dev0
+    dev: matplotlib>=0.0.dev0
+    dev: git+https://github.com/contourpy/contourpy.git
+    dev: git+https://github.com/pandas-dev/pandas.git
     lts: astropy==5.0.*
     lts: matplotlib==3.5.*
     # Pin numpy-lts until permanent solution for #2353/#2428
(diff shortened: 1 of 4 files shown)


pllim:
For pandas and matplotlib, I did a hack in my packages to let them downgrade numpy first but then force numpy dev reinstall right before test start running and ignore the pip warning. It's not pretty but easier than --no-deps . 

shapely is indeed more troublesome. They never replied to my request.

dhomeier:
@pllim & @jorisvandenbossche thanks for following up on the `shapely` update; this is now working with the wheel from scientific-python-nightly although I still needed to put a hard pin on the version until a `2.1.0.dev0` is added in shapely/shapely#1999 – so the revision might not be very durable.

@astrofrog I think the `numpy<2.0` pin on `casa-formats-io` could be removed as the main branch installed with numpy-dev without problems (though maybe needs to be tested on more platforms and Python versions); this is the only devdep left forcing us to install with `--no-deps` and all the resultant hassle.
And of course would be good to have numpy-2.0 wheels for that and `fast_histogram` at some point.

dhomeier:
Starting to test Python 3.12 as well addressing #2470; [Windows data_factory confusion](https://github.com/glue-viz/glue/actions/runs/8071751779/job/22052086460?pr=2471#step:10:301) to be investigated.

jorisvandenbossche:
I just uploaded wheels that should have the higher version number (https://anaconda.org/scientific-python-nightly-wheels/shapely/files), that _should_ fix the issue with having to pin

dhomeier:
> I just uploaded wheels that should have the higher version number (https://anaconda.org/scientific-python-nightly-wheels/shapely/files), that _should_ fix the issue with having to pin

Yes, everything works with `>=0.0.dev0` now.
Thanks again!

astrofrog:
Just FYI I'm working on casa-formats-io (https://github.com/radio-astro-tools/casa-formats-io/pull/61) and fast-histogram is currently building wheels for a new Numpy 2.0-compatible release.

dhomeier:
Running now with numpy 2.0.0rc1, astropy 6.1.0rc1 and scipy 1.13.0; fast-histogram still needs its own build (see https://github.com/astrofrog/fast-histogram/pull/86#issuecomment-2059600339); have not checked for released versions of other dependencies yet.
The actual glue test failures all seem to be on `np.array(copy=False)` like
https://github.com/glue-viz/glue/actions/runs/8709773487/job/23890261129?pr=2471#step:10:311
The fix looks straightforward enough, going to look into that next.

dhomeier:
@astrofrog I have used a simplified version of the Scipy fix, since we should not need to worry about earlier numpy 2.0.0dev versions that did not support `copy=None` yet.
The new casa-formats-io wheel on PyPI still depends on and was built against numpy 1.26 – I could not follow the cibuildwheel process entirely, but it is also testing against that:
https://github.com/radio-astro-tools/casa-formats-io/actions/runs/8720397145/job/23921900123#step:7:371
perhaps need to remove the `python_version<'3.9'` case from `pyproject.toml` altogether?

Shapely has a new release built on shapely/shapely#2033 but seems to suffer the same problems with wheels for older Pythons being pulled in. For all other dependencies the current releases or candidates now work. 🚀

Before merging we should decide which `0.0.dev0` versions to keep in the `dev` env (numpy, scipy, astropy?)

Don't have a way to debug the unrelated Windows failures.

astrofrog:
@dhomeier - are you sure it's still not working with casa-formats-io v0.3.0? There is a devdeps job there that shows that it works fine with numpy 2.0.0rc1 at runtime: https://github.com/radio-astro-tools/casa-formats-io/actions/runs/8720397149/job/23921906624

tox.ini
```diff
@@ -22,7 +22,11 @@ changedir =
     docs: doc
 deps =
     dev: numpy>=0.0.dev0
+    dev: scipy>=0.0.dev0
     dev: astropy>=0.0.dev0
+    dev: pandas>=0.0.dev0
+    dev: git+https://github.com/radio-astro-tools/casa-formats-io.git
```
astrofrog, line 28: Can you double check that v0.3.0 works now?

dhomeier: I've tried to pip install on macOS several times now, always getting a numpy 1.26 version

dhomeier: Maybe a caching problem just as with shapely; seems to be running here as well, but then the installed versions of other deps don't make any sense with what is in `tox.ini` now. Rebasing to try to get this sorted...

dhomeier:
https://github.com/glue-viz/glue/actions/runs/8723259903/job/23931144239?pr=2471 looks OK now

tox.ini
```diff
@@ -46,6 +48,12 @@ extras =
     all: all
     docs: docs
     visual: visualtest
+# Need `--pre` for packages like pandas having no released version supporting numpy>=2.0 yet,
+# + `--no-deps` for casa-formats-io having no branch for numpy>=2.0 -
+# for as long all test deps need to be manually pulled in above as well!
+install_command =
+    !dev: python -I -m pip install
+    dev: python -I -m pip install -v
```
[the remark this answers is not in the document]
dhomeier: Just removed this as we should no longer need any git installs; the `>=0.0.dev0` versions apparently install without it 🤪

astrofrog (approved):

astrofrog:
@dhomeier is this ok to merge? (If so feel free to!)

dhomeier:
Yes, think so – I made on run with pandas dev to check the Windows failures (still persisted), but don't think we need that in general; thanks!

PR #2476, opened 2024-01-31
Allow setting slices or `subset_state` for determining min/max levels in `StateAttributeLimitsHelper`

dhomeier:
## Description
This enables passing a `subset_state` to the `compute_statistic` calls in `StateAttributeLimitsHelper` to derive limits based on that subset only and adds a `set_slice` convenience method to create a `SliceSubsetState` for that purpose.
That function is also exposed through `ImageLayerState`.

TBD: ~~`StateAttributeLimitsHelper` should automatically update (re-run statistics) on changes to `subset_state`; currently have to trigger this via setting `percentile` or other attributes (see tests)~~.
Adding this to Callbacks looks a bit hackish, but update seems to work so far.

Perhaps should test multi-dimensional case as well and explore other types of subsets (`SliceSubsetState` has special handling in `compute_statistic`).

### DIFF
--- a/glue/viewers/image/state.py
+++ b/glue/viewers/image/state.py
@@ -587,6 +587,19 @@ class ImageLayerState(BaseImageLayerState, StretchStateMixin):
         """
         self.attribute_lim_helper.flip_limits()
 
+    def set_slice(self, slices):
+        """
+        Select a subset slice for determining image levels.
+
+        Parameters
+        ----------
+        slices : iterable of :class:`slice` or `None`
+            An iterable containing :class:`slice` objects that can instantiate
+            a :class:`~glue.core.subset.SliceSubsetState` and has to be consistent
+            with the shape of `self.data`; `None` to unslice.
+        """
+        self.attribute_lim_helper.set_slice(slices)
+
     def reset_contrast_bias(self):
         with delay_callback(self, 'contrast', 'bias'):
             self.contrast = 1
--- a/glue/viewers/image/state.py
+++ b/glue/viewers/image/state.py
@@ -9,7 +9,7 @@ from glue.viewers.matplotlib.state import (MatplotlibDataViewerState,
                                            DeferredDrawSelectionCallbackProperty as DDSCProperty)
 from glue.core.state_objects import StateAttributeLimitsHelper
 from glue.utils import defer_draw, view_shape
-from echo import callback_property, delay_callback
+from echo import delay_callback
 from glue.core.data_combo_helper import ManualDataComboHelper, ComponentIDComboHelper
 from glue.core.exceptions import IncompatibleDataException
 from glue.viewers.common.stretch_state_mixin import StretchStateMixin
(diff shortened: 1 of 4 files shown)


glue/viewers/image/state.py
```diff
@@ -572,6 +572,19 @@ def flip_limits(self):
         """
         self.attribute_lim_helper.flip_limits()
 
+    def set_slice(self, slices):
```
astrofrog, line 575: I think this should not be a method to set the slices but instead we should have a callback property on image layer state which is whether to apply only to current slice or use whole data for limits, and if the former then it would use the slices from the viewer state. We should then listen to the viewer state slices and update the attribute limits helper.

astrofrog (dismissed):
I'm on board with having this in glue-core but a comment below about API

glue/viewers/image/state.py
dhomeier: @astrofrog that property setting would then be passed as parameter to `attribute_lim_helper` (similar e.g. to `percentile`), and then create an updated `StateAttributeLimitsHelper` as needed? Or should the `ImageLayerState` property still use `self.attribute_lim_helper.set_slice` similar to the current method?

astrofrog: I don't think it matters too much since ``StateAttributeLimitsHelper`` is private API, so I would go for what ends up being cleanest code-wise. However, since we don't always want ``StateAttributeLimitsHelper`` to listen to changes in the ``slice`` callback property (for example by default we don't), maybe keeping the ``self.attribute_lim_helper.set_slice`` approach would be easiest? (and fewest changes compared to what you have now)

dhomeier: That's just the second question coming up now – just that switch would be added as callback (DDsomethingProperty), or `slices` as well? Because it seems once it is set to use current `slices`, it would have to act on any change of that, right?

astrofrog: @dhomeier - we'd have to add a new callback property which would be the boolean switch. When that becomes ``True``, we could call ``add_callback`` which would watch for changes to ``slices``, and when the boolean switch becomes ``False`` it would call ``remove_callback``. Does that make sense?

dhomeier: @astrofrog more or less what I figured out, too – just turned out that adding this to `ImageLayerState` did not seem to go anywhere, as `slices` is an `ImageViewerState` property. I could access it through `viewer_state`, but could not work out how to add that as a callback.
So setting the whole thing up in `ImageViewerState` as in 2e08899b works as it should, but now I have the problem of making sure I am really referencing the right level in `layers` (there's been only one in the examples so far).
Taking your point made in the Slack that this _should_ belong into `ImageLayerState`; I see if I added a new property there updated from `slices` (or directly on `numpy_slice_aggregation_transpose`, which is already used here), that could be set up as a callback – would that work?

dhomeier: @astrofrog experimenting now with defining this via `@callback_property` decorator, but this, after some struggle to delay the initialisation from `numpy_slice_aggregation_transpose` now fails in glue-jupyter on
`TypeError: Object of type slice is not JSON serializable` 🤯
Should I go back the pedestrian way to construct the slice objects outside the callback_property, or is this whole approach doomed to fail?

dhomeier: Pushing this mess for reference; this does not raise at least, but also does not update with changes to `self.viewer_state.slices`. 🤷‍♂️

astrofrog: @dhomeier - regarding listening to changes in slices, have you tried doing the following from the layer state:

```
self.viewer_state.add_callback('slices', self._method_on_layer_state)
```

Does this not work?

dhomeier: Whether in `__init__` or `_set_global_limits`, this gives a
```python
AttributeError: 'BqplotImageLayerState' object has no attribute '_method_on_layer_state'
```
(`self.viewer_state` does not have one either) – does this need to be defined yet, and what exactly should the function do?

astrofrog: Just to be clear, ``_method_on_layer_state`` was a placeholder, I just mean you can use ``self.viewer_state.add_callback(...)``

dhomeier: Excellent. Don't even need an extra `callback_property` then!

dhomeier:
@astrofrog is there anything else needed here, if the 'Custom' handling is acceptable? There are no tests yet, but I did not find any existing tests directly on `attribute_lim_helper`, are the limits tested anywhere indirectly?

astrofrog:
@dhomeier - I'll put this at the top of my list to look at for tomorrow morning.

kecnry:
I'm planning to test this new implementation within jdaviz in the next week or so and will report back here

kecnry:
This seems to work well for what we need in jdaviz (see https://github.com/spacetelescope/jdaviz/pull/2814), thanks!
I'm slightly worried the state name `global_limits` could cause confusion that it somehow is related to x/y limits, rather than the vmin/vmax used for stretch.  In that jdaviz PR, I refer to this switch as `stretch_global` in the intermediate layer between the user and glue-state, although I'm not sure that would be general enough for the internal glue name.

dhomeier:
Yes, the naming is definitely still up to suggestions. I did not use anything related to `stretch` as that is not used anywhere within the relevant glue code – it is basically controlling the behaviour of an `AttributeLimitsHelper`, thus my choice.
But it is also a property of a `StretchStateMixin` subclass – maybe `global_stretch_limits` would make most sense, if it wasn't so long. ;-)

astrofrog:
@dhomeier - this looks good! I have a couple of small suggestions but I need to play around with this tomorrow first to see if those make sense.

dhomeier:
Looking at the description of `StretchStateMixin`, `stretch_global` actually seems appropriate enough.

dhomeier:
OK, just rebased to have the new tests run.

astrofrog:
I've pushed an additional commit adding a test for the image viewer. I've also removed subset_state as a 'modifier' for now as it isn't really needed - this is for if we wanted to link the subset state to one that exists as a callback properly on a state class, which we don't really use. It's also not clear if modifying the subset state in-place would result in the correct events being fired. So for now, as it isn't needed, I decided to remove/simplify this.

@dhomeier - are you happy with these changes? If so, I think we should be good to go assuming the CI passes.

dhomeier:
Thanks for the updates and new tests. When I initially worked on this, I thought the option to pass any subset_state (not just a slices one) accepted by `compute_statistic` through, but obviously there is no immediate application for that.
Checked functionality with Cubeviz again, all looking good!

astrofrog:
Yes I agree it could be useful to allow arbitrary subsets - but it will be a bit more work as we need to figure out how to watch for subset updates.

astrofrog (approved):

PR #2478, opened 2024-03-19
Add the ability to specify a random subset size for computing histogram

astrofrog:
This is to provide a long-term solution for issues such as https://github.com/spacetelescope/jdaviz/pull/2735 where a histogram doesn't have to be exact.

### DIFF
--- a/glue/core/tests/test_data.py
+++ b/glue/core/tests/test_data.py
@@ -1173,15 +1173,15 @@ def test_compute_histogram_random_subset_dask():
                 y=da.ones(1024 ** 2).rechunk((1024,)))
 
     result = data.compute_histogram([data.id['x']], range=[[-0.5, 10.5]], bins=[5], random_subset=10_000_000)
-    assert_allclose(result, [178535., 230694., 229997., 231215., 178135.], atol=10_000)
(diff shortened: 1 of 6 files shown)


glue/core/tests/test_data.py
```diff
@@ -1132,3 +1132,56 @@ def test_compute_histogram_dask_mixed():
...
     assert_allclose(result, [[0, 1, 0], [0, 2, 2]])
+
+
+def test_compute_histogram_random_subset():
+
+    # Make sure that compute_histogram works with the random_subset option
+
+    with NumpyRNGContext(12345):
+
+        data = Data(x=np.random.uniform(0, 10, 1024**2), y=np.ones(1024 ** 2))
```
pllim, line 1143: I don't see a test for 2D image like how Jdaviz would store it. We have a 2D array under one component. Here, you store each dimension as a separate component. I don't think it is equivalent?

Since we do not use histogram viewer directly in Jdaviz, I am trying to emulate this calculation directly over at https://github.com/spacetelescope/jdaviz/pull/2763 but without success. There must be some subtlety I am missing. Are you able to have a look over there and tell me if I went down the wrong rabbit hole?

Cami has a 8GB test image if you need one, if she is willing to share with you.

pllim (approved):

PR #2479, opened 2024-03-20
Fix the logic for region display when exchanging x/y world coordinates

jfoster17:
# Pull Request Template

## Description

This is a fix and test for https://github.com/glue-viz/glue-qt/issues/16, where regions defined in world coordinates were not rotating properly when the x and y coordinates are exchanged.


jfoster17:
This is ready for review

astrofrog (approved):
This looks good to me!

PR #2480, opened 2024-03-28
Implement support for unit conversion in attribute limits in image viewer

astrofrog:
Work in progress


astrofrog:
The main thing to do at this point is to make sure that if it's just the units that change, we convert the units (including in 'Custom' mode)

astrofrog:
This mostly works but we now need to make sure that we use the native units when actually passing the limits to e.g. Matplotlib or bqplot.

astrofrog:
Think this is ready to try out. To use, set ``.attribute_limits_unit`` in the layer state.

PR #2481, opened 2024-03-28
Adds custom list of plugins to be passed on load

CyclingNinja:
# Adds customisable list of plugins on load

Adds an argument to load_plugins in main that allows a custom list to be passed.
On empty, defaults to required plugins which have been added to `_plugin_helpers`


glue/main.py
```diff
@@ -45,17 +49,26 @@ def load_plugins(splash=None, require_qt_plugins=False):
     config = PluginConfig.load()
 
     n_plugins = len(list(iter_plugin_entry_points()))
```
astrofrog, line 51: I think it might be best to actually define ``n_plugins`` as ``len(plugins_to_use)`` once ``plugins_to_use`` has been defined.

glue/main.py
```diff
@@ -45,17 +49,26 @@ def load_plugins(splash=None, require_qt_plugins=False):
...
+    if plugins_to_use is None:
+        if require_qt_plugins:
+            plugins_to_use = [*REQUIRED_PLUGINS, *REQUIRED_PLUGINS_QT]
+        else:
+            plugins_to_use = REQUIRED_PLUGINS
```
astrofrog, line 56: I don't think ``plugins_to_use`` should be based on the ``REQUIRED_*`` variables - these are just used to raise exceptions below if they are not present. I think we actually need two different variables inside this function - ``plugins_to_load``(perhaps a better name than ``plugins_to_use``) and ``plugins_to_require``. I think the logic should be:

* If ``plugins_to_load`` is not defined by the user, then ``plugins_to_load`` defaults to all entry points (``iter_plugin_entry_points``) and ``plugins_to_require`` defaults to either ``REQUIRED_PLUGINS`` or the combination of ``REQUIRED_PLUGINS`` and ``REQUIRED_PLUGINS_QT``
* If ``plugins_to_load`` is defined by the user, then ``plugins_to_require`` should just be set to ``plugins_to_load`` because if the user is explicitly asking for them we should error if they can't be loaded

glue/main.py
```diff
@@ -90,7 +103,7 @@ def load_plugins(splash=None, require_qt_plugins=False):
             _loaded_plugins.add(item.module)
```
astrofrog, line 103: I can't easily comment on lines above but based on my previous comment,

```
     if item.module in REQUIRED_PLUGINS:
                raise
            elif item.module in REQUIRED_PLUGINS_QT and require_qt_plugins:
                raise
```

should then become:

```
     if item.module in plugins_to_require:
                raise
```

astrofrog (changes requested):
This is looking good so far, but think we need to tweak the logic related to requiring plugins - see below. API-wise I think the keyword argument to ``load_plugins`` should probably just be called ``names`` so that it would look like:

```
load_plugins(names=['glue.core.something'])
```

rather than

```
load_plugins(plugins_to_load=['glue.core.something'])
```

which is a bit repetitive. Internally we could still use ``plugins_to_load`` though.

PR #2496, opened 2024-05-23
Add list plugins method and tests

CyclingNinja:


astrofrog:
@CyclingNinja - could you try rebasing on upstream/main to get rid of the merge commit?

glue/tests/test_main.py
```diff
@@ -18,3 +32,12 @@ def test_no_duplicate_loading(capsys):
...
             assert 'failed' in acall[0][0]
+
+
+def test_list_plugins():
+    """
+    Regression test for retrieving the list of currently loaded plugins
+    """
+    load_plugins(require_qt_plugins=False)
+    plugins = list_plugins()
+    assert isinstance(plugins, list)
```
astrofrog, line 43: Maybe you can check that certain plugins are in the list

glue/main.py
```diff
@@ -115,3 +114,10 @@ def load_plugins(splash=None, require_qt_plugins=False, plugins_to_load=None):
     # that were previously read.
     from glue._settings_helpers import load_settings
     load_settings()
+
+
+def list_plugins():
```
astrofrog, line 119: Thinking about this, I think we might want to actually have ``list_available_plugins`` and ``list_loaded_plugins`` as there are use cases for both. Could you implement both here?

astrofrog (changes requested):

glue/tests/test_main.py
```diff
@@ -18,3 +37,13 @@ def test_no_duplicate_loading(capsys):
...
+    assert len(plugins) == 5
```
astrofrog, line 49: Sorry I didn't get back to you sooner about this - rather than checking the absolute number of plugins, you might want to just check that a specific set of them are inside the returned list (but there could be others present)

astrofrog (changes requested):

dhomeier:
May need another rebase to get the updated hash for visual test.

astrofrog (approved):
Think this is good to go, thank you!

PR #2503, opened 2024-07-15
MNT: replace `np.in1d` (deprecated) with `np.isin`

neutrinoceros:
This removes about 25 warnings from the test suite
Patch performed with
```shell
ruff check --select NPY201 --fix glue
```


dhomeier (approved):
LGTM, thanks!

PR #2511, opened 2024-08-30
DEP: drop runtime dependency on `setuptools`

neutrinoceros:
## Description

This is a follow up to #2365
At the moment this breaks tests because `mpl-scatter-density` has an undeclared runtime dependency on `setuptools`
(addressed at https://github.com/astrofrog/mpl-scatter-density/pull/43)

I also note that the following comment
https://github.com/glue-viz/glue/blob/e9df453717b05a632737fbf7a1cefd760459af3c/glue/main.py#L75-L87

which was introduced in #1487, has been irrelevant since #2365 (at least in part) since it mentions behavior details that were specific to setuptools. Should this comment be removed entierely or just rephrased ?

### DIFF
--- a/glue/main.py
+++ b/glue/main.py
@@ -41,9 +41,7 @@ def load_plugins(splash=None, require_qt_plugins=False, plugins_to_load=None):
     # where ``setup`` is a function that does whatever is needed to set up the
     # plugin, such as add items to various registries.
 
-    import setuptools
-    logger.info("Loading external plugins using "
-                "setuptools=={0}".format(setuptools.__version__))
+    logger.info("Loading external plugins")
 
     from glue._plugin_helpers import iter_plugin_entry_points, PluginConfig
     config = PluginConfig.load()
--- a/glue/main.py
+++ b/glue/main.py
@@ -70,21 +70,9 @@ def load_plugins(splash=None, require_qt_plugins=False, plugins_to_load=None):
             if not config.plugins[item.name]:
                 continue
 
-        # We don't use item.load() because that then checks requirements of all
-        # the imported packages, which can lead to errors like this one that
-        # don't really matter:
-        #
-        # Exception: (pytest 2.6.0 (/Users/tom/miniconda3/envs/py27/lib/python2.7/site-packages),
-        #             Requirement.parse('pytest>=2.8'), set(['astropy']))
-        #
-        # Just to be clear, this kind of error does indicate that there is an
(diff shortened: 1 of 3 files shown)


glue/main.py
```diff
@@ -40,11 +40,6 @@ def load_plugins(splash=None, require_qt_plugins=False, plugins_to_load=None):
     #
     # where ``setup`` is a function that does whatever is needed to set up the
     # plugin, such as add items to various registries.
-
-    import setuptools
-    logger.info("Loading external plugins using "
-                "setuptools=={0}".format(setuptools.__version__))
```
astrofrog, line 46: I'd like to keep 'Loading external plugins' - just don't give the setuptools version

astrofrog (commented):
Just a small comment below. Otherwise, regarding the comment about not using item.load(), are you sure that these kind of errors should not happen anymore? If not, then in addition to removing the comment we should then actually use item.load() below in the try...except?

neutrinoceros:
> are you sure that these kind of errors should not happen anymore?

I don't know if there's a test to exercise this condition, but I tried replacing `module = import_module(item.module)` with `module = item.load()` and found it broke some tests for a different reason: `EntryPoint.load` doesn't *always* return a module, which is actually documented in this function's docstring:
```
        """Load the entry point from its definition. If only a module
        is indicated by the value, return that module. Otherwise,
        return the named object.
        """
```
so it stills seems simpler/safer to use `import_module` here, but I think we should update the comment to reflect the real reason. Should I do it here ?

astrofrog:
@neutrinoceros - sounds good to me!

neutrinoceros:
@astrofrog done. Note that this is still blocked by https://github.com/astrofrog/mpl-scatter-density/pull/43

neutrinoceros:
Actually, `mpl-scatter-density` 0.8 unblocked this (thanks !); undrafting now.

tox.ini
```diff
@@ -35,13 +35,12 @@ deps =
     legacy: pandas==1.2.*
     legacy: echo==0.5.*
     legacy: astropy==4.0.*
-    legacy: setuptools==30.3.*
     legacy: ipython==7.16.*
     legacy: ipykernel==5.3.*
     legacy: dill==0.2.*
     legacy: xlrd==1.2.*
     legacy: h5py==2.10.*
-    legacy: mpl-scatter-density==0.7.*
```
neutrinoceros, line 44: this part is now a natural follow up to #2521

PR #2513, opened 2024-09-11
Correct y limits in ProfileViewerState when min same as max

pllim:
This fixes a display limit problem we see over at Cubeviz when cube is all ones.

```python
import numpy as np
from astropy import units as u
from specutils import Spectrum1D

from jdaviz import Cubeviz

sp = Spectrum1D(flux=np.ones((7, 8, 9)) * u.nJy)

cubeviz_helper = Cubeviz()
cubeviz_helper.load_data(sp, data_label='test_cube')
cubeviz_helper.show()
```

Without this patch:

Screenshot 2024-09-11 155700

With this patch:

Screenshot 2024-09-11 155442

[🐱](https://jira.stsci.edu/browse/JDAT-4767)

### DIFF
--- a/glue/viewers/profile/state.py
+++ b/glue/viewers/profile/state.py
@@ -231,9 +231,8 @@ class ProfileViewerState(MatplotlibDataViewerState):
                     self.y_min = y_min
                     self.y_max = y_max
                 elif np.allclose(y_min, y_max):
-                    dy = max(0.1, 0.1 * y_min)
-                    self.y_min = y_min - dy
-                    self.y_max = y_max + dy
+                    self.y_min = y_min * 0.9
+                    self.y_max = y_max * 1.1
                 else:
                     self.y_min = 0
                     self.y_max = 1
--- a/glue/viewers/profile/state.py
+++ b/glue/viewers/profile/state.py
@@ -230,6 +230,10 @@ class ProfileViewerState(MatplotlibDataViewerState):
                 if y_max > y_min:
                     self.y_min = y_min
                     self.y_max = y_max
+                elif np.allclose(y_min, y_max):
+                    dy = max(0.1, 0.1 * y_min)
+                    self.y_min = y_min - dy
+                    self.y_max = y_max + dy
                 else:
                     self.y_min = 0
                     self.y_max = 1
--- a/glue/viewers/profile/state.py
+++ b/glue/viewers/profile/state.py
@@ -231,8 +231,12 @@ class ProfileViewerState(MatplotlibDataViewerState):
                     self.y_min = y_min
                     self.y_max = y_max
                 elif np.allclose(y_min, y_max):
-                    self.y_min = y_min * 0.9
-                    self.y_max = y_max * 1.1
+                    if np.allclose(y_min, 0):
+                        self.y_min = -0.1
+                        self.y_max = 0.1
+                    else:
+                        self.y_min = y_min * 0.9
+                        self.y_max = y_max * 1.1
                 else:
                     self.y_min = 0
                     self.y_max = 1
--- a/glue/viewers/profile/state.py
+++ b/glue/viewers/profile/state.py
@@ -231,12 +231,12 @@ class ProfileViewerState(MatplotlibDataViewerState):
                     self.y_min = y_min
                     self.y_max = y_max
                 elif np.allclose(y_min, y_max):
-                    if np.allclose(y_min, 0):
-                        self.y_min = -0.1
-                        self.y_max = 0.1
+                    if y_min == 0.0:
+                        dy = np.finfo(y_min).resolution**2
                     else:
-                        self.y_min = y_min * 0.9
-                        self.y_max = y_max * 1.1
+                        dy = abs(0.1 * y_min)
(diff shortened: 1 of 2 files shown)


glue/viewers/profile/state.py
```diff
@@ -230,6 +230,10 @@ def _reset_y_limits(self, *event):
                 if y_max > y_min:
                     self.y_min = y_min
                     self.y_max = y_max
+                elif np.allclose(y_min, y_max):
+                    dy = max(0.1, 0.1 * y_min)
+                    self.y_min = y_min - dy
+                    self.y_max = y_max + dy
```
dhomeier, line 236:
```suggestion
                    self.y_min = y_min * 0.9
                    self.y_max = y_max * 1.1
```
Setting a minimum range of ±0.1 seems quite arbitrary here and would lead to the data values basically appearing as zero if they do not happen to be all 1 but something much smaller.
Could in principle even check for `y_min == y_max` here, since by construction it cannot be larger, and the only other possible case is both still being `np.inf`, which is covered in the `else`.

dhomeier (changes requested):
Sorry for the late reply. A good choice for a default range still is somewhat arbitrary, since I cannot conceive any real data where this case would apply. But I've tested this branch with some other examples, and setting a range of at least a few percents would probably be good to have it visible in the tick marks.
I think it should be possible to extend `test_limits` to cover this case?

glue/viewers/profile/state.py
pllim: I don't understand your objection. My code isn't +/- 0.1 absolute. It is at minimum 0.1 (say, if the data is zeroes) but otherwise 10 percent of the value (which scales). Your suggested change does not guard against data that is all zeroes.

dhomeier: Yes, but if it's all << 0.1, it does not scale. E.g. setting the data in you example to `1e-6 * np.ones((7, 8, 9))` with the current version it is displayed as
max(0.1, 0.1 * y_min)
and having all negative data (–1) as
flux[:] == –1
while with my suggestion it looks like
0.1 * y_min
though for negative values the scale is inverted, and the all-zeros case indeed needs to be addressed separately.

pllim:
@dhomeier good point about very small values. I updated the code. How does it look now?

pllim:
Where is `test_limits`? I cannot find it.

dhomeier:
I had just added a slightly different version to account for negative values as well, which is probably now hidden in the _resolved_.

dhomeier:
https://github.com/glue-viz/glue/blob/7481caf980192df0a6ea7fc6ae21ce37f5c492b3/glue/viewers/profile/tests/test_state.py#L130

pllim:
@dhomeier , can u please re-suggest the neg fix so commit will credit you properly? Also feel free to push directly to this branch if that is easier. Thanks!

glue/viewers/profile/state.py
```diff
@@ -230,6 +230,13 @@ def _reset_y_limits(self, *event):
                 if y_max > y_min:
                     self.y_min = y_min
                     self.y_max = y_max
+                elif np.allclose(y_min, y_max):
+                    if np.allclose(y_min, 0):
+                        self.y_min = -0.1
+                        self.y_max = 0.1
+                    else:
+                        self.y_min = y_min * 0.9
+                        self.y_max = y_max * 1.1
```
dhomeier, line 239:
```suggestion
                elif np.allclose(y_min, y_max):
                    if y_min == 0.0:
                        dy = np.finfo(y_min).resolution**2
                    else:
                        dy = abs(0.1 * y_min)
                    self.y_min = y_min - dy
                    self.y_max = y_max + dy
```
still somewhat arbitrary choice for "very small" limits, but I'd prefer something that still shows it is not all 1e-12 or similar.

pllim:
I don't think the Windows job failures are related?

dhomeier:
Suggestion is easier for now, but if you don't find time to write a test, let me know, and I can probably look into that later.

dhomeier:
No, the image loader failures I have seen elsewhere.

pllim:
Added a couple of tests. Please re-review. Thanks!

pllim:
Hmm I enabled push by maintainer, so weird it won't let you push. Did you work off my branch? If you rebase, then that might mess up the history, and not let you push without a force.

dhomeier:
No, I had reset to e9df4537, so should be on the same history, and it is not failing on that afaict:
>To github.com:pllim/glue.git
 ! [remote rejected]   pllim-fix-yminmax-same -> pllim-fix-yminmax-same (permission denied)

pllim:
PR to my branch works too. Pls ping me there. Thx

dhomeier:
pllim/glue#1

pllim:
@dhomeier , I merged your PR to my PR, so your changes is in this PR now. Thanks!

dhomeier (approved):

dhomeier:
Thanks, LGTM now!

pllim:
Thanks for your help!

astrofrog (approved):
Thanks!

PR #2514, opened 2024-09-17
Drop support for Python 3.8 and 3.9

CyclingNinja:
Removes old python versions from builds and min required python

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,6 +11,8 @@ classifiers =
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
+    Programming Language :: Python :: 3.11
+    Programming Language :: Python :: 3.12
     Topic :: Scientific/Engineering :: Visualization
     License :: OSI Approved :: BSD License
 description = Core library for the glue multidimensional data visualization project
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -11,6 +11,8 @@ classifiers =
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
```
astrofrog, line 12: Could you remove 3.8 and 3.9 here?

astrofrog (changes requested):
Small comment below - also could you check whether the documentation needs updating in the installation instructions? (regarding the minimum version of Python)

astrofrog (approved):
Thanks! I reworded the title just as this gets used for the user-facing change notes

setup.cfg
dhomeier: These have been forgotten

dhomeier:
A bit puzzled why `py310-test-dev` is still running here, but not for `glue-qt` – is `astropy>=0.0.dev0` resolved differently from `git+https://github.com/astropy/astropy` (weekly wheels or something)?
Either way this may still start to fail here as well once the former are updated to astropy HEAD.
Also does the LTS test need no replacement?

dhomeier:
> Either way this may still start to fail here as well once the former are updated to astropy HEAD.

@astrofrog, @CyclingNinja, `astropy>=0.0.dev0` is now getting `astropy==6.1.3` installed on `py310-test-dev`, so one dev test should probably be pushed to py312.

PR #2515, opened 2024-09-17
Avoid double messaging on subset creation

astrofrog:
Avoid broadcasting a create subset message with an empty subset state followed by an update message, instead just broadcast a create subset message with the correct subset state

### DIFF
--- a/glue/core/tests/test_edit_subset_mode.py
+++ b/glue/core/tests/test_edit_subset_mode.py
@@ -133,6 +133,45 @@ def test_no_double_messaging():
     assert message.subset.subset_state.right == 3
 
 
+def test_broadcast_to_collection():
+    """A data collection input works on each data component"""
+
+    handler = MagicMock()
+
+    app = Application()
+
+    class Listener(HubListener):
+        pass
+
+    listener = Listener()
+
+    app.session.hub.subscribe(listener, SubsetCreateMessage, handler=handler)
+    app.session.hub.subscribe(listener, SubsetUpdateMessage, handler=handler)
+
+    data = Data(x=[1, 2, 3, 4, 5])
+    data2 = Data(x=[2, 3, 4, 5, 6])
+
+    app.data_collection.append(data)
+    app.data_collection.append(data2)
+
+    cmd = ApplySubsetState(data_collection=app.data_collection,
+                           subset_state=data.id['x'] >= 3,
+                           override_mode=ReplaceMode)
+
+    app.session.command_stack.do(cmd)
+
+    assert len(data2.subsets) == 1
+    assert data2.subsets[0].label == 'Subset 1'
+    # fails with `IncompatibleAttribute` exception
+    # assert data2.get_subset_object(subset_id='Subset 1', cls=NDDataArray)
+
+    assert handler.call_count == 2
+
+    assert data2.subsets[0].subset_state.left is data.id['x']
+    assert data2.subsets[0].subset_state.operator is operator.ge
+    assert data2.subsets[0].subset_state.right == 3
+
+
 def test_message_once_all_subsets_exist():
 
     # Make sure that when we create a new subset via EditSubsetMode, we don't
--- a/glue/core/tests/test_edit_subset_mode.py
+++ b/glue/core/tests/test_edit_subset_mode.py
@@ -1,12 +1,19 @@
 # pylint: disable=I0011,W0613,W0201,W0212,E1101,E1103
 
+import operator
+from unittest.mock import MagicMock
+
 import numpy as np
 
+from ..application_base import Application
+from ..hub import HubListener
+from ..message import SubsetCreateMessage, SubsetUpdateMessage
+from ..command import ApplySubsetState
 from ..data import Component, Data
 from ..data_collection import DataCollection
 from ..edit_subset_mode import (EditSubsetMode, ReplaceMode, OrMode, AndMode,
                                 XorMode, AndNotMode)
-from ..subset import ElementSubsetState
+from ..subset import ElementSubsetState, InequalitySubsetState
 
 
 class TestEditSubsetMode(object):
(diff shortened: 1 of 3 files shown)


astrofrog:
@dhomeier - would you like to review this? I've had confirmation from the jdaviz side that it works for them.

rosteen:
Actually, maybe hold off on merging - I think we might need a small PR in Jdaviz, I'm seeing spectral extraction fail to extract when a spatial subset is created. Hopefully a quick fix on our end.

dhomeier:
But nothing needed in this side, is that right?

rosteen:
> But nothing needed on this side, is that right?

I'm pretty sure that's correct, still investigating the fix on our end.

rosteen:
@astrofrog It seems like at the point we get the `SubsetCreate` message, the subset hasn't finished propagating/being applied to all of the data - line 441 in [spectral extraction](https://github.com/spacetelescope/jdaviz/blob/main/jdaviz/configs/cubeviz/plugins/spectral_extraction/spectral_extraction.py) is causing our automatic spectral extraction upon making a subset to fail:

```
uncertainties = uncert_cube.get_subset_object(
                        subset_id=aperture.selected, cls=StdDevUncertainty
                    )
```
with the error

```
File [~/projects/glue/glue/core/data.py:330](http://localhost:8888/lab/tree/~/projects/glue/glue/core/data.py#line=329), in BaseData.get_subset_object(self, subset_id, cls, **kwargs)
    326         raise ValueError('Specify the object class to use with cls= - supported '
    327                          'classes are:' + format_choices(data_translator.supported_classes))
    329 if len(self.subsets) == 0:
--> 330     raise ValueError("Dataset does not contain any subsets")
    331 elif subset_id is None:
    332     if len(self.subsets) == 1:

ValueError: Dataset does not contain any subsets
```

It seems like this is a timing conflict, would it be possible to send the `SubsetCreate` message after the subset has propagated to all of the relevant data? Or do you have a workaround you could recommend on our side?

dhomeier:
The setter is already calling `self._broadcast()` afaics; any ideas if adding something in `BaseData.add_subset` could help?

glue/core/tests/test_edit_subset_mode.py
```diff
@@ -87,3 +94,40 @@ def test_combines_make_copy(self):
...
+
+    app.session.command_stack.do(cmd)
+
+    assert handler.call_count == 1
+    message = handler.call_args[0][0]
+    assert isinstance(message, SubsetCreateMessage)
+    assert isinstance(message.subset.subset_state, InequalitySubsetState)
+    assert message.subset.subset_state.left is data.id['x']
+    assert message.subset.subset_state.operator is operator.ge
+    assert message.subset.subset_state.right == 3
```
dhomeier, line 133:
```suggestion
    assert message.subset.subset_state.right == 3


def test_broadcast_to_collection():
    """A data collection input works on each data component"""

    handler = MagicMock()

    app = Application()

    class Listener(HubListener):
        pass

    listener = Listener()

    app.session.hub.subscribe(listener, SubsetCreateMessage, handler=handler)
    app.session.hub.subscribe(listener, SubsetUpdateMessage, handler=handler)

    data = Data(x=[1, 2, 3, 4, 5])
    data2 = Data(x=[2, 3, 4, 5, 6])

    app.data_collection.append(data)
    app.data_collection.append(data2)

    cmd = ApplySubsetState(data_collection=app.data_collection,
                           subset_state=data.id['x'] >= 3,
                           override_mode=ReplaceMode)

    app.session.command_stack.do(cmd)

    assert len(data2.subsets) == 1
    assert data2.subsets[0].label == 'Subset 1'
    # fails with `IncompatibleAttribute` exception
    # assert data2.get_subset_object(subset_id='Subset 1', cls=NDDataArray)

    assert handler.call_count == 2

    assert data2.subsets[0].subset_state.left is data.id['x']
    assert data2.subsets[0].subset_state.operator is operator.ge
    assert data2.subsets[0].subset_state.right == 3
```
basic test to check that subset is broadcast to a second component in the dataset.

astrofrog:
@rosteen - I have now added some changes to make sure that we delay any subset creation or deletion messages until the subset has been defined or removed from all datasets. Would you be able to take this for a spin?

@dhomeier - do you think we still need the extra test you suggested above?

dhomeier:
I've tested this with the Cubeviz example, which seems to work – uncertainty subsets are updated on main; automated spectral extraction still needs spacetelescope/jdaviz#3238 (`subscribe(SubsetCreateMessage)`, I assume) to work.

> @dhomeier - do you think we still need the extra test you suggested above?

Not sure; functionality within datasets seemed a bit undertested in general, but since your test caught the actual issue, that's obviously the more important one.

rosteen:
> > @rosteen - I have now added some changes to make sure that we delay any subset creation or deletion messages until the subset has been defined or removed from all datasets. Would you be able to take this for a spin?
> 
> Great, I'll try to test it today before I'm off for vacation.

It looks like this works now, extracted spectra are getting created automatically as expected with https://github.com/spacetelescope/jdaviz/pull/3238.

astrofrog:
@dhomeier sounds good, could you push your test as a commit to this PR?

dhomeier:
Ah, in the commit view it looked like `test_broadcast_to_collection` was already included in 36b74e2 – can simply commit from the suggestion.

glue/core/tests/test_edit_subset_mode.py
```diff
@@ -87,3 +94,81 @@ def test_combines_make_copy(self):
...
+    data1 = Data(x=[1, 2, 3, 4, 5])
+    app.data_collection.append(data1)
+
+    data2 = Data(x=[1, 2, 3, 4, 5])
+    app.data_collection.append(data2)
+
+    count = [0]
+
+    def handler(msg):
+        assert len(data1.subsets) == len(data2.subsets)
```
dhomeier, line 197:
```suggestion
        assert len(data1.subsets) == len(data2.subsets)
        assert data2.subsets[0].subset_state.left is data.id['x']
        assert data2.subsets[0].subset_state.operator is operator.ge
        assert data2.subsets[0].subset_state.right == 3
```
Looking again at the suggested `test_broadcast_to_collection` above, I am wondering if the extra checks could simply be integrated here. Or rather
`assert data2.subsets[0].subset_state.foo == assert data1.subsets[0].subset_state.foo`?

astrofrog: The subset state objects might actually be identical and could check with 'is'? (So not bothering to check attributes of the subset states since the subset states are the same)

dhomeier: That fails with
```python
>       assert data2.subsets is data1.subsets
E       assert (Subset: Subset 1 (data: ),) is (Subset: Subset 1 (data: ),)
E        +  where (Subset: Subset 1 (data: ),) = Data (label: ).subsets
E        +  and   (Subset: Subset 1 (data: ),) = Data (label: ).subsets
```
(as well as `data2.subsets[0] is data1.subsets[0]`, so apparently it's a copy) but
`assert data2.subsets[0].subset_state is data1.subsets[0].subset_state`
works.

astrofrog: Yes sorry I was referring to subset_state

dhomeier:
Hmm
```python
        assert len(data1.subsets) == len(data2.subsets)
>       assert data2.subsets[0].subset_state is data1.subsets[0].subset_state
E       IndexError: tuple index out of range
```
any idea why this is passing in a local run, but failing in all CI envs??

astrofrog:
@dhomeier - I think that error makes sense once the subset is removed as there should no longer be any subsets. Maybe you should have an ``if len(data1.subsets) > 0`` before that assert?

glue/core/tests/test_edit_subset_mode.py
```diff
@@ -87,3 +94,121 @@ def test_combines_make_copy(self):
...
+        assert data2.subsets[0].subset_state is data1.subsets[0].subset_state
```
dhomeier, line 198:
```suggestion
        for i in range(len(data1.subsets)):
            assert data2.subsets[i].subset_state is data1.subsets[i].subset_state
```
Ah, I had missed the updates from 6559b18 – makes sense then as this failed in `remove_subset_group`!

dhomeier (approved):
Obviously; that's what you get from not properly syncing your local branch ;-)
I think they are still planning to make some tests in spacetelescope/jdaviz#3238, but this looks good to go now.

Issue #2517, opened 2024-09-23
TST: Python 3.13


pllim:
Would be nice to test against Python 3.13.

tvwenger:
With python 3.13, installation via `conda` fails to place an executable in the path.

```bash
conda create -n glue-env -c conda-forge python
conda activate glue-env
conda install -c conda-forge glueviz
which glue
# no output
```

Following the newly updated `conda` installation instructions, `glueviz` grabs python 3.12.7 and everything works fine.

tvwenger:
With python 3.13, installation via `pip` works, but throws a `DeprecationWarning`:

```bash
conda create -n glue-env -c conda-forge python
conda activate glue-env
pip install glueviz[all,qt]
# WARNING: glueviz 1.4.0 does not provide the extra 'all'
python --version
# Python 3.13.0
glue
```
Warning:
```
/home/twenger/miniconda3/envs/glue-env/bin/glue:8: DeprecationWarning: Parsing dates involving a day of month without a year specified is ambiguious
and fails to parse leap day. The default behavior will change in Python 3.15
to either always raise an exception or to use a different default year (TBD).
To avoid trouble, add a specific year to the input & format.
See https://github.com/python/cpython/issues/70647.
  sys.exit(main())
```

dhomeier:
closing this as completed by #2518 ; conda install issues to be followed up in #2522

PR #2518, opened 2024-09-23
Add CI for Python 3.13 and update dev envs

dhomeier:
## Description

Introducing test jobs for currently 3.13rc2 and pushing the `py310-test-dev` envs, which are now stuck at Astropy 6.1.3, to higher Python versions.

Non-dev job for py313 is failing on missing libhdf5-dev when trying to build h5py from source; don't know how to install the apt for just a specific env.
~~I have _down-versioned_ the windows-dev job here so it runs with astropy 6.1.3~~; upgrading to py312 produced the same data_factory failures 
> E           AssertionError: assert <function astropy_tabular_data at 0x0000020F8CEE2C00> is <function img_data at 0x0000020F8CEE8FE0>

but the also occur with the non-dev Windows job, so it appears to be a Python 3.11+ issue on that platform.
As far as I can tell this is incorrectly identifying the .png file as readable as an `astropy_tabular_data` _"Catalog"_.
@pllim any ideas what could have happened there?

Addresses #2517

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -43,17 +43,19 @@ jobs:
 
         # Test some configurations on Windows
         - windows: py310-test
-        - windows: py311-test-all
+        - windows: py312-test-all
 
         # Test against latest developer versions of some packages
-        - linux: py310-test-dev-all
-        - linux: py311-test-dev
-        - linux: py312-test-dev-all
+        - linux: py311-test-dev-all
+        - linux: py312-test-dev
+        - linux: py313-test-dev-all
 
-        - macos: py311-test-dev-all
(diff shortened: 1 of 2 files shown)


pllim:
Looks like it tests some sort of third-party PNG loader, so I am not sure. What is this actually testing? That `df.find_factory(fname)` must be the same object as `df.img_data`? Perhaps somehow a copy is made on Windows, breaking that assumption?

pllim:
Did this test ever pass before?

dhomeier:
It did, and does for Python <=3.10. Turned out to be a locale/encoding issue hopefully solved in #2519.

astrofrog (approved):

PR #2519, opened 2024-09-24
Fix an issue that caused the astropy table reader on Windows to behave differently to other platforms

astrofrog:
On Windows, the default encoding/locale seems to be ``cp1252`` which will read random binary files without complaining, which is not ideal for automatic format recognition.

cc @dhomeier

### DIFF
--- a/glue/core/data_factories/astropy_table.py
+++ b/glue/core/data_factories/astropy_table.py
@@ -54,7 +54,7 @@ def astropy_table_read(*args, **kwargs):
             pass
 
     # If the above didn't work, attempt to read with no specified format
-    return Table.read(*args, encoding=encoding, **kwargs)
+    return Table.read(*args, **kwargs)
 
 
 @data_factory(label="Catalog (astropy.table parser)",
--- a/glue/core/data_factories/astropy_table.py
+++ b/glue/core/data_factories/astropy_table.py
@@ -29,6 +29,12 @@ def is_readable_by_astropy(filename, **kwargs):
 
 def astropy_table_read(*args, **kwargs):
 
+    # We need to specify the encoding because otherwise on Windows the default
+    # encoding might be cp1252 which seems to be able to read in most binary
+    # files, which is an issue since it will start recognizing e.g. PNGs as
+    # valid tables.
+    encoding = 'utf-8'
+
     from astropy.table import Table
 
     # In Python 3, as of Astropy 0.4, if the format is not specified, the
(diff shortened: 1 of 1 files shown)


glue/core/data_factories/astropy_table.py
```diff
@@ -40,15 +46,15 @@ def astropy_table_read(*args, **kwargs):
...
+            t = Table.read(*args, format='ascii', encoding=encoding, **kwargs)
             # Double-check for certain FITS files that may be read in as ASCII in Python 3.11
             if not (len(t) == 1 and [c.value[0] for c in t.itercols()][:3] == ['SIMPLE', '=', 'T']):
                 return t
         except Exception:
             pass
 
     # If the above didn't work, attempt to read with no specified format
-    return Table.read(*args, **kwargs)
+    return Table.read(*args, encoding=encoding, **kwargs)
```
dhomeier, line 57:
```suggestion
    return Table.read(*args, **kwargs)
```
Looks like this is failing on trying VOTable format; maybe it's enough to pass encoding explicitly only for 'ascii' as that is tried first anyway?

dhomeier (changes requested):
May give it a try with `encoding` only set in the first pass.

dhomeier:
That seems to resolve the test failures; wondering if there still might be text formats that we are not testing here?
Seeing the same dev test installation failures over at Astropy now; must be somewhere with anaconda, hopefully only temporary.

dhomeier:
Finally got the dev jobs through, so this seems to work at least for all tests.
I've been wondering if pandas text input might run into similar problems. On macOS I get for the test data
```python
>>> with make_file(data, '.png') as fname:
...     for df in data_factory:
...         print(df.label, df.priority, df.identifier(fname))
...     
FITS file 100 False
HDF5 file 100 False
Numpy save file 100 False
ASCII Table 1 False
FITS table 1 False
VO table 1 False
AASTeX Table 0 False
Auto 0 True
CDS Catalog 0 False
Catalog (astropy.table parser) 0 False
DAOphot Catalog 0 False
Excel 0 False
IPAC Catalog 0 False
Image 0 True
LaTeX Table 0 False
Pandas Table 0 False
SExtractor Catalog 0 False
CASA PPV Cube -1000 False
```
My interpretation of that is even if Pandas or LaTeX identified as `True` on Windows, they would not get into the way because Image is tried prior to them.
Wondering if
`with locale.setlocale(locale.LC_ALL, locale=(None, 'utf-8'))`
could be an alternative that would not let the `Table.read()` without `format` fail.

pllim:
CI is green, let's merge!

dhomeier:
Yes, I guess if issues with other text formats should pop up, we can always fix them later.
Otherwise this will hopefully get us through until [3.15](https://peps.python.org/pep-0686/#enable-utf-8-mode-by-default) ;-)

Issue #2525, opened 2024-12-12
Add data to data collection without creating links


javerbukh:
**Is your feature request related to a problem? Please describe it:**

Is there a way to add data to `data_collection` without it also creating links? This line in jdaviz app.py
https://github.com/spacetelescope/jdaviz/blob/cf9664f64db15311c837ce10845500f558f662bc/jdaviz/app.py#L1487
 creates 6 links per cube added and we would prefer to handle that link creation ourselves if possible. I know `data_collection.delay_link_manager_update()` is used for batch loading but I was not able to get that working here.

astrofrog:
@javerbukh - could you try out https://github.com/glue-viz/glue/pull/2526?

PR #2526, opened 2025-01-06
Provide a way to avoid creating world<->pixel links in datasets

astrofrog:
This requires setting:

```python
from glue.config import settings
settings.AUTO_COMPUTE_COORDS_LINKS = False
```

The reason this is done via settings instead of via a kwarg to ``Data`` is because it is not always possible to pass in kwargs when e.g. loading datasets, so having a global setting makes more sense.

Fixes https://github.com/glue-viz/glue/issues/2525

### DIFF
--- a/glue/core/data.py
+++ b/glue/core/data.py
@@ -1179,7 +1179,8 @@ class Data(BaseCartesianData):
                     label = axis_label(self.coords, i)
                     cid = self.add_component(comp, label)
                     self._world_component_ids.append(cid)
-                self._set_up_coordinate_component_links(ndim)
+                if settings.AUTO_COMPUTE_COORDS_LINKS:
+                    self._set_up_coordinate_component_links(ndim)
 
     def _set_up_coordinate_component_links(self, ndim):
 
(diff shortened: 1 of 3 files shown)


javerbukh:
This works for Jdaviz, thank you!

dhomeier:
So it's also working in datasets context as it should?

glue/core/data.py
```diff
@@ -1179,7 +1179,8 @@ def delay_callbacks():
                     label = axis_label(self.coords, i)
                     cid = self.add_component(comp, label)
                     self._world_component_ids.append(cid)
-                self._set_up_coordinate_component_links(ndim)
+                if settings.AUTO_COMPUTE_COORDS_LINKS:
+                    self._set_up_coordinate_component_links(ndim)
```
dhomeier, line 1183: I assume the _creation_ of the components above is still required.

dhomeier, line 1183: I assume the _creation_ of the components above is still required.

astrofrog: We could remove that but it would be a bigger change - best to try this in a separate PR

PR #2530, opened 2025-01-28
Implement hub callback priorities

rosteen:
This PR allows users to set a priority for the hub callback when subscribing to a hub message, e.g.:

```
self.hub.subscribe(self, SubsetCreateMessage,
                           handler=lambda _: self._on_subset_created(),
                           priority=100)
```

I implemented this with a default priority of 10, where lower integer values are higher priority. I confirmed that this PR (and the code snippet above in Jdaviz) fixes the bug with the Layer not being populated before our data menu processes the `SubsetCreateMessage`.

@astrofrog @dhomeier There are two failing tests, I haven't figured out yet why those are now failing, so if you have thoughts on that or the implementation in general let me know.

### DIFF
--- a/glue/core/hub.py
+++ b/glue/core/hub.py
@@ -91,8 +91,8 @@ class Hub(object):
 
         :param priority:
            An optional integer to set the priority of the handler. Handlers
-           are sorted such that lower integer priority handlers get called
-           first when broadcasting a message.
+           are sorted such that higher priority handlers get called first
+           when broadcasting a message.
 
 
         Raises:
--- a/glue/core/hub.py
+++ b/glue/core/hub.py
@@ -3,9 +3,10 @@ from contextlib import contextmanager
 from weakref import WeakKeyDictionary
 from inspect import getmro
 from collections import Counter
+import numpy as np
 
 from glue.core.exceptions import InvalidSubscriber, InvalidMessage
-from glue.core.message import Message
+from glue.core.message import Message, SubsetCreateMessage
 from glue.core.hub_callback_container import HubCallbackContainer
 
 __all__ = ['Hub', 'HubListener']
@@ -68,7 +69,8 @@ class Hub(object):
 
     def subscribe(self, subscriber, message_class,
                   handler=None,
-                  filter=lambda x: True):
+                  filter=lambda x: True,
+                  priority=10):
         """Subscribe an object to a type of message class.
 
         :param subscriber: The subscribing object
@@ -87,6 +89,11 @@ class Hub(object):
            An optional function of the form filter(message). Messages
            are only passed to the subscriber if filter(message) == True.
            The default is to always pass messages.
+
+        :param priority:
+           An optional integer to set the priority of the handler. Handlers
+           are sorted such that lower integer priority handlers get called
+           first when broadcasting a message.
 
 
         Raises:
@@ -113,7 +120,7 @@ class Hub(object):
         if subscriber not in self._subscriptions:
             self._subscriptions[subscriber] = HubCallbackContainer()
 
-        self._subscriptions[subscriber][message_class] = handler, filter
+        self._subscriptions[subscriber][message_class] = handler, filter, priority
 
     def is_subscribed(self, subscriber, message):
         """
@@ -160,9 +167,11 @@ class Hub(object):
         """Yields all (subscriber, handler) pairs that should receive a message
         """
         # self._subscriptions:
-        # subscriber => { message type => (filter, handler)}
+        # subscriber => { message type => (filter, handler, priority)}
 
         # loop over subscribed objects
+        priorities = []
+        prioritized_handlers = []
         for subscriber, subscriptions in list(self._subscriptions.items()):
 
             # subscriptions to message or its superclasses
@@ -175,9 +184,13 @@ class Hub(object):
             # narrow to the most-specific message
             candidate = max(messages, key=_mro_count)
 
-            handler, test = subscriptions[candidate]
+            handler, test, priority = subscriptions[candidate]
             if test(message):
-                yield subscriber, handler
+                priorities.append(priority)
+                prioritized_handlers.append((subscriber, handler))
+
+            for i in np.argsort(priorities):
+                yield prioritized_handlers[i]
 
     @contextmanager
     def ignore_callbacks(self, ignore_type):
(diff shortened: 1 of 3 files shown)


glue/core/hub.py
```diff
@@ -175,9 +184,13 @@ def _find_handlers(self, message):
             # narrow to the most-specific message
             candidate = max(messages, key=_mro_count)
 
-            handler, test = subscriptions[candidate]
+            handler, test, priority = subscriptions[candidate]
             if test(message):
-                yield subscriber, handler
+                priorities.append(priority)
+                prioritized_handlers.append((subscriber, handler))
```
astrofrog, line 190: I'd rather keep the priorities in the same container as the handlers, so a better way might be to do something like ``prioritized_handlers.append((subscriber, handler, priority))`` then do a ``sort(prioritized_handlers, key=x[2], reversed=True)`` below

glue/core/hub.py
```diff
@@ -175,9 +184,13 @@ def _find_handlers(self, message):
...
+
+            for i in np.argsort(priorities):
+                yield prioritized_handlers[i]
```
astrofrog, line 193: This would then be something like:

```
for subscriber, handler, _ in sort(prioritized_handlers, key=x[2], reversed=True):
    yield subscriber, handler
```

astrofrog (changes requested):
Looks promising! Just a couple of comments and then if you could look into the tests that'd be great, but if you don't have time, just let me know and I'll see if I can investigate.

glue/core/hub.py
dhomeier, line 193:
```suggestion
            for subscriber, handler, _ in sorted(prioritized_handlers, key=lambda x: x[2], reverse=True):
                yield subscriber, handler
```
close

rosteen:
> I only see the one failed test on Mock called twice; tried with @astrofrog's suggestion and removing the `priorities.append(priority)`, but there are still two calls (apparently with the correct message though).

Oh good, I was seeing two locally, the second must have been due to something else in my local setup.

glue/core/hub.py
```diff
@@ -175,9 +182,12 @@ def _find_handlers(self, message):
             # narrow to the most-specific message
             candidate = max(messages, key=_mro_count)
 
-            handler, test = subscriptions[candidate]
+            handler, test, priority = subscriptions[candidate]
             if test(message):
-                yield subscriber, handler
+                prioritized_handlers.append((subscriber, handler, priority))
+
+        for subscriber, handler, _ in sorted(prioritized_handlers, key=lambda x: x[2], reverse=True):
```
astrofrog, line 189: Should this loop be moved up one level in indentation?

rosteen:
> Looks promising! Just a couple of comments and then if you could look into the tests that'd be great, but if you don't have time, just let me know and I'll see if I can investigate.

I fixed the failing test, the second loop to yield the handlers shouldn't have been inside the loop to populate the `prioritized_handlers`

glue/core/hub.py
rosteen: I pushed that fix as you were typing! That was the problem.

dhomeier: Need to remember to force-reload GH interface before looking into stuff!

astrofrog:
Ideally it would be good to add at least one test for this - is that something you think you would be able to look into @rosteen or should @dhomeier or I look into it?

rosteen:
> Ideally it would be good to add at least one test for this - is that something you think you would be able to look into @rosteen or should @dhomeier or I look into it?

If @dhomeier has time today that would be awesome, otherwise I can try to figure something out tomorrow.

glue/core/hub.py
```diff
@@ -88,6 +89,11 @@ def subscribe(self, subscriber, message_class,
            are only passed to the subscriber if filter(message) == True.
            The default is to always pass messages.
 
+        :param priority:
+           An optional integer to set the priority of the handler. Handlers
+           are sorted such that lower integer priority handlers get called
+           first when broadcasting a message.
```
dhomeier, line 95:
```suggestion
           are sorted such that higher priority handlers get called first
           when broadcasting a message.
```
Only realised that looking at the test, which is running handler3-handler1-handler2.
Seems more intuitive that way, too.
`sorted` will work the same with non-integers; don't know if we should check input type nonetheless or just silently accept them.

glue/core/tests/test_hub.py
```diff
@@ -165,6 +165,44 @@ def test_invalid_init(self):
         assert exc.value.args[0] == ("Inputs must be HubListener, data, "
                                      "subset, or data collection objects")
 
+    @pytest.mark.parametrize('priorities', (False, True))
+    def test_handler_priorities(self, priorities):
```
dhomeier, line 169:
```suggestion
    def test_handler_priorities(self, priorities):
        """
        Test that handlers are called in order of descending priority, if set,
        in order subscribed to otherwise.
```
Just because it took me a little to figure that order out ;-)

dhomeier (changes requested):
Just some corrections to docstrings, otherwise looks good now, thanks!

glue/core/hub.py
rosteen: Oh, this is the reverse of what my intent was, I missed the `reverse=True` in the code snippet @astrofrog sent with the update `sorted` call. I would rather remove that keyword there so that the current docstring is accurate and update the test, unless you two feel strongly that reverse sorting is the way it should work.

dhomeier: Thought it might have sneaked in there – but as I commented, `"priority"` to me would rather imply that the highest are done first, so maybe if you prefer ascending order, maybe it would be better to rename the keyword?

rosteen: In my head, if one things is "First priority/priority 1", that is higher priority than the thing that is "second priority/priority 2". Let's hear the tiebreaker vote from @astrofrog 🙂

dhomeier: Yes, my training from the German educational and astronomical magnitude systems is leading me that way, too – for numbers ~0 - 6, for the examples coming into play here it's rather the opposite.

astrofrog: In echo and glue-core, higher priority value = higher priority = do it first 😄 (we already use priorities in other places so we should remain consistent)

rosteen: Got it, consistency it is then! I committed @dhomeier's suggestions, I think this is good to go.

dhomeier (approved):

dhomeier:
Excellent, thanks!

PR #2533, opened 2025-03-02
Fix int dtype test

olebole:
During the update of the Debian package for **glue** I observed one failure when the [test runs on a 32-bit machine](https://salsa.debian.org/debian-astro-team/glue/-/jobs/7186302):

```
________________________________ test_dtype_int ________________________________
    def test_dtype_int():
        data = b'# a, b\n1, 1 \n2, 2 \n3, 3'
        with make_file(data, '.csv') as fname:
            d = df.load_data(fname)
>       assert d['a'].dtype == int
E       AssertionError: assert dtype('int64') == int
E        +  where dtype('int64') = array([1, 2, 3], dtype=int64).dtype
glue/core/data_factories/tests/test_data_factories.py:125: AssertionError
```

This patch fixes this by explicitly using `np.int64` instead of `int`.

### DIFF
--- a/glue/core/data_factories/tests/test_data_factories.py
+++ b/glue/core/data_factories/tests/test_data_factories.py
@@ -118,11 +118,11 @@ def test_csv_pandas_factory():
     assert isinstance(d.get_component(cat_comp), CategoricalComponent)
 
 
-def test_dtype_int():
+def test_dtype_int64():
     data = b'# a, b\n1, 1 \n2, 2 \n3, 3'
     with make_file(data, '.csv') as fname:
         d = df.load_data(fname)
-    assert d['a'].dtype == int
+    assert d['a'].dtype == np.int64
 
 
 def test_dtype_float():
--- a/glue/core/data_factories/tests/test_data_factories.py
+++ b/glue/core/data_factories/tests/test_data_factories.py
@@ -122,7 +122,7 @@ def test_dtype_int64():
     data = b'# a, b\n1, 1 \n2, 2 \n3, 3'
     with make_file(data, '.csv') as fname:
         d = df.load_data(fname)
-    assert d['a'].dtype == np.int64
+    assert d['a'].dtype.kind == 'i'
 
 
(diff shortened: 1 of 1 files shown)


dhomeier:
Thanks for finding this! I suspect this has sneaked in with Numpy 2; just wondering if this still passes with Numpy 1.x on 32bit, but we don't have either in our test matrix.

glue/core/data_factories/tests/test_data_factories.py
```diff
@@ -118,11 +118,11 @@ def test_csv_pandas_factory():
     assert isinstance(d.get_component(cat_comp), CategoricalComponent)
 
 
-def test_dtype_int():
+def test_dtype_int64():
     data = b'# a, b\n1, 1 \n2, 2 \n3, 3'
     with make_file(data, '.csv') as fname:
         d = df.load_data(fname)
-    assert d['a'].dtype == int
+    assert d['a'].dtype == np.int64
```
astrofrog, line 125: I think it might be sufficient to check ``assert d['a'].dtype.kind == 'i'`` - as we don't actually care what kind of integer it is

astrofrog (changes requested):

glue/core/data_factories/tests/test_data_factories.py
dhomeier, line 125:
```suggestion
    assert d['a'].dtype.kind == 'i'
```
yes, let's give this a try

dhomeier:
@olebole I don't see how the `dtype.kind` test could not work on 32 bit, but if you have time to verify, that would be great.

glue/core/data_factories/tests/test_data_factories.py
olebole: I can confirm that this change works with both32 and 64 bit. Thanks for the hint.

dhomeier: Thanks for your fix!

PR #2536, opened 2025-04-08
fix: support relative paths for icons

maartenbreddels:
When using PyInstaller, importlib_resources.files("glue") can return a relative path.
If this path is not resolved, it will lead to an error like:

  File "glue/icons/__init__.py", line 35, in icon_path
RuntimeError: Icon does not exist: /Users/maartenbreddels/github/spacetelescope/jdaviz/standalone/dist/jdaviz/_internal/glue_jupyter/bqplot/common/../../icons/glue_polygon.svg


dhomeier (approved):
Looks good to me; I guess there's no easy way to test this from within `glue`?

maartenbreddels:
No, unless we mock it, but with the comment in it, this is unlikely to be broken in the future.

PR #2547, opened 2025-07-09
fix: make the data translator registry prefer exact cls matches over isinstance matches

bmorris3:
# Pull Request Template

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.


glue/config.py
```diff
@@ -511,7 +511,11 @@ def supported_classes(self):
 
     def get_handler_for(self, data_or_class):
         for translator in self:
-            if isinstance(data_or_class, translator.target_cls) or data_or_class is translator.target_cls:
+            if data_or_class is translator.target_cls or type(data_or_class) is translator.target_cls:
```
astrofrog, line 514: In principle should we do the loop twice, once over exact matches then once with isinstance?

Issue #2548, opened 2025-07-09
Regression with Astropy 7.2: WCSLink does not raise on WCS without physical_types


dhomeier:
**Describe the bug**
Failure of `test_wcs_autolink_emptywcs` in all dev jobs, can be tracked down to `wcs_autolink` returning a non-empty list with one link.

**To Reproduce**
Test `glue/plugins/wcs_autolinking/tests/test_wcs_autolinking.py` or manually run `WCSLink(data1, data2)` with the data as setup there.

**Expected behavior**
`WCSLink` should raise in
https://github.com/glue-viz/glue/blob/bf65166f1f8a4367d4b43195a38e4ff653385d6b/glue/plugins/wcs_autolinking/wcs_autolinking.py#L210-L211
but returns a link with `link._physical_types_* = [None]` instead.
Might need to check explicitly for valid `_physical_types_` there in addition to `pixel_cids1`, but perhaps should investigate the responsible change in `astropy.wcs` first.

**Details:**

- Operating System: macOS 11.7.10 or any in -dev job, e.g. in
  [#2547 CI](https://github.com/glue-viz/glue/actions/runs/16174223343/job/45655275147?pr=2547)

dhomeier:
Thanks for identifying the change @CyclingNinja! It looks like `WCSLink` previously has been failing on
https://github.com/glue-viz/glue/blob/bf65166f1f8a4367d4b43195a38e4ff653385d6b/glue/plugins/wcs_autolinking/wcs_autolinking.py#L79-L82
raising a `TypeError: iteration over a 0-d array` on the empty (NAXIS=0) wcs, but that part now passes after the astropy fix changed the `pixel_to_world_values` return to a length-1 list/array.
So the whole check for WCS compatibility is relying on their dimensionality rather than actual physical types; should that be reconsidered?

PR #2551, opened 2025-07-22
Empty WCS autolink handle

CyclingNinja:
The astro PR fixes a broken world -> pixel on sliced 1D WCS 

WCSLink now checks for empty WCS property values on init 
Also adds 2D test to empty calls

### DIFF
--- a/glue/plugins/wcs_autolinking/tests/test_wcs_autolinking.py
+++ b/glue/plugins/wcs_autolinking/tests/test_wcs_autolinking.py
@@ -40,6 +40,22 @@ def test_wcs_autolink_emptywcs():
     links = wcs_autolink(dc)
     assert len(links) == 0
 
+def test_wcs_autolink_2D_emptywcs():
+
+    # No links should be found because the WCS don't actually have well defined
+    # physical types.
+
+    data1 = Data()
+    data1.coords = WCS(naxis=2)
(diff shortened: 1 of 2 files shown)


dhomeier:
fixes #2548

dhomeier (commented):
To follow up on the comments in https://github.com/glue-viz/glue/pull/2534#discussion_r2220499924, I have looked at the test in `test_celestial_with_unknown_axes` for different combinations of unknown physical types in data1 and data2:
1. invalid + valid (current test)
2. invalid + invalid
3. valid + missing
4. missing + missing (comparable to the 1D test, but with additional celestial axes)

I think all combinations _should_ pass (i.e. create only links for the celestial axes), but with the patch so far 2 & 4 are failing in
```python
>       assert len(link) == 4
E       assert 6 == 4
```
so it's not an issue of invalid vs. entirely missing physical type, but with unknowns in both of the datasets. Both could be fixed with the additional
`if physical_type1 is not None:` check.

I'd suggest to parametrise the test e.g. as
```python
@pytest.mark.parametrize('physical_types',
                         [['SPAM', 'FREQ'], ['SPAM', 'EGGS'], ['WAVE', ''], ['', '']])
def test_celestial_with_unknown_axes(physical_types):
```

PR #2553, opened 2025-07-25
Add type handling tests

CyclingNinja:
Addresses the type handling for wcs_autolinking

### DIFF
--- a/glue/plugins/wcs_autolinking/tests/test_wcs_autolinking.py
+++ b/glue/plugins/wcs_autolinking/tests/test_wcs_autolinking.py
@@ -196,6 +196,35 @@ def test_link_editor():
     assert link2.data2.label == 'Data 2'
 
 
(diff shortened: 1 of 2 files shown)


dhomeier:
Hmm, unless I am missing something, this would set
```python
wcs1.wcs.ctype = ['WAVE', 'TIME']

wcs2.wcs.ctype = ['SPAM', 'EGGS']
```
and so on, but it passes even without the fixes to `WCSLink`. Do you think anything is missing from #2552?

dhomeier:
What's 255? Anyway this is yet another case, two pixel dimensions and no celestial dimensions, thus unrelated to what #2552 is handling. I think it's correctly asserting no links are created, but if we add this test, it probably could replace `test_wcs_autolink_2D_emptywcs` completely – it corresponds almost exactly to the `[['', ''], ['', '']]` case here, right?

PR #2554, opened 2025-08-09
Migrate configuration to pyproject.toml

dhomeier:
## Description
Converting setup.cfg to toml and setting up tool.setuptools instead of setup.py, following glue-viz/glue-qt#49.


dhomeier:
pre-commit.ci autofix

dhomeier:
I am very much clueless why the [B905 errors](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/) are popping up all of a sudden now, as it has been in the default rules long before, but I'd rather fix all the Ruff failures in #2546.

dhomeier:
Turns out ZipWithoutExplicitStrict is a Python>=3.10 only thing, but Ruff was not aware of the
`python_requires = >=3.10` in setup.cfg even after the minimum version was increased.
Therefore adding the ignore here after all (I suspect most cases should use `zip(..., strict=True)` really, but that needs further investigation.

dhomeier:
Giving some envs for Python 3.14 (at rc2 now) a try here; probably needs scientific-python-nightly-wheels for compatible h5py and shapely versions, so py314 might be best left to the dev jobs for now.

dhomeier:
As we have just removed pre-3.10 support, I think it makes sense to add some 3.14 envs now, but probably should weed out some more older ones. Currently there are 5 extra CI jobs.

dhomeier:
I've set up Python rc tests as allowed failures now, since at least at the moment there are no Pandas nightly wheels available for 3.14.

PR #2468, opened 2023-12-07
Preserve alpha while compositing with cmap.set_bad

bmorris3:
(in progress, for discussion)

PR #2556, opened 2025-09-03
Public image layer attributes for colormap 'bad' value

bmorris3:
This PR adds public attributes to the image layer artist, state, and `CompositeArray`. The goal is to speficy how colormaps in image layers use the `bad` color and alpha.

This PR is a followup to: 
* #2468 
* #2482

and specifically implements the suggestion in https://github.com/glue-viz/glue/pull/2468#issuecomment-1956890380.

# Demo

```python
import numpy as np
import matplotlib.pyplot as plt
from astropy.io import fits
from glue_jupyter import JupyterApplication as App


# generate two example images with different x/y extents:
fits.HDUList(fits.PrimaryHDU(data=np.random.uniform(size=(20, 20)))).writeto('arr_a.fits', overwrite=True)
fits.HDUList(fits.PrimaryHDU(data=np.random.uniform(size=(10, 10)) + 2)).writeto('arr_b.fits', overwrite=True)

# load two images
app = App()
a = app.load_data('arr_a.fits')
b = app.load_data('arr_b.fits')

# link the x/y coordinates:
app.add_link(a, 'Pixel Axis 0 [y]', b, 'Pixel Axis 0 [y]')
app.add_link(a, 'Pixel Axis 1 [x]', b, 'Pixel Axis 1 [x]')

# create an image viewer with both data entries:
viewer = app.new_data_viewer('image')
viewer.add_data(a)
viewer.add_data(b)

colormaps = [plt.cm.viridis, plt.cm.Grays]

for layer, cmap in zip(viewer.layers, colormaps):
    layer.state.cmap = cmap

    # the default behavior has the out-of-bounds opacity set to `None`, which
    # makes the layer's out-of-bounds area opaque black:
    # layer.state.cmap_bad_alpha = None

    # alternatively, set the out-of-bounds opacity to zero 
    # to see the lower, colorful layer:    
    layer.state.cmap_bad_alpha = 0

    # alternatively, set the out-of-bounds opacity to `1` to 
    # render the `cmap.get_bad()` color (white) in the out-of-bounds areas:
    # layer.state.cmap_bad_alpha = 1
```

using the three options available in comments in the loop:

### `cmap_bad_alpha = None`
This is the default behavior, both before and after this PR. The out-of-bounds area is opaque black:

Screenshot 2025-09-03 at 15 50 13

### `cmap_bad_alpha = 0`
This is the use case that motivated this PR. The out-of-bounds area is completely transparent:
Screenshot 2025-09-03 at 15 50 41

### `cmap_bad_alpha = 1`
One could argue this _should_ be the default behavior, but I'm not asserting that in this PR for backwards compatibility. The out-of-bounds area is opaque white (the default `bad` color/alpha for this colormap):
Screenshot 2025-09-03 at 15 51 27


glue/viewers/image/composite_array.py
```diff
@@ -146,16 +147,12 @@ def __call__(self, bounds=None):
...
                     bad_color = bad_rgba[:3]
-                    bad_alpha = bad_rgba[3:]
-
-                    if self._allow_bad_alpha:
-                        bad_rgba = bad_color + bad_alpha
-                    else:
-                        bad_rgba = bad_color + [layer['alpha']]
-
+                    bad_alpha = bad_rgba[-1] if self.cmap_bad_alpha is None else self.cmap_bad_alpha
+                    bad_rgba = bad_color + (bad_alpha,)
```
astrofrog, line 155: Looks like ``bad_color`` is a [0,0,0] array at this point so adding 1 to it will give [1,1,1] rather than (0,0,0,1)

```suggestion
                    bad_rgba = tuple(bad_color) + (bad_alpha,)
```

astrofrog (changes requested):
Looks good overall, thanks! See suggestion below for a possible fix.

bmorris3:
It's good for me, so long as you don't have better name suggestions.

PR #2559, opened 2025-10-02
Started adding 3D state classes to glue-core

astrofrog:
We should move the 3D state classes, at least the part that is unrelated to vispy. to glue-core so that they can be used by any 3D viewers including ipyvolume, without having to depend on glue-vispy-viewers. This will help us not require glue-vispy-viewers in glue-jupyter and therefore make it easier to use in jupyterlite (for which vispy doesn't work).

### DIFF
--- a/glue/viewers3d/common/layer_state.py
+++ b/glue/viewers3d/common/layer_state.py
@@ -1,0 +1,47 @@
+from echo import CallbackProperty, keep_in_sync
+from glue.core.message import LayerArtistUpdatedMessage
+from glue.viewers.common.state import LayerState
+
+__all__ = ['LayerState3D']
+
+
+class LayerState3D(LayerState):
+    """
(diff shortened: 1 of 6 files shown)


Carifio24:
@astrofrog I've been playing around with using these in the ipyvolume viewers, and I'm wondering whether we also want to move certain pieces of the logic from the Vispy viewers upstream here as well. I'm thinking things like the FRB calculation that we do for the Vispy volume viewer: https://github.com/glue-viz/glue-vispy-viewers/blob/main/glue_vispy_viewers/volume/layer_artist.py#L54 or the computation of the permutation needed for the axes: https://github.com/glue-viz/glue-vispy-viewers/blob/main/glue_vispy_viewers/volume/viewer_state.py#L91.

Pretty much any 3D glue viewer is going to want this sort of functionality, so maybe it's better to put it in one place?

PR #2569, opened 2026-02-09
Rename cmap_mode in image viewer to color_mode for consistency

astrofrog:
This is testing out a new feature added in echo to be able to have aliases for callback properties. For now this just changes the property without changing all the instances of using cmap_mode to check that the alias does indeed work.

The 2D scatter mode and glue-wwt both use color_mode, and I think that name makes more sense since the choice is between fixed color and cmap.

cc @Carifio24

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
     "scipy>=1.1",
     "pandas>=1.2",
     "astropy>=4.0",
-    "echo>=0.6",
+    "echo @ git+https://github.com/glue-viz/echo",
     "fast_histogram>=0.12",
     "ipython>=4.0",
     "dill>=0.2",


Carifio24:
@astrofrog glue-ar is only concerned with the 3D viewers, so it won't be affected by this unless the 3D scatter state inherits from this one.

glue-plotly probably does though, so I could use that for testing this

astrofrog:
@Carifio24 - that'd be great, could you confirm if glue-ar works fine with this?

Carifio24:
@astrofrog I ran the glue-plotly test suite with this PR and dev echo, and it passed fine! So at that level I think this is okay

However, I also tried opening up glue-qt in the same environment. The application opens fine, but the layer options panel doesn't open. The glue terminal gives an error about the C++ combo box being deleted, but the issue ultimately comes from [here](https://github.com/glue-viz/echo/blob/main/echo/qt/connect.py#L418)

I can confirm that if I replace that line with (kinda gross, maybe there's a cleaner check?)
```python
type_prop = getattr(type(instance), prop)
        if not (isinstance(type_prop, SelectionCallbackProperty) or (isinstance(type_prop, CallbackPropertyAlias) and isinstance(type_prop._target_property, SelectionCallbackProperty))):
```

then the panel will open, so we might need to make checks of this sort in the echo connection classes.

astrofrog:
@Carifio24 I think the Qt issue will be fixed in echo 0.12.1 (building right now)

PR #2573, opened 2026-02-12
Update methods to use `world_n_dim` instead of `self._data.ndim`

MatthewPortman:
* This is done to correct for a units mismatch for data with higher dimensionality coordinates than data (e.g. 3D WCS with 2D data). See [Issue 2574](https://github.com/glue-viz/glue/issues/2574)

### DIFF
--- a/glue/core/component.py
+++ b/glue/core/component.py
@@ -236,13 +236,15 @@ class CoordinateComponent(Component):
     @property
     def units(self):
         if self.world:
-            return self._data.coords.world_axis_units[self._data.ndim - 1 - self.axis] or ''
+            world_n_dim = getattr(self._data.coords, 'world_n_dim', self._data.ndim)
+            return self._data.coords.world_axis_units[world_n_dim - 1 - self.axis] or ''
         else:
             return ''
 
     def _calculate(self, view=None):
 
         if self.world:
+            world_n_dim = getattr(self._data.coords, 'world_n_dim', self._data.ndim)
 
             # Calculating the world coordinates can be a bottleneck if we aren't
             # careful, so we need to make sure that if not all dimensions depend
@@ -268,7 +270,7 @@ class CoordinateComponent(Component):
             # convert these straight to world coordinates since the indices
             # of the pixel coordinates are the pixel coordinates themselves.
             if isinstance(view, (tuple, list)) and isinstance(view[0], np.ndarray):
-                axis = self._data.ndim - 1 - self.axis
+                axis = world_n_dim - 1 - self.axis
                 return pixel2world_single_axis(self._data.coords, *view[::-1],
                                                world_axis=axis)
 
@@ -328,7 +330,8 @@ class CoordinateComponent(Component):
             pix_coords = np.meshgrid(*pix_coords, indexing='ij', copy=False)
 
             # Finally we convert these to world coordinates
-            axis = self._data.ndim - 1 - self.axis
+            world_n_dim = getattr(self._data.coords, 'world_n_dim', self._data.ndim)
+            axis = world_n_dim - 1 - self.axis
             world_coords = pixel2world_single_axis(self._data.coords,
                                                    *pix_coords[::-1],
                                                    world_axis=axis)
--- a/glue/core/component.py
+++ b/glue/core/component.py
@@ -229,6 +229,13 @@ class CoordinateComponent(Component):
         self._data = data
         self.axis = axis
 
+        # From issue #2574: initialize world_n_dim and error for data ndim > world ndim.
+        if self.world and self._data:
+            self.world_n_dim = getattr(self._data.coords, 'world_n_dim', self._data.ndim)
+            if self.world_n_dim < self._data.ndim:
+                raise ValueError("World must have at least the same number of "
+                                 "dimensions as data.")
+
     @property
     def data(self):
         return self._calculate()
@@ -236,15 +243,13 @@ class CoordinateComponent(Component):
     @property
     def units(self):
         if self.world:
-            world_n_dim = getattr(self._data.coords, 'world_n_dim', self._data.ndim)
-            return self._data.coords.world_axis_units[world_n_dim - 1 - self.axis] or ''
+            return self._data.coords.world_axis_units[self.world_n_dim - 1 - self.axis] or ''
         else:
             return ''
 
     def _calculate(self, view=None):
 
         if self.world:
-            world_n_dim = getattr(self._data.coords, 'world_n_dim', self._data.ndim)
 
             # Calculating the world coordinates can be a bottleneck if we aren't
             # careful, so we need to make sure that if not all dimensions depend
@@ -260,7 +265,7 @@ class CoordinateComponent(Component):
             # To optimize this, we therefore essentially consider only the
             # dependent dimensions and then broacast the result to the full
             # array size at the very end.
-            axis = world_n_dim - 1 - self.axis
+            axis = self.world_n_dim - 1 - self.axis
 
             # view=None actually adds a dimension which is never what we really
             # mean, at least in glue.
@@ -318,7 +323,7 @@ class CoordinateComponent(Component):
                 else:
                     final_shape.append(self._data.shape[i])
 
-                if world_n_dim == self._data.ndim and i not in dep_coords:
+                if self.world_n_dim == self._data.ndim and i not in dep_coords:
                     # The axis is not dependent on this instance's axis, so we
                     # just compute the values once and broadcast along this
                     # dimension later.
--- a/glue/core/component.py
+++ b/glue/core/component.py
@@ -235,8 +235,8 @@ class CoordinateComponent(Component):
             if self.world_n_dim < self._data.ndim:
                 raise ValueError(f"World[{self.world_n_dim}] must have at least the same "
                                   "number of dimensions as data[{self._data.ndim}].")
-            else:
-                self.world_n_dim = getattr(self._data, 'ndim', 0)
+        else:
+            self.world_n_dim = getattr(self._data, 'ndim', 0)
 
     @property
     def data(self):
--- a/glue/core/component.py
+++ b/glue/core/component.py
@@ -234,7 +234,7 @@ class CoordinateComponent(Component):
             self.world_n_dim = getattr(self._data.coords, 'world_n_dim', self._data.ndim)
             if self.world_n_dim < self._data.ndim:
                 raise ValueError(f"World[{self.world_n_dim}] must have at least the same "
-                                  "number of dimensions as data[{self._data.ndim}].")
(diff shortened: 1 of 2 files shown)


glue/core/component.py
```diff
@@ -328,7 +330,8 @@ def _calculate(self, view=None):
             pix_coords = np.meshgrid(*pix_coords, indexing='ij', copy=False)
 
             # Finally we convert these to world coordinates
-            axis = self._data.ndim - 1 - self.axis
+            world_n_dim = getattr(self._data.coords, 'world_n_dim', self._data.ndim)
+            axis = world_n_dim - 1 - self.axis
```
dhomeier:
```suggestion
            axis = world_n_dim - 1 - self.axis
```
Should not need to recalculate here as this is unchanged from the top.
But either way, I seem to be getting the right data arrays this way (at least for Wavelength and Lat), but always only the first value broadcast to the full shape. I think this is due to the calculation of `dep_coords`  becoming offset with the change to `world_n_dim`.

dhomeier (changes requested):
Offsetting `i in dep_coords` by something like `world_n_dim - self._data.ndim` might resolve the dependent coordinates issue, but I am not sure about this in the general case. Could be safer to skip the `pix_coord = 0` in this case altogether; worst case should probably be some redundant computation.

dhomeier:
Just to summarise my findings, using `axis = world_n_dim - 1 - self.axis` in `world_coords` pulls in the correct arrays for Wavelength and Lat, but the setting of `pix_coord - 0` above lets them only return the first value. Either the calculation of `dep_coords` has to account for the offset in this case, or just skip that block whenever `world_n_dim != self._data.ndim` (basically playing it safe and calculating the whole array).

dhomeier:
Seems I cannot push to this branch, so for the record my diff:
```diff
--- a/glue/core/component.py
+++ b/glue/core/component.py
@@ -260,6 +260,7 @@ class CoordinateComponent(Component):
             # To optimize this, we therefore essentially consider only the
             # dependent dimensions and then broacast the result to the full
             # array size at the very end.
+            axis = world_n_dim - 1 - self.axis
 
             # view=None actually adds a dimension which is never what we really
             # mean, at least in glue.
@@ -270,7 +271,6 @@ class CoordinateComponent(Component):
             # convert these straight to world coordinates since the indices
             # of the pixel coordinates are the pixel coordinates themselves.
             if isinstance(view, (tuple, list)) and isinstance(view[0], np.ndarray):
-                axis = world_n_dim - 1 - self.axis
                 return pixel2world_single_axis(self._data.coords, *view[::-1],
                                                world_axis=axis)
 
@@ -318,7 +318,7 @@ class CoordinateComponent(Component):
                 else:
                     final_shape.append(self._data.shape[i])
 
-                if i not in dep_coords:
+                if world_n_dim == self._data.ndim and i not in dep_coords:
                     # The axis is not dependent on this instance's axis, so we
                     # just compute the values once and broadcast along this
                     # dimension later.
@@ -330,7 +330,6 @@ class CoordinateComponent(Component):
             pix_coords = np.meshgrid(*pix_coords, indexing='ij', copy=False)
 
             # Finally we convert these to world coordinates
-            axis = world_n_dim - 1 - self.axis
             world_coords = pixel2world_single_axis(self._data.coords,
                                                    *pix_coords[::-1],
                                                    world_axis=axis)
```
Modifying the example from #2574 to get a sample of the coordinate arrays
```python
for comp in components:
    comp_obj = data.get_component(comp.label)
    print(f'component {comp.label} -> unit={comp_obj.units} -- axis {getattr(comp_obj, "axis", None)} of {comp_obj._data.ndim} {comp_obj.data.shape}')
    print(comp_obj.data.T[100::1000,::5])
```
with this PR I get
```python
component Pixel Axis 0 [y] -> unit= -- axis 0 of 2 (28, 3320)
[[ 0  5 10 15 20 25]
 [ 0  5 10 15 20 25]
 [ 0  5 10 15 20 25]
 [ 0  5 10 15 20 25]]
component Pixel Axis 1 [x] -> unit= -- axis 1 of 2 (28, 3320)
[[ 100  100  100  100  100  100]
 [1100 1100 1100 1100 1100 1100]
 [2100 2100 2100 2100 2100 2100]
 [3100 3100 3100 3100 3100 3100]]
component Wavelength -> unit=um -- axis 0 of 2 (28, 3320)
[[2.86158317 2.86158317 2.86158317 2.86158317 2.86158317 2.86158317]
 [2.86158317 2.86158317 2.86158317 2.86158317 2.86158317 2.86158317]
 [2.86158317 2.86158317 2.86158317 2.86158317 2.86158317 2.86158317]
 [2.86158317 2.86158317 2.86158317 2.86158317 2.86158317 2.86158317]]
component Lat -> unit=deg -- axis 1 of 2 (28, 3320)
[[30.14620542 30.14620542 30.14620542 30.14620542 30.14620542 30.14620542]
 [30.14620542 30.14620542 30.14620542 30.14620542 30.14620542 30.14620542]
 [30.14620542 30.14620542 30.14620542 30.14620542 30.14620542 30.14620542]
 [30.14620542 30.14620542 30.14620542 30.14620542 30.14620542 30.14620542]]
component flux -> unit=MJy -- axis None of 2 (28, 3320)
[[           nan -1.0203675e-09 -8.2875534e-10  2.2015776e-09  -1.2906758e-09 -5.5979354e-10]
 [           nan -7.1334699e-10 -6.1566885e-10  1.6646968e-09  -9.7279906e-10 -4.9086002e-10]
 [           nan -4.6162213e-10 -4.5124834e-10  1.2105327e-09  -6.9065759e-10 -3.5875511e-10]
 [           nan -2.5304384e-10 -3.8256823e-10  1.0252108e-09  -4.3328327e-10 -2.6573441e-10]]
component uncertainty -> unit=MJy -- axis None of 2 (28, 3320)
[[          nan 2.1030137e-12 2.2375760e-12 8.2385140e-12 2.2478037e-12  2.1513822e-12]
 [          nan 2.1044958e-12 2.0450718e-12 6.6202395e-12 2.2610462e-12  2.0251385e-12]
 [          nan 5.7586080e-12 4.4686230e-12 1.0688311e-11 6.6592925e-12  6.2164562e-12]
 [          nan 2.3650879e-12 2.3732617e-12 7.9581671e-12 2.3579600e-12  2.7324607e-12]]
```
while with the patch above it becomes
```python
component Pixel Axis 0 [y] -> unit= -- axis 0 of 2 (28, 3320)
[[ 0  5 10 15 20 25]
 [ 0  5 10 15 20 25]
 [ 0  5 10 15 20 25]
 [ 0  5 10 15 20 25]]
component Pixel Axis 1 [x] -> unit= -- axis 1 of 2 (28, 3320)
[[ 100  100  100  100  100  100]
 [1100 1100 1100 1100 1100 1100]
 [2100 2100 2100 2100 2100 2100]
 [3100 3100 3100 3100 3100 3100]]
component Wavelength -> unit=um -- axis 0 of 2 (28, 3320)
[[2.92947268 2.92947268 2.92947268 2.92947268 2.92947268 2.92947268]
 [3.60587639 3.60587639 3.60587639 3.60587639 3.60587639 3.60587639]
 [4.37357866 4.37357866 4.37357866 4.37357866 4.37357866 4.37357866]
 [5.03441268 5.03441268 5.03441268 5.03441268 5.03441268 5.03441268]]
component Lat -> unit=deg -- axis 1 of 2 (28, 3320)
[[30.14620542 30.14612544 30.14604545 30.14596546 30.14588548 30.14580549]
 [30.14620542 30.14612544 30.14604545 30.14596546 30.14588548 30.14580549]
 [30.14620542 30.14612544 30.14604545 30.14596546 30.14588548 30.14580549]
 [30.14620542 30.14612544 30.14604545 30.14596546 30.14588548 30.14580549]]
component flux -> unit=MJy -- axis None of 2 (28, 3320)
...
```
(`Lon`, if extracted manually as axis 2, decreases from 247.89120798 to 247.89050047 along the short axis)
@astrofrog I believe the offset in determining the correct `dep_coords` arises from still constructing `pix_coord` over `self._data.ndim`, while `axis` is now counted from `world_n_dim` (1 higher in this case); but I don't know how to reliably correct for this – any ideas?

MatthewPortman:
@dhomeier I pushed the refactor to this branch but looking at your diff, our line numbers are slightly off. Any idea why that might be?

dhomeier:
Looks like I removed the world_n_dim in L333 already in an earlier refactor I forgot about; everything should be correct here though, thanks!

dhomeier (approved):
Approving this as functional for the scope tested here, but I am holding off merging for the moment in case @astrofrog still has comments or suggestions for an alternative solution.

astrofrog (changes requested):
@MatthewPortman - sorry for the delay, I think this seems fine but could you add a unit test to exercise these changes? Also could you make sure we properly handle cases where world_n_dim < data.ndim? (or explicitly disallow it if it's tricky to get it to work)

MatthewPortman:
> @MatthewPortman - sorry for the delay, I think this seems fine but could you add a unit test to exercise these changes? Also could you make sure we properly handle cases where world_n_dim < data.ndim? (or explicitly disallow it if it's tricky to get it to work)

@astrofrog I tried previously and it's quite tricky... beyond what I'm capable to do with my level of familiarity. I can raise an error on `CoordinateComponent` instantiation if you think that's appropriate.

astrofrog:
I've pushed a regression test for the unit issue - I would like to figure out the proper fix for the missing component in a follow-up PR.

glue/core/component.py
```diff
@@ -229,14 +229,21 @@ def __init__(self, data, axis, world=False):
         self._data = data
         self.axis = axis
 
+        # From issue #2574: initialize world_n_dim and error for data ndim > world ndim.
+        if self.world and self._data:
+            self.world_n_dim = getattr(self._data.coords, 'world_n_dim', self._data.ndim)
+            if self.world_n_dim < self._data.ndim:
+                raise ValueError("World must have at least the same number of "
+                                 "dimensions as data.")
```
dhomeier:
```suggestion
        if hasattr(self._data, 'coords'):
            self.world_n_dim = getattr(self._data.coords, 'world_n_dim', self._data.ndim)
            if self.world_n_dim < self._data.ndim:
                raise ValueError(f"World[{self.world_n_dim}] must have at least the same "
                                  "number of dimensions as data[{self._data.ndim}].")
            else:
                self.world_n_dim = getattr(self._data, 'ndim', 0)
```
surprisingly difficult to proof this against unexpected referencing of `world_n_dim`, but apparently test_python_export.py is running into it even when `ndim` does not exist!
This passes locally, though maybe we should think about putting the checks in `_calculate` back in to only set and access `world_n_dim` where it's really needed...

dhomeier:
I had missed your added test, thanks for fixing!
CI is looking good now.

astrofrog (approved):

PR #2577, opened 2026-03-20
Fix issue with reference data choices and update attributes

### LINKED ISSUE #2576: Reference data dropdown doesn't update
**Describe the bug**
The reference data dropdown currently does not update with new options when more 3D data sets are added.

**To Reproduce**
Steps to reproduce the behavior such as:
1. Open glue
2. Add in two 3D datasets that are linked
3. Add one to a volume viewer. The reference data selection should have one item, that dataset
4. Add the other dataset
5. The reference data dropdown still only has one layer, but it should have two

**Expected behavior**
The reference data dropdown should get updated

**Details:**

- Operating System: Ubuntu 24.04
- Python version (``python --version``): Python 3.12.13
- Glue version (``glue --version``): current dev
- How you installed glue: pip, from source

**Additional context**
This is my fault, due to a change that I made [here](https://github.com/glue-viz/glue-vispy-viewers/pull/399/changes/3660dc49a48a47838cc1d71d8afd66827a567adf). I was trying to cut down on excessive updates, but mistakenly believed that the `layers` callback got both the old and new state of the list (it doesn't, so the equality check is always true)

Carifio24:
This PR aims to fix #2576 by removing the check on the layers list, while also trying to resolve the issue that the check was put in there to work around. It also makes a couple of other tweaks to the attributes setup. The basic changes are:
* The reference data dropdown now gets updated on a `layers` callback
* x/y/z attribute choices are now set specifically from the reference data
* We set up default x/y/z attributes for the standard cube pixel coordinate ordering of 2, 1, 0 (currently we are just using 0, 1, 2). I'm not sure if there is any sort of convention for 4D+ cubes, but users can always change attributes anyways

### DIFF
--- a/glue/viewers/volume3d/viewer_state.py
+++ b/glue/viewers/volume3d/viewer_state.py
@@ -44,8 +44,17 @@ class VolumeViewerState3D(ViewerState3D):
                 return layer_state.layer
 
     def _layers_changed(self, *args):
-        self._update_combo_ref_data()
-        self._set_reference_data()
+
+        # By default if any of the state properties change, this triggers a
+        # callback on anything listening to changes on self.layers - but here
+        # we just want to know if any layers have been removed/added so we keep
+        # track of the UUIDs of the layers and check this before continuing.
+        current_layers = [layer_state.layer.uuid for layer_state in self.layers]
+        if not hasattr(self, '_last_layers') or self._last_layers != current_layers:
+            self._update_combo_ref_data()
+            self._set_reference_data()
+            self._last_layers = current_layers
+            return
 
     def _update_combo_ref_data(self, *args):
         self.ref_data_helper.set_multiple_data(self.layers_data)
--- a/glue/viewers/volume3d/viewer_state.py
+++ b/glue/viewers/volume3d/viewer_state.py
@@ -26,7 +26,8 @@ class VolumeViewerState3D(ViewerState3D):
         self.ref_data_helper = ManualDataComboHelper(self, 'reference_data')
         self.slices = ()
 
-        self.add_callback('layers', self._layers_changed, echo_old=True)
+        self.add_callback('layers', self._layers_changed)
+        self.add_callback('reference_data', self._reference_data_changed)
         self.add_callback('x_att', self._on_xatt_changed, echo_old=True)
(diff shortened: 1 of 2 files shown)


astrofrog:
Thanks! Could you add a regression test?

Carifio24:
@astrofrog I've added a few tests that I think should cover everything here

glue/viewers/volume3d/viewer_state.py
```diff
@@ -42,29 +43,33 @@ def _first_3d_data(self):
             if getattr(layer_state.layer, 'ndim', None) >= 3:
                 return layer_state.layer
 
-    def _layers_changed(self, old_layers, new_layers):
-        if self.reference_data is not None and old_layers == new_layers:
-            return
+    def _layers_changed(self, *args):
```
astrofrog, line 46: I think the reason for the old code was that if any attributes of the layers are changed, e.g. color on one of the layers, this callback will fire, so I think we were checking whether the list of layers actually changed, or just their contents. Are we going to be updating too often now?

Carifio24: Yeah that is why I had put the guard in there before, but the problem is that it's actually too restrictive (hence this PR). echo will give back the [same list instance](https://github.com/glue-viz/echo/blob/main/echo/containers.py#L224) for both the old and new arguments and so the combobox never gets updated because the equality check always resolves as true. I guess what we could do here is to keep a cached list of the previous set of layers and check the new value against that.

astrofrog: I think we should probably keep a list of the layer UUIDs and compare that to know whether to proceed with the fallback - I could have sword we had this pattern elsewhere, will try and find.

astrofrog: Found it:

https://github.com/glue-viz/glue/blob/9f6b72ffc0c392cd0c01a6ec8a06d5386bb5f797/glue/viewers/profile/state.py#L253-L264

Carifio24: Perfect, I've updated what we do here to have the same pattern (and explanatory comment)

astrofrog (approved):

PR #2587, opened 2026-07-07
Write the plugins config file atomically and tolerate duplicate options when reading it

astrofrog:
This should hopefully fix the issues seen in jdaviz  when running the tests in parallel

### DIFF
--- a/glue/_plugin_helpers.py
+++ b/glue/_plugin_helpers.py
@@ -93,6 +93,13 @@ class PluginConfig(object):
 
         plugin_cfg = os.path.join(cfg_dir, 'plugins.cfg')
 
+        # Don't rewrite the file when nothing has changed. In the common case
+        # the config already lists every known plugin, so this avoids touching
+        # disk on every startup and removes almost all of the opportunity for
+        # concurrent processes to write the file at the same time.
+        if dict(PluginConfig.load().plugins) == dict(self.plugins):
+            return
+
         import configparser
 
         config = configparser.ConfigParser()
@@ -108,30 +115,30 @@ class PluginConfig(object):
         # destination in a single step, so a concurrent reader (e.g. another
         # process during a parallel test run) always sees either the old or the
         # new file, never a partially-written or interleaved one.
-        import time
         import tempfile
         fd, tmp_path = tempfile.mkstemp(dir=cfg_dir, prefix='plugins.cfg.',
                                         suffix='.tmp')
         try:
             with os.fdopen(fd, 'w') as fout:
                 config.write(fout)
-            # Windows refuses to rename over a file that another process or
-            # thread currently has open (PermissionError / WinError 5), unlike
-            # POSIX. Retry briefly to ride out those transient sharing windows.
-            for attempt in range(20):
-                try:
-                    os.replace(tmp_path, plugin_cfg)
-                    break
-                except PermissionError:
-                    if attempt == 19:
-                        raise
-                    time.sleep(0.05)
-        except BaseException:
             try:
-                os.remove(tmp_path)
+                os.replace(tmp_path, plugin_cfg)
             except OSError:
-                pass
-            raise
+                # Windows refuses to rename over a file that another process
+                # currently has open, unlike POSIX. That can only happen once a
+                # valid file already exists (every write is a whole-file rename)
+                # and the config is non-critical, so ignore it. A genuine
+                # failure leaves no file in place and is re-raised.
+                if not os.path.exists(plugin_cfg):
+                    raise
+        finally:
+            # Remove the temporary file if the rename did not consume it (an
+            # error, or an ignored Windows sharing conflict).
+            if os.path.exists(tmp_path):
+                try:
+                    os.remove(tmp_path)
+                except OSError:
+                    pass
 
     def filter(self, keep):
         """
--- a/glue/_plugin_helpers.py
+++ b/glue/_plugin_helpers.py
@@ -9,6 +9,7 @@
 import os
 import sys
 import warnings
+from pathlib import Path
 from collections import defaultdict
 
 if sys.version_info >= (3, 10):
@@ -122,7 +123,7 @@ class PluginConfig(object):
             with os.fdopen(fd, 'w') as fout:
                 config.write(fout)
             try:
-                os.replace(tmp_path, plugin_cfg)
+                Path(tmp_path).replace(plugin_cfg)
             except OSError:
                 # Windows refuses to rename over a file that another process
                 # currently has open, unlike POSIX. That can only happen once a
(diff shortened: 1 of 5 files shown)


glue/tests/test_plugin_helpers.py
```diff
@@ -0,0 +1,69 @@
+import os
+import configparser
+from concurrent.futures import ThreadPoolExecutor
+
```
dhomeier, line 4:
```suggestion
from pathlib import Path, PosixPath

```
ruff-flake8 suggestion

glue/tests/test_plugin_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+    assert loaded.plugins['plugin_a'] is True
+    assert loaded.plugins['plugin_b'] is False
+
+
+def test_save_leaves_no_temporary_files(tmp_path, monkeypatch):
+    monkeypatch.setattr('glue.config.CFG_DIR', str(tmp_path))
+
+    PluginConfig(plugins={'plugin_a': True}).save()
+
+    assert os.listdir(tmp_path) == ['plugins.cfg']
```
dhomeier:
```
    assert List(Path(tmp_path).iterdir()) == [PosixPath('plugins.cfg')]
```
seems to be, in all seriousness, what Ruff suggests as more readable solution for [PTH208](https://docs.astral.sh/ruff/rules/os-listdir/). Maybe this will just do

```suggestion
    assert os.listdir(tmp_path) == ['plugins.cfg']  # noqa: PTH208
```

glue/tests/test_plugin_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+            if os.path.exists(plugin_cfg):
+                configparser.ConfigParser(strict=True).read(plugin_cfg)
+
+    with ThreadPoolExecutor(max_workers=6) as executor:
+        futures = [executor.submit(writer) for _ in range(4)]
+        futures += [executor.submit(reader) for _ in range(2)]
+        for future in futures:
+            future.result()
+
+    assert os.listdir(tmp_path) == ['plugins.cfg']
```
dhomeier:
```suggestion
    assert os.listdir(tmp_path) == ['plugins.cfg']  # noqa: PTH208
```

glue/_plugin_helpers.py
```diff
@@ -8,6 +8,7 @@
 
 import os
 import sys
+import warnings
 from collections import defaultdict
```
dhomeier, line 14:
```suggestion
from pathlib import Path

```

glue/_plugin_helpers.py
```diff
@@ -88,11 +101,26 @@ def save(self):
...
+        # POSIX and Windows, so a concurrent reader (e.g. another process during
+        # a parallel test run) always sees either the old or the new file, never
+        # a partially-written or interleaved one.
+        import tempfile
+        fd, tmp_path = tempfile.mkstemp(dir=cfg_dir, prefix='plugins.cfg.',
+                                        suffix='.tmp')
+        try:
+            with os.fdopen(fd, 'w') as fout:
+                config.write(fout)
+            os.replace(tmp_path, plugin_cfg)
```
dhomeier:
```suggestion
            Path(tmp_path).replace(plugin_cfg)
```
PTH105 looks rather more sensible

dhomeier (changes requested):
For fixing pre-commit I'd probably just add an exception for PTH208 (permanent or not), but replacing `os.replace` may be reasonable.

glue/_plugin_helpers.py
```diff
@@ -88,11 +102,26 @@ def save(self):
...
+        os.makedirs(cfg_dir, exist_ok=True)
+
+        # Write atomically: serialize to a uniquely-named temporary file in the
+        # same directory and rename it into place. os.replace is atomic on both
+        # POSIX and Windows, so a concurrent reader (e.g. another process during
+        # a parallel test run) always sees either the old or the new file, never
+        # a partially-written or interleaved one.
+        import tempfile
+        fd, tmp_path = tempfile.mkstemp(dir=cfg_dir, prefix='plugins.cfg.',
+                                        suffix='.tmp')
```
dhomeier, line 121:
```suggestion
        fd, tmp_path = tempfile.mkstemp(dir=Path(plugin_cfg).parent,
                                        prefix=Path(plugin_cfg).name, suffix='.tmp')
```
(PTH119 & PTH120) should actually make this more robust.
Or alternatively maybe at least move the declaration of `plugin_cfg` right above this, just so in case one every decides to use a different filename it becomes obvious here?

dhomeier:
Seems like in the end `Path.replace` is running `os.replace` under the hood, so I don't believe the Windows failure is due to that – general permissions restrictions on Win? Don't know if explicit `os.remove(plugin_cfg)` first might help, in case the overwriting is the problem. But those functions are all labeled as "posix"...

astrofrog:
@dhomeier - I decided to update the approach here:
* Define a pytest plugin that will make sure that when in a pytest context, we use temporary directories for the config (a good idea anyway) - when using pytest-xdist each worker will get its own directory. This alone should fix jdaviz
* Only ever write out the file if it has changed on disk
* Make the atomic writing more robust to try and fix the Windows issue

astrofrog:
@dhomeier - I think this is ready, can you review? (the visual test failure is unrelated)

glue/_plugin_helpers.py
```diff
@@ -80,6 +94,13 @@ def save(self):
 
         plugin_cfg = os.path.join(cfg_dir, 'plugins.cfg')
 
+        # Don't rewrite the file when nothing has changed. In the common case
+        # the config already lists every known plugin, so this avoids touching
+        # disk on every startup and removes almost all of the opportunity for
+        # concurrent processes to write the file at the same time.
+        if dict(PluginConfig.load().plugins) == dict(self.plugins):
+            return
```
dhomeier, line 95: 👍 could probably be at the very top though?

glue/_plugin_helpers.py
```diff
@@ -88,11 +109,37 @@ def save(self):
...
+        try:
+            with os.fdopen(fd, 'w') as fout:
+                config.write(fout)
+            try:
+                Path(tmp_path).replace(plugin_cfg)
+            except OSError:
+                # Windows refuses to rename over a file that another process
+                # currently has open, unlike POSIX. That can only happen once a
+                # valid file already exists (every write is a whole-file rename)
+                # and the config is non-critical, so ignore it. A genuine
```
dhomeier: If the write is genuinely redundant (no changes in config), shouldn't `save` have already exited above?
IOW, is this guaranteed to only be used when no relevant settings need to be saved, or would it be better to raise a Warning here as well (of course risking to fail in downstream tests)?

dhomeier (approved):
I am not too familiar with the plugin system, but this works now (at least in our CI environment, macOS failure was transient as well), so just a few minor comments on the implementation.

PR #2588, opened 2026-07-08
Check if reference data before updating volume attributes

Carifio24:
This PR is a follow-up to #2577 (and hopefully the last piece of this story). The volume viewer state now has the problem that `_reference_data_changed` will be called even when the reference data hasn't actually changed, which has the problematic side effect of updating the x/y/z attributes whenever a new layer is added to the viewer. I'm not 100% sure but I think this issue ultimately comes from the combo helper. The approach adopted here is to copy what the image viewer state does, and store a `_last_reference_data` member that we can compare to inside the `_reference_data_changed` callback.

### DIFF
--- a/glue/viewers/volume3d/viewer_state.py
+++ b/glue/viewers/volume3d/viewer_state.py
@@ -67,14 +67,19 @@ class VolumeViewerState3D(ViewerState3D):
                     return
 
     def _reference_data_changed(self, data):
-        if data is None:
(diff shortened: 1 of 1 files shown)


Carifio24:
Also worth noting that the problem fixed here is what's causing the current failures in the glue-vispy-viewers CI.

astrofrog (approved):
Looks good, thanks!

PR #2589, opened 2026-07-12
Pass viewer state through to superclass in 3D layer states

Carifio24:
Currently the 3D layer state classes (scatter and volume) don't forward the viewer state keyword argument through to their superclasses. This means that their `viewer_state` members are not set correctly (they end up being `None`). Note that the base `LayerArtist` class explicitly sets this [here](https://github.com/glue-viz/glue/blob/main/glue/viewers/common/layer_artist.py#L25).


Carifio24:
Originally I had this pass all of the keyword arguments through, but I can see now why that causes problems. Adding just `viewer_state` seems to work fine though.

astrofrog (approved):

PR #2594, opened 2026-08-19
Add ndim check for image viewer

MatthewPortman:
# Pull Request Template

## Description

We noticed a circumstance in Jdaviz where glue attempts to set the viewer's reference layer to 1D data, i.e. catalog data, and errors out with the following error (full traceback can be provided on request):

```
File /opt/miniconda3/envs/jdaviz/lib/python3.13/site-packages/glue/viewers/image/state.py:272, in ImageViewerState._on_xatt_world_change(self, forced, *args)
    270     world_ids = self.reference_data.pixel_component_ids
    271 if self.x_att_world == world_ids[-1]:
--> 272     self.y_att_world = world_ids[-2]
    273 else:
    274     self.y_att_world = world_ids[-1]

IndexError: list index out of range
```

This can be reproduced by:
1. Loading an Image into an Image viewer and then a Catalog as a scatter plot on the viewer.
2. Removing the image from the viewer. 

The fix therefore places a check on the number of dimensions required for data to be used as a reference layer.


astrofrog:
Looks good, thanks! Could you add a regression test? (not a visual test, just a regular test to make sure things don't crash when you remove the image dataset)

glue/viewers/image/tests/test_state.py
```diff
@@ -471,3 +471,42 @@ def test_attribute_units_invalid():
...
+    # Verify that reference_data is None (1D data is filtered out)
+    assert viewer_state.reference_data is None
+
+    data_2d = Data(x=np.ones((5, 10)))
+    layer_state_2d = ImageLayerState(layer=data_2d, viewer_state=viewer_state)
+    viewer_state.layers.append(layer_state_2d)
+
+    # Verify that 2D data was selected as reference
+    assert viewer_state.reference_data is data_2d
+
```
dhomeier, line 496:
```suggestion

    # Add a second 2D dataset, set this as reference, then remove it again
    more_data_2d = Data(x=np.ones((10, 20)) * np.arange(1, 21))
    another_layer_state_2d = ImageLayerState(layer=more_data_2d, viewer_state=viewer_state)
    viewer_state.layers.append(another_layer_state_2d)
    assert viewer_state.reference_data is data_2d

    viewer_state.reference_data = more_data_2d
    assert viewer_state.reference_data is more_data_2d

    # Verify that the original 2D data becomes reference again
    viewer_state.layers.remove(another_layer_state_2d)
    assert viewer_state.reference_data is data_2d

```
This came up when wondering how a second 2D dataset is handled – turns out it apparently does the right thing and falls back to the remaining 2D set.

MatthewPortman: Great idea, thanks!

dhomeier (approved):
Looks all good to me now, thanks!

